Issue #92, opened 2013-08-02
Mutation Killer for MiniTest


CyborgMaster:
I'm starting a new Rails project.  I currently use SimpleCov to monitor my test code coverage, and I would love to use Mutant to do mutation testing and coverage as well.  I have seen that some progress has been done on getting Mutant running in a rails environment, and was wondering if anyone has started any work on developing a mutation killer for the minitest framework, which is what my project uses.  I know the correct response is to fork, impliment it myself, and submit a PR, but we are on a tight deadline and I don't have time to do it now and would prefer to collaborate with someone if there is already effort in that direction.

kbrock:
Is there a non-working/stagnant branch somewhere that I could look at?
164 references @ahawkins but I didn't track down anything.

plexus:
I'm guessing https://github.com/ahawkins/mutant/tree/minitest is the branch being referred to.

mbj:
@plexus In meanwhile there is a much more maintained one by me, that lacks people testing it: https://github.com/kbrock/mutant/tree/minitest this branch is "far less stale" than the one you are referencing.

@kbrock With your agreement I'd move this branch to my repo + create a PR to give it more visibility.

kbrock:
:+1:

plexus:
Awesome!

PR #164, opened 2014-03-16
[WIP] Minitest

mbj:
References: #92 

TODO:
- [x] Noop reporter
- [ ] Test selection


kbrock:
fun to test minitest with rspec.

mbj:
both minitest and rspec are not fun to test at all. As both maintain global accessible state between test runs. Reducing the chance for parallelization and idempotent execution of specs from specs.

PR #247, opened 2014-08-17
[WIP] Config file support

mbj:
A work in progress branch on supporting configuration files for mutants.

Configuration files should also be persistable form CLI, hence I'm using the binjective mappings from morpher.


mbj:
Will not be finished in this repo. Closing.

PR #285, opened 2015-01-14
Support 2.2

mbj:


mbj:
Rubygems fails to propagate freshly released `unparser-0.2.2` to all of its edge locations. The build passes locally so I'm merging now.

PR #290, opened 2015-01-20
The gist mentioned in the README is no longer present / public

StoneCypher:
Removed the link and added the text ", now removed, "


mbj:
Thx.

PR #312, opened 2015-04-21
Bump parser dependency to allow 2.2.2.1

chastell:
RuboCop 0.30.1 depends on parser 2.2.2.1, which means it can’t work in a project where mutant requires parser 2.2.0.x. This allows mutant to work with parser 2.2.2.1.

### DIFF
--- a/mutant.gemspec
+++ b/mutant.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = '>= 1.9.3'
 
-  gem.add_runtime_dependency('parser',        '~> 2.2.0.2')
+  gem.add_runtime_dependency('parser',        '~> 2.2', '>= 2.2.0.2')
   gem.add_runtime_dependency('ast',           '~> 2.0')
   gem.add_runtime_dependency('diff-lcs',      '~> 1.2')
   gem.add_runtime_dependency('parallel',      '~> 1.3')


mutant.gemspec
```diff
@@ -23,7 +23,7 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = '>= 1.9.3'
 
-  gem.add_runtime_dependency('parser',        '~> 2.2.0.2')
+  gem.add_runtime_dependency('parser',        '~> 2.2', '>= 2.2.0.2')
```
mbj: See discussion in https://github.com/mbj/unparser/pull/43#discussion_r28829610

chastell:
Actually mutant needs to have this applied before unparser can use parser 2.2.2.1, so they need to be changed in lockstep.

Do you want me to change the dependency to `~> 2.2.2` here as well? Note that this will make mutant unbundlable (as it depends on unparser which forces parser 2.2.0).

mbj:
@chastell I think we want `~> 2.2.2` for both. I'll break the cycle via verifying mutant is all fine with `~> 2.2.2` against unparser git version, but release both under the assumption the other is already released.

chastell:
…or you could just merge this PR (which is green) as-is and then trigger https://github.com/mbj/unparser/pull/43 rebuild: unparser depends on mutant’s master, so merging this means https://github.com/mbj/unparser/pull/43 should rebuild against parser 2.2.2 without any broken releases in the meantime.

This way (a) mutant is temporarily accepting both 2.2.0 and 2.2.2, (b) mutant’s master remains green, (c) unparser’s tests can be run against parser 2.2.2 and (d) if it’s red unparser can be fixed in https://github.com/mbj/unparser/pull/43 before that’s merged to unparser’s master.

Afterwards both dependencies can be tightened to parser 2.2.2+, of course.

mbj:
Okay, this branch is an iterative improvement. I'll probably want to specify the dependency in a different way longterm. But till I have the time to do this I do not see any reason to hold this back. Merging.

Thx @chastell.

PR #315, opened 2015-04-26
remove duplicate definition warnings

kbrock:
These are throwing warning


mbj:
@kbrock Good catch. I was probably too blind to rspec warnings over time.

I'll merge this one now. For the future, can you for consistence with the other commits please use a capitalized first word in your commit message?

PR #316, opened 2015-04-26
Move test classes (Foo, Bar) from global space to anonymous local

kbrock:
In rspec, defining a class in a describe actually describes in the global scope (e.g. outside the describe).

So this was complaining that the class was defined multiple times. Which is true.

Just let me know where you want to move and what to call.
(probably shouldn't be defining classes in specs in general)

### DIFF
--- a/spec/unit/mutant/matcher/methods/instance_spec.rb
+++ b/spec/unit/mutant/matcher/methods/instance_spec.rb
@@ -1,39 +1,40 @@
 RSpec.describe Mutant::Matcher::Methods::Instance, '#each' do
-  let(:object) { described_class.new(env, Foo) }
+  let(:object) { described_class.new(env, class_under_test) }
   let(:env)    { Fixtures::TEST_ENV            }
 
   subject { object.each { |matcher| yields << matcher } }
 
   let(:yields) { [] }
 
-  module Bar
-    def method_d
+  let(:class_under_test) do
+    parent = Module.new do
+      def method_d
+      end
+
+      def method_e
+      end
     end
 
-    def method_e
+    Class.new do
+      include parent
+
+      private :method_d
+
+      public
+
+      def method_a
+      end
+
+      protected
+
+      def method_b
+      end
+
+      private
+
+      def method_c
+      end
     end
-  end
-
-  class Foo
-    include Bar
-
-    private :method_d
-
-  public
-
-    def method_a
-    end
-
-  protected
-
-    def method_b
-    end
-
-  private
-
-    def method_c
-    end
-
   end
 
   let(:subject_a) { double('Subject A') }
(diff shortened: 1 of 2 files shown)


kbrock:
this worked for me via `chruby ruby-2.0.0-p643 ; bundle exec rake ci`

(only test that failed was the tty one)

kbrock:
Trouble reproducing this.
I installed 2.0.0-p598.

```
rspec --order rand:3821 ./spec/integration/mutant/rspec_spec.rb:20 # rspec integration RSpec 3.0 behaves like rspec integration it allows to exclude mutations

bundle exec rake rspec --order rand:3821# no rake ci --order option
```

Both of these worked for me.

ideas?

mbj:
> (probably shouldn't be defining classes in specs in general)

Its IMO fine in case we define a class in a *unique* position in the RSpec generated class hierarchy.

I actually do not like we move the non unique class definition into a new "global top level namespace" `Foo` and `Bar`. I wounder if we want to have a namespace like `MutantSpec` we can use to centralize all "example constructs" we need to setup for testing parts of mutant that need to read global state the way a mutant usage would define it.

mbj:
I think both specs do actually not *need* to define the classes in a global namespace. We could move them to anonymous classes within a `let` block.

Like:

```ruby
describe ... do
  let(:class_under_test) do
    Class.new do
      def method_b
      end
      # ...
   end

     let(:object) { described_class.new(env, class_under_test) }
end
```

We could have a shared rspec context to share this class structure between the both instance and singleton specs.

This strategy (untested) is better because:

* No "require namespace polution" (does not add `foo_bar` require entry point)
* Reduces global / leaked state
* Fixed the uniqueness problem that let you do this PR in the first place

@kbrock Just let me know if you want to do this change for your own. I'm fine to do it by myself. Just trying to avoid duplication of effort.

kbrock:
Tricky to get this through.

Kept getting nailed by indentation.

Also apparently `include Module.new` fails while `extend Module.new` works

mbj:
@kbrock Do you want me to finish this? I really like the idea to cut the duplication, but want to avoid duplicate effort with this question.

kbrock:
@mbj yea, take it.

I put it into `Class.new` classes but seems travis is dinging me for some minor stuff.

kbrock:
pushed the identical test again. Curious to seeing what the failures look like this time.

kbrock:
Huh. Don't know how many times I saw those errors and didn't realize they were rubocop errors. I botched the indentation on a method. Hope that will fix it.

mbj:
I'm still not having any time for OSS, thx for your patience.

mbj:
@kbrock Thx!

PR #318, opened 2015-04-26
single spec for all literal tests

kbrock:
This code confuses me, as 99% of the time, there is a single spec per lib file.
Textmate and sublime uses this assumption to switch between tests too.

Do these really need to be separate files?

Is the problem here that the test app is too simple?

### DIFF
--- a/test_app/spec/unit/test_app/literal/command_spec.rb
+++ b/test_app/spec/unit/test_app/literal/command_spec.rb
@@ -1,9 +1,0 @@
-require 'spec_helper'
-
-RSpec.describe TestApp::Literal, '#command' do
-  subject { object.command(double) }
-
-  let(:object) { described_class.new }
-
-  it { should be(object) }
-end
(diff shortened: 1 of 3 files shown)


mbj:
> Do these really need to be separate files?

No, the fact these are separate files is from a very ancient mutant version that did use the file name as matching criteria.

> Is the problem here that the test app is too simple?

The test app is a minimal integration test. It should have the following sematics under mutation testing:

* A fully covered subject
* A partially covered subject
* Does not maintain mutable global state
* Has instance methods
* Has singleton methods
* (future) has subjects with failing tests

What kind of mutations are happending is not important (anymore), since the full mutation specification is in the [mutant-meta](https://github.com/mbj/mutant/tree/master/meta) so mutation wise it should NOT be exhaustive.

mbj:
@kbrock After I pushed c392cb5e4bdec88073eca0ee9f8b9ef97c5fde33 this does not merge clearly anymore. The rebase on your side should be trivial as I only nuked redundant encoding headers.

PR #319, opened 2015-04-26
add adamantium to test_app gemspecs

kbrock:
Looks like adamantium and rake were missing from the test_app's gemspec

I made these changes so I could:

```
cd test_app
export BUNDLE_GEMFILE=Gemfile.rspec3.2
bundle exec rake spec
bundle exec mutant -I lib -I spec --require test_app --use rspec TestApp::Literal
```


kbrock:
ugh - line numbers changes. ok. fixed

mbj:
@kbrock Why do you need / want to suppor the following for the test_app:

```
cd test_app
export BUNDLE_GEMFILE=Gemfile.rspec3.2
bundle exec rake spec
```

mbj:
Full +1 for adding adamantium as dependency, just curious for the need to add a Rakefile. If you'd had broken up the commits to only contain "one semantic change" (one for adding the missing adamantium dep, one for adding the Rakefile) I'd have cherry-picked your adamantium commit already to master.

Maybe its worth you break up the commits anyway. I like each commit does only one self-contained atomic change to the code base & builds on CI.

kbrock:
This is a test app. running rake spec and mutant in it should work fine.

pulled out that change and changed title

mbj:
> Is it worth adding another PR with Gemfile changes to add rake?

I personally never use rake on the test app. Will you?

PR #325, opened 2015-05-03
Fix expected coverage to use an approximate match

### LINKED ISSUE #323: Score rounding precision
I run mutant with the `--score 74.10` option.

I expect it to return the 0 exit code when the coverage is equal to that. However, due to the rounding precision the current coverage is shown as "74.10 %", even though the coverage is actually "927/1251" which is about "74,10071942446".

I'd like to have a way to express the expected coverage as the current coverage. When it's equal, I expect the exit code to be 0. When it's different (lower or higher) I expect the code to be 1.

dkubb:
This branch simplifies how expected coverage matches the actual coverage. It should fit people's mental model more closely, while not giving up too much precision.

The idea is that since Mutant reports coverage with 2 decimal places, this code makes sure that the comparison takes that into account. If the coverage is actually `74.333` but reported as `74.33` this will allow for people to specify `74.33` as the expected coverage and have it match. The code will match anything equal to or greater than `74.33` and less than `74.34` (but not equal to).

I realize this is a departure from the *exact* match method we've used, but it's not really that far off. it matches what how most people would expect things to work. Unless the printer is going to be changed to display the exact `Rational` state I think this, or something like it, is the best way forward for matching coverage.

[Fix #323]

### DIFF
--- a/lib/mutant/reporter/cli/printer.rb
+++ b/lib/mutant/reporter/cli/printer.rb
@@ -203,7 +203,7 @@ module Mutant
             info 'Mutant configuration:'
             info 'Matcher:         %s',      object.matcher.inspect
             info 'Integration:     %s',      object.integration.name
-            info 'Expect Coverage: %0.2f%%', (object.expected_coverage * 100)
+            info 'Expect Coverage: %0.2f%%', (object.expected_coverage * 100).floor(2)
             info 'Jobs:            %d',      object.jobs
             info 'Includes:        %s',      object.includes.inspect
             info 'Requires:        %s',      object.requires.inspect
@@ -244,8 +244,8 @@ module Mutant
             info 'Runtime:         %0.2fs',    runtime
             info 'Killtime:        %0.2fs',    killtime
             info 'Overhead:        %0.2f%%',   overhead_percent
-            status 'Coverage:        %0.2f%%', coverage_percent
-            status 'Expected:        %0.2f%%', (env.config.expected_coverage * 100)
+            status 'Coverage:        %0.2f%%', coverage_percent.floor(2)
+            status 'Expected:        %0.2f%%', (env.config.expected_coverage * 100).floor(2)
             self
           end
 
--- a/config/flog.yml
+++ b/config/flog.yml
@@ -1,2 +1,2 @@
 ---
-threshold: 30.9
+threshold: 33.2
--- a/lib/mutant/result.rb
+++ b/lib/mutant/result.rb
@@ -85,6 +85,11 @@ module Mutant
     class Env
       include Coverage, Result, Anima.new(:runtime, :env, :subject_results)
 
+      # Number of decimals to round the expected coverage successor for upper
+      # bounds comparison. The scale 2 was chosen to match the reported
+      # coverage amount.
+      ROUND_SCALE = 2
+
       # Test if run is successful
       #
       # @return [Boolean]
@@ -92,7 +97,9 @@ module Mutant
       # @api private
       #
       def success?
-        coverage.eql?(env.config.expected_coverage)
+        expected = env.config.expected_coverage
+        succ     = expected.floor(ROUND_SCALE) + 10**-ROUND_SCALE
+        (expected...succ).cover?(coverage)
       end
       memoize :success?
 
--- a/spec/unit/mutant/result/env_spec.rb
+++ b/spec/unit/mutant/result/env_spec.rb
@@ -19,7 +19,7 @@ RSpec.describe Mutant::Result::Env do
   let(:config) do
     double(
       'Config',
-      expected_coverage: Rational(1, 1)
+      expected_coverage: expected
     )
   end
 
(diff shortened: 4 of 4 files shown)


dkubb:
@mbj what do you think about this suggestion?

I couldn't find any direct unit tests for the printer. Nothing that tests the expected and actual coverage amounts at least. If there's some place where this is tested let me know and I'll add more specs for the printer changes I made that use `Rational#floor`.

dkubb:
@mbj one other point. I think this is probably a better approach than the one used in #324. It doesn't change the flags and (I think) matches what most people would expect.

mbj:
I think we should allow both, the explicit coverage expectation, and a "fuzzy apprach". A flag like `--expected-coverage-within` could determine "how far off" the coverage can be within the `--expected-coverage`.

But I personally do not see myself using that flag. Because its less exact.

mbj:
I like the tools I build to fulfill my needs, and not the needs of an artificial "what most users expect", because this leads me into the rabbit hole of combinatoric explosions based on my imagination of "most people".

Right now I could convince any mutant user I know why explicit matches are needed, latest example here http://blog.arkency.com/2015/05/mutation-testing-and-continuous-integration/.

I wounder if most people can be convinced to follow the explicit approach? Before adding code weight to support a fuzzy match.

If I compare missed expectations to a failing test: In case a test fails, one I do not really care about I typically delete that test, or set it to pending to fix later. The equivalent would be to ignore that subject in mutant. IMO this is the best appraoch.

Before doing a decision, lets discuss a bit more.

dkubb:
> I like the tools I build to fulfill my needs, and not the needs of an artificial "what most users expect"

Generally I would agree with you, but this isn't artificial. It's bitten you and I before, more than once. It's also been reported by other users. I think it is a valid usability problem.

I do agree that whitelisting failing classes and then setting the threshold to 100% and then working on removing classes from the whitelist is a long-term better approach, but it's harder to explain. What I was hoping for is a better out-of-box experience for first time users; it should be easy enough to do the right thing before frustrating them with a usability problem that is a side effect of the internal implementation.

Would you consider reporting the errors with the actual fractions *by default* so that people can use those in their threshold specifications? I would posit that will be an odd UI for the same reasons as requiring a fraction to specify the command-line threshold.

dkubb:
As an alternative idea for implementation, there must be a way we can define the expected coverage so that it would *always* result in an exactly precisely precise match -- one extra pass or failure would be excluded.

We know the total number of mutations. We know how many pass and fails there are. We also could calculate what the percentage would be for one more pass or one more failure. We could then define the success method as:

```ruby
def succes?
  expected = env.config.expected_coverage
  coverage > min(expected) && coverage < max(expected)
end
```

The hypothetical `#min` and `#max` methods (we'll need better names) would return the previous increment before, and the next increment after, the expected threshold. The *only* possible coverage amount that could match in between this range is the one that we are looking for.

This would give us *perfect* precision in every case, giving you the behaviour you want, while also being more usable to myself and others who have been bitten by this bug before.

NOTE: this may require a bit more real-world tweaking to work with cases where we are at `0%` or `100%` coverage, but I think those are easy to work-around by returning `-Float::INFINITY` from `#min` when coverage is `0%` and `Float::INFINITY` from `#max` when coverage is `100%`.

lib/mutant/reporter/cli/printer.rb
```diff
@@ -258,7 +258,7 @@ def run
           # @api private
           #
           def coverage_percent
-            (coverage * 100).to_f
+            coverage * 100
```
dkubb: @mbj if you want I could extract this into it's own commit, since the extra `#to_f` is not needed in either case.

mbj: Yeah, please do so. Push that one to master please.

mbj:
> As an alternative idea for implementation, there must be a way we can define the expected coverage so that it would always result in an exactly perfect precise match -- one extra pass or failure would be excluded.

What about adding a 2nd flag? Right now we have `--expect-coverage`, if you pass in `--expect-coverage-2nd-option-well-named` that allows to set the upper bound of a "expected coverage range". `--expect-coverage` would set the lower bound. By default that 2nd argument would default to the value given for `--expect-coverage`. - Maybe I'm just summing your ideas in my words here.

I'd be open for that approach, the default would be "exact", but when you want to allow "more than specified-coverage" you are fine in setting that 2nd flag?

mbj:
BTW with a fine grained configuration (file) mechanism that allows to set coverage expectations per subject (group) we maybe do not need all this anymore.

mbj:
@dkubb Master meanwhile got the feature to place "arbitrary precision" coverage expectations via the `--expected-coverage` flag that now also takes the `113/1123` rational syntax as input. I think that approximate matches are not needed anymore. In my commercial and opensource projects I never had the need for them.

mbj:
@dkubb I'll close this as the PR does not apply anymore. And arbitrary precision coverage expectation is in master. Also we do not have a use case for approximated coverage expectations.

PR #340, opened 2015-06-12
Remove useless quotes

nicolasleger:


mbj:
IMO these are not useless. I do not want the values to be interpreted as Floats, than being serialized to Strings again when Circle passes the values to the CLI.

mbj:
@nicolasleger BTW even when this one was not merged, feel free to discuss, and or come up with more ;)

nicolasleger:
I proposed to modify the `.travis.yml` as everyone does that way, not the `circle.yml` ^^

mbj:
> I proposed to modify the .travis.yml

Thats true, it slipped by me. Still I think we should avoid implicit conversions, and not rely on `Float#to_s` equivalent travis will do at a certain point to move the float back to the CLI invocation of RVM.

>  as everyone does that way,

Everyone would include me. Also everyone (excluding me) could be wrong. The argument "everyone does $foo" never works IMO. "everyone does $foo because of $convincing_technical_argument" would make me think.

nicolasleger:
Yes Travis does its own mapping (not always updated !) to use patched versions but this is the way they use in their examples at http://docs.travis-ci.com/user/languages/ruby/.
I hesitated on the argument of "everyone does", I should have quoted Travis examples directly.

It is up to you, it was just for styling ! :cat: Travis still understand the string way but I think we should use the recommended way IMO.

mbj:
Travis delegates to rvm, and rvm interprets a string like `'2.2'` to use the latest installed `2.2.x` version. Thats what I want. I do not use the "travis mappnigs" for `jruby-2.0-mode` etc.

PR #346, opened 2015-06-13
Support rspec 3.3

mbj:
Adds support for rspec 3.3

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,6 +3,8 @@
 * Drop support for ruby < 2.1
 * Remove broken `--code` option
 * Remove deprecated `--score` option
+* Add support for rspec-3.3
+* End support for rspec-3.{0,1}
 
 # v0.7.9 2015-05-30
 
--- a/mutant-rspec.gemspec
+++ b/mutant-rspec.gemspec
@@ -18,7 +18,7 @@ Gem::Specification.new do |gem|
   gem.extra_rdoc_files = %w[TODO LICENSE]
 
   gem.add_runtime_dependency('mutant', "~> #{gem.version}")
-  gem.add_runtime_dependency('rspec-core', '>= 3.0.0', '< 3.4.0')
+  gem.add_runtime_dependency('rspec-core', '>= 3.2.0', '< 3.4.0')
 
   gem.add_development_dependency('bundler', '~> 1.3', '>= 1.3.5')
 end
--- a/mutant-rspec.gemspec
+++ b/mutant-rspec.gemspec
@@ -18,7 +18,7 @@ Gem::Specification.new do |gem|
   gem.extra_rdoc_files = %w[TODO LICENSE]
 
   gem.add_runtime_dependency('mutant', "~> #{gem.version}")
-  gem.add_runtime_dependency('rspec-core', '>= 3.0.0', '< 3.3.0')
+  gem.add_runtime_dependency('rspec-core', '>= 3.0.0', '< 3.4.0')
 
   gem.add_development_dependency('bundler', '~> 1.3', '>= 1.3.5')
 end
--- a/spec/integration/mutant/rspec_spec.rb
+++ b/spec/integration/mutant/rspec_spec.rb
@@ -2,21 +2,11 @@ RSpec.describe 'rspec integration', mutant: false do
 
   let(:base_cmd) { 'bundle exec mutant -I lib --require test_app --use rspec' }
 
-  context 'RSpec 3.0' do
-    let(:gemfile) { 'Gemfile.rspec3.0' }
+  %w[3.0 3.1 3.2 3.3].each do |version|
+    context "RSpec #{version}" do
+      let(:gemfile) { "Gemfile.rspec#{version}" }
 
-    it_behaves_like 'framework integration'
-  end
-
-  context 'RSpec 3.1' do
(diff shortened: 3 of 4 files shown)


mbj:
@solnic, @snusnu, @dkubb, @splattael, @tjchambers, @plexus Would be cool to get feedback if this works for your projects that are on rspec-3.3 already.

dkubb:
@mbj we should make a experimental branch for our work project and try this out. I can try it out on axiom and some of my other projects that use mutant.

tjchambers:
@mbj Not yet on 3.3 and had gem conflicts to be resolved when I went to 3.2. Will give it a try and see what I can accomplish.

solnic:
works fine with rom

snusnu:
not yet on 3.3 either ...

mbj:
> works fine with rom

Thats cool. I'll check our work project too, and my `auom` gem thats the most integration tested gem against mutant. Than will release.

dkubb:
It works with axiom-types.

tjchambers:
@mbj I skipped to 3.4.0.pre as I battle gem dependency issues, but that appears to work as well with 0.8.0 of mutant.

mbj:
> @mbj I skipped to 3.4.0.pre as I battle gem dependency issues, but that appears to work as well with 0.8.0 of mutant.

That should not possibly work because of: https://github.com/mbj/mutant/blob/master/mutant-rspec.gemspec#L21 that explicitly refutes `3.4.0`. Or are `.pre` versions not affected by `>` operator?

tjchambers:
@mbj - let me double check that I am running what I think I am. My bundle install shows:

```
Using rspec-support 3.4.0.pre from git://github.com/rspec/rspec-support.git (at master)
Using rspec-core 3.4.0.pre from git://github.com/rspec/rspec-core.git (at master)
Using mutant 0.8.0 from git://github.com/mbj/mutant.git (at master)
Using mutant-rspec 0.8.0 from git://github.com/mbj/mutant.git (at master)
```

and my command was:

```
RAILS_ENV=test bundle exec mutant -r ./config/environment.rb -j1 --use rspec -I app/models/align/analytics/pairs/theme_guba_category_pair.rb ThemeGubaCategoryPair > ./log/mutant/ThemeGubaCategoryPair.log
```

but let me see if I can confirm.

tjchambers:
@mbj - asking versions directly:

```
~/projects/aot/aotv2@localmbr [09:41:29]> mutant -v
mutant-0.8.0
~/projects/aot/aotv2@localmbr [09:41:36]> rspec -v
3.4.0.pre
```

I don't see anything in the responses so far to indicate I am NOT using 3.4.0.pre.

tjchambers:
@mbj just to be clear I have not altered any specs or config to use any 3.3 ONLY RSpec capabilities.

mbj:
@tjchambers I think that rubygems version comparision operator is true for `3.4.0.pre < 3.4.0`, that would be awkward and is probably a bug.

mbj:
But this AFAIK does not indicate a mutant problem, so I'm moving on.

tjchambers:
Concur on both counts. I will let you know if I do find issues with 3.4.0.pre + mutant 0.8.0 but I will not expect it to be supported at this juncture.

mbj:
@tjchambers Dont you use `bundle exec`? It could be the case you mixed rspec-3.4.0.pre with mutant-0.8.0 (which should not be possible) because you do not use bundler correctly? (This could be the source for other dependency issues you mentioned above).

tjchambers:
@mbj yes - I included my bundle exec oriented command above. Not sure how I could as you put it "not use bundler" correctly. No offense intended. I do recognize my -v commands were outside the bundle.

mbj:
@tjchambers To verify whats actually bundled you can use `bundle show $gem`. Using `-v` outside of bundle is not giving you a definitive result. I just was thinking out loud what could be the problem you ended up with rspec-3.4 where `mutant-rspec.gemspec` explicitly forbids it.

tjchambers:
@mbj - I do recognize that was outside bundle, which is why I included the responses from bundle install above as I believe they are definitive. SO here is bundle show gem snippet:

```
  * mutant (0.8.0 8e29780)
  * mutant-rspec (0.8.0 8e29780)
  * rspec-collection_matchers (1.1.2)
  * rspec-core (3.4.0.pre aaf2964)
  * rspec-expectations (3.4.0.pre d03a69b)
  * rspec-mocks (3.4.0.pre f7837f5)
  * rspec-rails (3.4.0.pre 5ee901f)
  * rspec-support (3.4.0.pre 400b69d)
```

tjchambers:
My main gem conflict has been with guard-rspec which is IMHO not specifying the rspec dependency in gemspec as it should. So guard-rspec does not like later rspec core (because of incorrect dependency spec) and so I have to work around that. 

That is distinct from the mutant/rspec dependency issue issue we are discussing. So IMHO either there is a specification issue or a bug (I am leaning towards bundler bug).

mbj:
So it for real looks like a bundler / rubygems bug.

tjchambers:
As I said earlier I would concur with your assessment. My action is if I want to report an issue in mutant I have to reproduce on 3.3 RSpec or my bad. You can choose to try and work around the dependency specification "bug" or ignore it and move on. I am good with whatever and will NOT report a bug against 3.4.0.pre. That's my commitment to mutant.

PR #350, opened 2015-06-13
Simplify Rails setup instructions

andyw8:
As `mutant` is already specified as a dependency in `mutant-rspec .gemspec`, it seems unnecessary to list it in the Gemfile.


mbj:
True. I intend to reduce that references but failed. Thx.

PR #351, opened 2015-06-13
Fix specification of Mutant::AST namespace

mbj:


mbj:

lib/mutant/ast/meta.rb
```diff
@@ -81,7 +44,8 @@ def assignment?
         # @api private
         #
         def attribute_assignment?
-          arguments.one? && attribute_assignment_selector?
+          !Types::METHOD_OPERATORS.include?(selector) &&
+          selector.to_s.end_with?(ATTRIBUTE_ASSIGNMENT_SELECTOR_SUFFIX)
```
dkubb: @mbj selector can be nil?

lib/mutant/ast/meta.rb
```diff
@@ -101,49 +65,10 @@ def index_assignment?
...
-        # @api private
-        #
-        def assignment?
-          Types::ASSIGNABLE_VARIABLES.include?(node.type)
-        end
-
-      end # Generic
-
-    end #
+    end # MEta
```
dkubb: I think you meant "e" here.

lib/mutant/ast/named_children.rb
```diff
@@ -13,8 +13,6 @@ module NamedChildren
       # @api private
       #
       def self.included(host)
-        super
```
dkubb: The main reason we used `super` before is if `Module#included` defined some functionality. I realize it doesn't (usually, unless you are using something crazy like AS) but it could.

Overriding the `.included` method is no different from overriding some other "hook" method like `#initialize`. Should we assume when we are overriding that we *know* the precise code in the parent method and make calling `super` optional when it is not needed?

lib/mutant/ast/meta.rb
dkubb: @mbj if so, is there a test for this? I didn't see anything when I looked for it in this diff (it could very well be covered elsewhere by an earlier commit)

dkubb:
@mbj I did a pass over this code.

lib/mutant/ast/meta.rb
mbj: @dkubb The selector is a `Symbol` instance. And only strings have `#end_with?`.

lib/mutant/ast/named_children.rb
mbj: I could not come up with a test that kills this mutation and gave up. As I control the environment here and `NamedChildren` will not be called outside of mutant I'm fine to fix this when it becomes a problem. If I'd provide an public interface where I ask my users to include a module of mine that defines `.included` I'd keep it.

lib/mutant/ast/meta.rb
dkubb: @mbj is the number of possible selectors known up-front, or is it dynamic? If it's know up-front, maybe it would be better to do `ATTRIBUTE_ASSIGNMENT_SELECTORS.include?(selector)` instead of inferring it's suitability by it's string representation?

mbj: You missed the `_SUFFIX` part in this constant name. The constant evaluates to `=` and the prefix is chosen by the ruby user. I can define any prefix like: `def foo=(value); end` etc.

dkubb: @mbj ahh ok, I get what it is now. What you're doing makes sense.

lib/mutant/ast/named_children.rb
dkubb: I am actually fine with just saying that if you know what the parent class method is doing that you can decide to omit `super` if it does not break LSP. I don't think we should use `super` "just in case" the parent class decides to change it's interface.

lib/mutant/ast/meta.rb
mbj: All these guards are needed to not mutate code like:

```
foo.bar = baz
```

```
(send (send nil :foo) :bar= (send nil :baz))
```

into an (invalid) AST that would unparse (unparser trusts the AST 100%) into:

```
bar = baz
```
(send nil :bar= (send nil :baz))

The AST in this case would say call `#bar=` on implicit self. But the ruby parser would say: Assign the lvar `bar=`. I see how this could be a cool mutation, but if we do this: It should be explicit, and not an artifact of a transform to invalid AST.

Valid AST would, for this case be:

```
(lvasgn :bar (send nil :baz))
```

mbj: There are other cases like this. I do not want to mutate into invalid AST, but interpreted as source via unparser, valid source. Even when semantics differ. Such "implicit" mutations should be made explicit.

mbj:

spec/unit/mutant/ast/named_children_spec.rb
```diff
@@ -0,0 +1,52 @@
+RSpec.describe Mutant::AST::NamedChildren do
+  describe '.included' do
+    let(:klass) do
+      Class.new do
+        include Mutant::AST::NamedChildren, Concord.new(:node)
+
+        children :foo, :bar
+      end
+
```
dkubb: Minor, extra blank line.

dkubb:
@mbj I am going to merge this. I found one place where there's an extra blank line, but it's not serious at all, so shouldn't block this merge.

spec/unit/mutant/ast/named_children_spec.rb
mbj: Fixed. thx.

PR #359, opened 2015-06-23
Improve expression parsing

mbj:


lib/mutant/expression/parser.rb
```diff
@@ -0,0 +1,75 @@
...
+      #
+      # @return [nil]
+      #   otherwise
+      #
+      # @api private
+      #
+      def try_parse(input)
+        expressions = expressions(input)
+        case expressions.length
+        when 0
```
dkubb: You could use `when 0, 1` here.

mbj: k, good catch.

mbj:

This change was verified against:

```
bundle exec mutant --zombie -I lib -r mutant --use rspec -- 'Mutant::Expression*'
```

Mutation covered subjects outside the expression namespace where also confirmed to *still* be mutation covered. Subjects that where touched that do not have yet mutation coverage where intentionally not covered in this PR. Most mutation coverage operations yield refactorings that would grow to big.

My plain is to continue with other 2nd level namespaces till mutant is fully self covered. Than do feature development again.

mbj:
@dkubb Please do not review. My latest addition of mutation covering specs showed are reduction of line coverage in other subjects.

This can happen because the now not perfectly line covered subjects had been "implicitly" covered by the implementation of *another* subjects specs.

Another reason I'm vehemently opposing against implicit coverage scenarios with mutation testing, implicit coverage already causes problem on line coverage based metrics.

mbj:
@dkubb Ready for review again.

config/reek.yml
```diff
@@ -65,6 +65,7 @@ TooManyInstanceVariables:
   exclude:
   - Mutant::Mutator # 4 vars
   - Mutant::Parallel::Master # 4 vars
+  - Mutant::Integration::Rspec # 4 vars
```
dkubb: I like that you give a small reason for the exception.

mbj: Yeah, ideally reek (just like rubocop and others) would not only allow me to add a whitelist, but allow me to specify the whitelisted value.

Right now those exceptions could grow to 1k of ivars without being constrainted.

lib/mutant/cli.rb
```diff
@@ -86,7 +86,7 @@ def parse(arguments)
     def parse_match_expressions(expressions)
       fail Error, 'No expressions given' if expressions.empty?
 
-      expressions.map(&Expression.method(:parse)).each do |expression|
+      expressions.map(&config.expression_parser.method(:call)).each do |expression|
```
dkubb: Instead of unnecessarily creating an intermediary array, why not do this:

```ruby
expressions.each do |expression|
  add_matcher(:match_expressions, config.expression_parser.call(expression))
end
```

mbj: Yeah, I like that. I wish we'd have first class function composition to make this thing a one liner.

lib/mutant/expression/method.rb
```diff
@@ -3,44 +3,44 @@ class Expression
...
       MATCHERS = IceNine.deep_freeze(
         '.' => Matcher::Methods::Singleton,
         '#' => Matcher::Methods::Instance
       )
+      private_constant(*constants(false))
 
-      register(
-        /\A(?<scope_name>#{SCOPE_PATTERN})(?<scope_symbol>[.#])(?<method_name>#{METHOD_NAME_PATTERN})\z/
-      )
+      REGEXP = /\A(?<scope_name>#{SCOPE_PATTERN})(?<scope_symbol>[.#])(?<method_name>#{METHOD_NAME_PATTERN})\z/.freeze
```
dkubb: This is a bit long. Have you considered using `/x` and making it multi-line? (or assembling it from variables)

mbj: Good catch. Will do that.

lib/mutant/expression/method.rb
```diff
@@ -3,44 +3,44 @@ class Expression
...
       # @api private
       #
       def matcher(env)
         methods_matcher = MATCHERS.fetch(scope_symbol).new(env, scope)
-        method = methods_matcher.methods.detect do |meth|
-          meth.name.equal?(method_name)
-        end or fail NameError, "Cannot find method #{method_name}"
-        methods_matcher.matcher.build(env, scope, method)
+
+        Matcher::Filter.new(methods_matcher, ->(subject) { subject.expression.eql?(self) })
```
dkubb: Would it make sense for `Matcher::Filter.new` to accept a block instead of the last argument being a proc?

lib/mutant/expression/method.rb
dkubb: Assembling from variables may not be a bad idea though, since it looks like some parts of the regexp, such as `(?<scope_name>#{SCOPE_PATTERN})` are the same in a few subclasses.

lib/mutant/expression/namespace.rb
```diff
@@ -2,24 +2,12 @@ module Mutant
...
-        match[__method__]
-      end
+      include AbstractType, Anima.new(:namespace)
+      private(*anima.attribute_names)
 
       # Recursive namespace expression
       class Recursive < self
-
-        register(/\A(?<namespace>#{SCOPE_PATTERN})?\*\z/)
+        REGEXP = /\A(?<namespace>#{SCOPE_PATTERN})?\*\z/.freeze
```
dkubb: Why not name this `<scope_name>` or name the instances of `scope_name` as `namespace` ? This might allow you to share some parts of the regexp between subclasses.

lib/mutant/matcher/filter.rb
```diff
@@ -16,7 +16,8 @@ class Filter < self
       #
       def each(&block)
         return to_enum unless block_given?
-        matcher.select(&predicate.method(:call)).each(&block)
+        subjects = matcher.select(&predicate.method(:call))
+        subjects.each(&block)
```
dkubb: Why assign this to a synthetic variable that is only used on the next line?

lib/mutant/expression/method.rb
mbj: I'll think about that when I get to the `Mutant::Matcher` namespace. 

Personally I prefer to have a "explicit" composition of a matcher and a predicate. We could have a `.build` method that captures the block. But as I have another call side for this where an explicit predicate gets constructed (for more complex selections / rejectsions) I think this would be overkill.

I think that only "DSL" methods should capture blocks, and everything else is better of with an simple traditional explicit argument.

Blocks as implemented in ruby are "just" a 2nd dimension in "argument complexity". APIs that mix both approaches are not easy to understand.

The more I work with sane languages the more I think rubies blocks are just a "bad implemented lambda" ;)

lib/mutant/subject/method.rb
```diff
@@ -13,12 +13,12 @@ class Method < self
...
       # Return method name
       #
-      # @return [Symbol]
+      # @return [Expression]
       #
       # @api private
       #
       def name
-        node.children[self.class::NAME_INDEX]
+        node.children.at(self.class::NAME_INDEX)
```
dkubb: Is there a mutation for `#[]` to `#at` that showed you this? I wonder if there are other places in the codebase that could use `#at` instead.

dkubb:
@mbj I finished a first pass over this code.

lib/mutant/subject/method.rb
mbj: There is no such a mutation. Easy to add, please open a ticket. But I think we actually should use `Array#fetch` here. As as `#at` defaults to `nil` and we know the name is always present.

Ideally we mutate `#[]` to `#fetch` and `#at`.

Also this change is not related at all to the subject of this PR, I'll move it to another iteration.

lib/mutant/matcher/filter.rb
mbj: Probably a refactoring / debugging leftover.

lib/mutant/expression/namespace.rb
mbj: I had the same idea early in this PR development. But postponed it for "lets do that later". Now as the PR grew in scope I think it should be part of this iteration. Good catch.

lib/mutant.rb
```diff
@@ -192,7 +193,13 @@ class Config
       reporter:          Reporter::CLI.build($stdout),
       zombie:            false,
       jobs:              Mutant.ci? ? CI_DEFAULT_PROCESSOR_COUNT : ::Parallel.processor_count,
-      expected_coverage: Rational(1)
+      expected_coverage: Rational(1),
+      expression_parser: Expression::Parser.new([
+        Expression::Method,
+        Expression::Methods,
+        Expression::Namespace::Exact,
+        Expression::Namespace::Recursive
```
mbj: @dkubb It was not part of the PRs description, but this opens the ability to add mutant extensions that need their own expressions such maybe as `erb:logical/name.html.erb`. On bootstrapping / CLI parsing an extension could inject extension specific parsers.

mbj:
@dkubb Assuming all green, this is ready for your next pass.

lib/mutant/env/bootstrap.rb
```diff
@@ -106,6 +106,31 @@ def infect
...
       end
 
+      # Return matched subjects
+      #
+      # @return [Enumerable<Subject>]
+      #
+      # @api private
+      #
+      def matched_subjects
+        Matcher::Compiler.call(self, config.matcher).to_a
```
dkubb: Not that it matters that much, but this collection is coerced to an Array. Is there a need for it to be an Array? If the object is properly enumerable (I assume it is and you didn't just define `#to_a` on it directly) can't that just be passed around?

If for some reason an Array is required, maybe the YARD doc should be updated to reflect that.

lib/mutant/expression.rb
```diff
@@ -2,83 +2,21 @@ module Mutant
 
   # Abstract base class for match expression
   class Expression
-    include AbstractType, Adamantium::Flat, Concord::Public.new(:match)
+    include AbstractType, Adamantium::Flat
 
-    include Equalizer.new(:syntax)
+    fragment             = /[A-Za-z][A-Za-z\d_]*/.freeze
```
dkubb: I know we talked about this as a way to define truly "private" constants, but it still seems weird to me.

lib/mutant/expression/method.rb
```diff
@@ -3,44 +3,50 @@ class Expression
...
       MATCHERS = IceNine.deep_freeze(
         '.' => Matcher::Methods::Singleton,
         '#' => Matcher::Methods::Instance
       )
 
-      register(
-        /\A(?<scope_name>#{SCOPE_PATTERN})(?<scope_symbol>[.#])(?<method_name>#{METHOD_NAME_PATTERN})\z/
-      )
+      METHOD_NAME_PATTERN = Regexp.union(
+        /(?<method_name>[A-Za-z_][A-Za-z\d_]*[!?=]?)/,
```
dkubb: Do you plan on supporting unicode method names in the future? Is the plan for right now to just restrict this as much as possible and in a future PR expand this to include all valid ruby method names?

mbj: > Do you plan on supporting unicode method names in the future?

Yeah, but thats more complex than I initially thought.

> Is the plan for right now to just restrict this as much as possible and in a future PR expand this to include all valid ruby method names?

My plan is to implement expression parsing semantically equivalent to the previous implementation, but fully mutation tested.

I plan a followup that allows all valid literal ruby method names parsed as expressions. Please note that through the changed inner workings generating expressions from matched subjects already works with any valid ruby method names, as the need to parse expressions from subjects was removed.

lib/mutant/expression/method.rb
dkubb: > I think that only "DSL" methods should capture blocks, and everything else is better of with an simple traditional explicit argument.

I don't necessarily agree with this. I think if the language provides the construct, provided there's nothing inherently wrong with it (unlike some JS "features"), I think it's better to use a more idiomatic approach when possible.

The block mechanism in ruby is well understood, and it's probably not idiomatic to create a proc and pass it in as the last argument in a method call. The only reason I might consider passing it in as an explicit argument might be if it's a richer object that supports `#call`, and I don't want to reduce it to something more primitive like a `Proc`.

lib/mutant/expression/namespace.rb
```diff
@@ -2,24 +2,12 @@ module Mutant
...
-        match[__method__]
-      end
+      include AbstractType, Anima.new(:scope_name)
+      private(*anima.attribute_names)
 
       # Recursive namespace expression
       class Recursive < self
-
-        register(/\A(?<namespace>#{SCOPE_PATTERN})?\*\z/)
+        REGEXP = /\A(?:#{SCOPE_NAME_PATTERN})?\*\z/.freeze
```
dkubb: What about defining `SCOPE_NAME_PATTERN` so it is wrapped in `(?:` and `)` ? Then it can be treated as an opaque matching pattern without knowing what it's contents are. You can then just use `#{SCOPE_NAME_PATTERN}?`.

lib/mutant/expression.rb
mbj: This is just an lvar used to make the definition of `SCOPE_NAME_PATTERN` more readable.

I do not need to access the `fragment` outside this scope anywhere, so it does not make sense to assign it to a constant that would be accessible in other scopes.

lib/mutant/expression/parser.rb
```diff
@@ -0,0 +1,74 @@
...
+      #
+      # @return [Array<Expression>]
+      #   if expressions can be parsed from input
+      #
+      # @api private
+      #
+      def expressions(input)
+        types.each_with_object([]) do |type, aggregate|
+          expression = type.try_parse(input)
+          aggregate << expression if expression
```
dkubb: This is totally OT, but I feel like there should be an enumerable method that works like `Enumerable#select`, but returns the "truthy" return values from the block like `#map`. This would eliminate the need for using `#each_with_object` like this.

lib/mutant/integration/rspec.rb
```diff
@@ -35,6 +37,7 @@ class Rspec < self
       #
       def initialize(*)
         super
+        @all_expression = expression_parser.call(ALL_EXPRESSION)
```
dkubb: It looks like in some places you use `expression_parser.(...)` and in some places you use `expression_parser.call(...)`.

I won't flag all of them, but I wanted to bring it to your attention. I'm wondering if there are other callsites for other objects where one or the other can be used too.

lib/mutant/subject/method.rb
```diff
@@ -28,7 +28,11 @@ def name
       # @api private
       #
       def expression
-        Expression.parse("#{context.identification}#{self.class::SYMBOL}#{name}")
+        Expression::Method.new(
+          scope_symbol: self.class::SYMBOL,
+          scope_name:   scope.name,
+          method_name:  name.to_s
```
dkubb: Why prefer a string for a method name? Wouldn't a `Symbol` be a simpler primitive?

spec/unit/mutant/expression/method_spec.rb
```diff
@@ -30,19 +29,16 @@
...
+        expect(subject.to_a.map(&:expression)).to eql([object])
       end
     end
 
     context 'with a singleton method' do
       let(:input) { singleton_method }
 
-      it { should eql(Mutant::Matcher::Method::Singleton.new(env, TestApp::Literal, TestApp::Literal.method(:string))) }
+      it 'returns correct matcher' do
+        expect(subject.to_a.map(&:expression)).to eql([object])
```
dkubb: Is it necessary to call `#to_a` and then `#map` here? If it is enumerable then `#map` should be supported natively.

lib/mutant/env/bootstrap.rb
mbj: The evaluation to `Array` via `#to_a` is needed to force evaluation of all subjects at this point, why we need to force:

 * Evaluation of matchers to subjects can trigger warnings, and I want all warnings to be printed before analysis starts
* When we lazily match later on, the object space might be "infected" with zillions of rspec classes, making matching very slow.

lib/mutant/expression/parser.rb
mbj: Yeah, I'd love that.

dkubb:
@mbj I left a few small notes on this PR.

Also, you may want to rebase this on top of master since I merged another simple PR earlier.

lib/mutant/integration/rspec.rb
mbj: I think I want to always used the shortest form. I'll fix the `.call` ones to be `.`. Good catch.

lib/mutant/subject/method.rb
mbj: Symmetry with parsing that produces string `method_name` values.

spec/unit/mutant/expression/method_spec.rb
mbj: Refactoring leftover, good catch.

lib/mutant/expression/method.rb
mbj: @dkubb I think I was not presenting my arguments consistent enough:

My primary argument are:

* Capturing blocks should be done in dedicated methods, methods like `Class#new` should only in rare cases capture a block, as the typical `#new` implementation simply should take the arguments to initialize an instance.
* Ive another call side for `#new` that relies on being able to pass the argument in non block form

So I added a `.build` method that does capture the block  and uses the new method.

lib/mutant/expression/namespace.rb
mbj: It was wrapped in a capturing group already. So the additional wrapping in a non captured group to apply `?` operator to was redundant. Good catch. I could drop the additional non capturing group.

mbj:

lib/mutant/expression/method.rb
dkubb: Ok, I can see that POV. The build allows you to use a new block, but an explicit argument allows you to retain the object state if it's callable, but something other than `Proc`.

lib/mutant/expression/parser.rb
dkubb: Interestingly there was a feature proposal for this a few years ago: https://bugs.ruby-lang.org/issues/5663

lib/mutant/expression/method.rb
mbj: > Ok, I can see that POV. The build allows you to use a new block, but an explicit argument allows you to retain the object state if it's callable, but something other than Proc.

Very good description of my thought.

PR #360, opened 2015-06-29
Change integrations to reference config object

mbj:
* Future changes depend on this one
* Also fixes the smell of "stateless" integration objects we had in the
  past

### DIFF
--- a/lib/mutant/integration/rspec.rb
+++ b/lib/mutant/integration/rspec.rb
@@ -19,7 +19,7 @@ module Mutant
     #   for unique reference.
     class Rspec < self
 
-      ALL                  = Mutant::Expression.parse('*')
+      ALL_EXPRESSION       = Expression.parse('*').freeze
       EXPRESSION_CANDIDATE = /\A([^ ]+)(?: )?/.freeze
       LOCATION_DELIMITER   = ':'.freeze
       EXIT_SUCCESS         = 0
@@ -33,7 +33,8 @@ module Mutant
       #
       # @api private
       #
-      def initialize
+      def initialize(*)
+        super
         @output = StringIO.new
         @runner = RSpec::Core::Runner.new(RSpec::Core::ConfigurationOptions.new(CLI_OPTIONS))
         @world  = RSpec.world
@@ -134,7 +135,7 @@ module Mutant
           Expression.parse(metadata.fetch(:mutant_expression))
         else
           match = EXPRESSION_CANDIDATE.match(metadata.fetch(:full_description))
-          Expression.try_parse(match.captures.first) || ALL
+          Expression.try_parse(match.captures.first) || ALL_EXPRESSION
         end
       end
 
--- a/spec/unit/mutant/env_spec.rb
+++ b/spec/unit/mutant/env_spec.rb
@@ -8,18 +8,15 @@ RSpec.describe Mutant::Env do
         selector:         selector,
         subjects:         [],
         mutations:        [],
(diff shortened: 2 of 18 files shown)


mbj:
@dkubb This is an extraction from unfinished #359, and assuming all green its ready for review.

spec/unit/mutant/env_spec.rb
```diff
@@ -8,18 +8,15 @@
...
+        integration:      Mutant::Integration::Null.new(config)
       )
     end
 
     let(:config) do
-      Mutant::Config::DEFAULT.update(
-        isolation:   isolation,
-        integration: integration
-      )
+      Mutant::Config::DEFAULT.update(isolation:   isolation)
```
dkubb: Looks like a tiny bit of extra whitespace here.

lib/mutant/integration/rspec.rb
```diff
@@ -33,7 +33,8 @@ class Rspec < self
       #
       # @api private
       #
-      def initialize
+      def initialize(*args)
```
dkubb: Is it necessary to use `*args` here? Can you just use `*`, and then use `super` without parenthesis?

spec/unit/mutant/integration/rspec_spec.rb
```diff
@@ -1,10 +1,10 @@
 require 'mutant/integration/rspec'
 
 RSpec.describe Mutant::Integration::Rspec do
-  let(:object) { described_class.new }
+  let(:object) { described_class.new(Mutant::Config::DEFAULT) }
 
-  let(:options)       { double('options')                           }
-  let(:runner)        { double('runner')                            }
+  let(:options) { double('options') }
```
dkubb: What do you think about using `instance_double` instead? It has stronger constraints about how it can be stubbed.

lib/mutant/integration/rspec.rb
mbj: Refactoring leftover. Good catch. Yeah we can use zsuper.

dkubb:
@mbj are these changes mutation covered?

I finished a first pass over the code.

mbj:
> @mbj are these changes mutation covered?

Only the changes that are touching subjects that had been mutation covered before this change.

My current focus is to get mutant fully mutation covered. This change is a prerequesite to mutation cover the `Mutant::Expression*` namespace fully

Mutation coverage triggers refactorings, so as a circuit breaker I decided to NOT fix the coverage of subjects that did not had mutation coverage before this change.

spec/unit/mutant/integration/rspec_spec.rb
mbj: I want to introduce `instance_double` in a global refactoring. Right now mutant consistently uses the weaker `double` form. To keep this consistency lets continue with `double` for this PR.

mbj:
> I finished a first pass over the code.

@dkubb Ready for the next pass. Flags are addressed by change for reply.

dkubb:
@mbj what is the best way to test this feature?

mbj:
> @mbj what is the best way to test this feature?

Its a pure internal refactoring, no visible semantic changes. Its an extraction from another PR to keep the diff small. The other PR introduces the concept of an "expression parser" object the config injects everywhere its passed in.

spec/unit/mutant/integration/rspec_spec.rb
dkubb: :+1:

PR #372, opened 2015-07-06
Use instance_double where appropriate

### LINKED ISSUE #361: Switch to `instance_double` from `double`.
A global switch to `instance_double` for all uses of `double` is needed as the former form provides stronger guarantees.

mbj:
[fix #361]

### DIFF
--- a/spec/unit/mutant/actor/binding_spec.rb
+++ b/spec/unit/mutant/actor/binding_spec.rb
@@ -1,12 +1,13 @@
 RSpec.describe Mutant::Actor::Binding do
-  let(:actor_a)    { double('Actor-A', sender: sender_a, receiver: receiver_a) }
-  let(:sender_a)   { double('Sender-A')                                        }
-  let(:sender_b)   { double('Sender-B')                                        }
-  let(:receiver_a) { double('Receiver-A')                                      }
-  let(:payload)    { double('Payload')                                         }
-  let(:type)       { double('Type')                                            }
+  let_instance(:sender_a,   Mutant::Actor::Sender                                           )
+  let_instance(:sender_b,   Mutant::Actor::Sender                                           )
+  let_instance(:mailbox,    Mutant::Actor::Mailbox, receiver: :receiver_a, sender: :sender_a)
+  let_instance(:receiver_a, Mutant::Actor::Receiver                                         )
 
-  let(:object) { described_class.new(actor_a, sender_b) }
+  let_anon(:payload)
+  let_anon(:type)
+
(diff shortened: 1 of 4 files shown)


mbj:
@dkubb Do you might giving this a 5% review.

I'd like to have feedback on the idea of adding a helper that allows to define a let binding and instance double the same moment. This would reduce a bit of boilerplate.

spec/unit/mutant/actor/binding_spec.rb
```diff
@@ -1,12 +1,13 @@
 RSpec.describe Mutant::Actor::Binding do
-  let(:actor_a)    { double('Actor-A', sender: sender_a, receiver: receiver_a) }
-  let(:sender_a)   { double('Sender-A')                                        }
-  let(:sender_b)   { double('Sender-B')                                        }
-  let(:receiver_a) { double('Receiver-A')                                      }
-  let(:payload)    { double('Payload')                                         }
-  let(:type)       { double('Type')                                            }
+  let_instance(:sender_a,   Mutant::Actor::Sender                                           )
+  let_instance(:sender_b,   Mutant::Actor::Sender                                           )
+  let_instance(:mailbox,    Mutant::Actor::Mailbox, receiver: :receiver_a, sender: :sender_a)
```
dkubb: I wonder if a better syntax would be:

```ruby
let_instance(:mailbox, Mutant::Actor::Mailbox) { { receiver: receiver_a, sender: sender_a } }
```

It's a tiny bit more curly braces, but it gives you the ability to do whatever you need within the block.

mbj: This syntax is also supported. The mapped ones get merged by the block ones.

I discovered that 99% of my let bindings can be handled without the explicit block at the call side. And its slightly less syntax in this form.

mbj: @dkubb do you agree to my argument above, so I can continue under the assumption the final review would not flag it?

dkubb: @mbj if you can do 99% of what you need with the syntax you came up with I guess that's good.

My main reason for flagging it is that you have a specific context where you could simply refer to things like `receiver_a` without having to do any special meta-programming. It was a red flag to me that extra logic is needed to work-around it.

mbj:
Do you mind rebasing again? See: #423 (comment)

mbj:
I'll close that one for now. Lets use traditional instance doubles initially and than in a later pass do something more sophisticated.

Issue #374, opened 2015-07-07
Question: RSpec selection overlap


tjchambers:
Just when I thought I was getting to understand this a new issue popped up:

I am issuing the following:

```
RAILS_ENV=test bundle exec mutant -r ./config/environment.rb -j2 --use rspec -I app/models/perform/activity.rb Activity > ./log/mutant/Activity.log
```
The header from mutant looks like:

```
Matcher:         #<Mutant::Matcher::Config match_expressions=[<Mutant::Expression: Activity>] subject_ignores=[] subject_selects=[]>
Integration:     rspec
Expect Coverage: 100.00%
Jobs:            2
Includes:        ["app/models/perform"]
Requires:        ["./config/environment.rb", "activity.rb"]
```
I *thought* my selection of examples would come from activity_rspec.rb, but I see the following in my logs:

```
- rspec:2:./spec/features/align/activity_log_spec.rb:9/Activity Log Admin views Assessment Activity log
- rspec:167:./spec/features/perform/activity_log_spec.rb:9/Activity Log Admin views Scorecard Activity log
- rspec:168:./spec/features/perform/activity_log_spec.rb:15/Activity Log Admin views RoadMap Activity log
- rspec:169:./spec/features/perform/activity_management_spec.rb:9/Activity Management User views an Activity
- rspec:170:./spec/features/perform/activity_management_spec.rb:15/Activity Management User views all Activities for a scorecards
- rspec:171:./spec/features/perform/activity_management_spec.rb:21/Activity Management User edits an Activity
- rspec:172:./spec/features/perform/activity_management_spec.rb:28/Activity Management User removes an Activity
- rspec:173:./spec/features/perform/activity_management_spec.rb:35/Activity Management User adds an Activity
- rspec:174:./spec/features/perform/activity_statuses_spec.rb:9/Activity statuses Admin lists Activity Status
- rspec:175:./spec/features/perform/activity_statuses_spec.rb:15/Activity statuses Admin creates new Activity Status
- rspec:211:./spec/features/solve/activity_log_spec.rb:9/Activity Log Admin views Activity log
- rspec:2631:./spec/models/perform/activity_spec.rb:7/Activity inactivates the swimlane
- rspec:2632:./spec/models/perform/activity_spec.rb:17/Activity inactivates the phase
- rspec:2633:./spec/models/perform/activity_spec.rb:27/Activity automatically assigns the default activity owner
[...]

PR #375, opened 2015-07-07
Add incremental working set selection

### LINKED ISSUE #262: Add ability to have stricter requirements for code written after a specific time
The idea with this is that you could configure mutant to be stricter with newer code than older code. You might want to say "all code written after today must be 100% covered", while still allowing the older code to be covered by the default thresholds.

I would use this on our $work project and force all new code to be 100% covered, while allowing us to be a bit more lenient with legacy code. I could see others using it this way too.

mbj:
[fix #262]

### DIFF
--- a/lib/mutant/cli.rb
+++ b/lib/mutant/cli.rb
@@ -155,6 +155,9 @@ module Mutant
       opts.on('--ignore-subject EXPRESSION', 'Ignore subjects that match EXPRESSION as prefix') do |pattern|
         add_matcher(:ignore_expressions, config.expression_parser.(pattern))
       end
+      opts.on('--since REVISION', 'Only select subjects touched since REVISION') do |revision|
+        add_matcher(:subject_filters, Repository::SubjectFilter.new(Repository::Diff.from_head(revision)))
+      end
     end
 
     # Add debug options
--- a/lib/mutant/matcher/config.rb
+++ b/lib/mutant/matcher/config.rb
@@ -4,7 +4,8 @@ module Mutant
     class Config
       include Adamantium, Anima::Update, Anima.new(
         :match_expressions,
-        :ignore_expressions
+        :ignore_expressions,
+        :subject_filters
       )
 
       INSPECT_FORMAT      = "#<#{self} %s>".freeze
@@ -12,6 +13,11 @@ module Mutant
       ATTRIBUTE_FORMAT    = '%s: [%s]'.freeze
       ENUM_DELIMITER      = ','.freeze
       EMPTY_ATTRIBUTES    = 'empty'.freeze
+      PRESENTATIONS       = IceNine.deep_freeze(
+        match_expressions:  :syntax,
+        ignore_expressions: :syntax,
+        subject_filters:    :inspect
+      )
       private_constant(*constants(false))
 
       DEFAULT = new(Hash[anima.attribute_names.map { |name| [name, []] }])
@@ -71,7 +77,9 @@ module Mutant
         ATTRIBUTE_FORMAT %
           [
             attribute_name,
-            public_send(attribute_name).map(&:syntax).join(ENUM_DELIMITER)
+            public_send(attribute_name)
+              .map(&PRESENTATIONS.fetch(attribute_name))
+              .join(ENUM_DELIMITER)
           ]
       end
 
--- a/lib/mutant/repository.rb
+++ b/lib/mutant/repository.rb
@@ -1,0 +1,66 @@
+module Mutant
+  module Repository
+    # Error raised on repository interaction problems
+    RepositoryError = Class.new(RuntimeError)
+
+    # Subject filter based on repository diff
+    class SubjectFilter
+      include Adamantium, Concord.new(:diff)
+
+      # Test if subject was touched in diff
+      #
+      # @param [Subject] subject
+      #
+      # @return [Boolean]
+      #
+      # @api private
+      def call(subject)
+        diff.touches?(subject.source_path, subject.source_lines)
+      end
+
+    end # SubjectFilter
+
+    # Diff between two objects in repository
+    class Diff
+      include Adamantium, Concord.new(:from, :to)
+
+      HEAD = 'HEAD'.freeze
+      private_constant(*constants(false))
+
+      # Create diff from head to revision
+      #
+      # @return [Diff]
+      #
+      # @api private
+      def self.from_head(to)
+        new(HEAD, to)
+      end
+
+      # Test if diff changes file at line range
+      #
+      # @param [Pathname] path
+      # @param [Range<Fixnum>] line_range
+      #
+      # @return [Boolean]
+      #
+      # @raise [RepositoryError]
+      #   when git command failed
+      #
+      # @api private
+      def touches?(path, line_range)
+        command = %W[
+          git log --pretty=oneline
+          #{from}...#{to}
+          -L #{line_range.begin},#{line_range.end}:#{path}
+        ]
+
+        stdout, status = Open3.capture2(*command, binmode: true)
+
+        fail RepositoryError, "Command #{command} failed!" unless status.success?
+
+        !stdout.empty?
+      end
+
+    end # Diff
+  end # Repository
+end # Mutant
--- a/spec/unit/mutant/matcher/compiler_spec.rb
+++ b/spec/unit/mutant/matcher/compiler_spec.rb
@@ -14,7 +14,7 @@ RSpec.describe Mutant::Matcher::Compiler do
   end
 
   let(:expected_predicate) do
-    Morpher.compile(s(:true))
+    Morpher.compile(s(:negate, s(:or)))
   end
 
   describe '.call' do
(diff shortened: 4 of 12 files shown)


mbj:
@dkubb Assuming all green this is ready for review and merge.

Validation has been done with:

```
bundle exec mutant --zombie -I lib -r mutant --since master --use rspec -- 'Mutant*'
```

Note the `--since` constrains what is returned from matching `Mutant*`.

Also there is no "subject widening yet". In later iterations of this feature I want to incldue `Foo#baz` when `Foo#bar` changed. But that should be done in later iterations.

mbj:
@tjchambers I rember you and @kbrock where interested in such a feature. Feel free to test it out and report any problems.

@dkubb Its enough for me when you can confirm the new feature, do not wait on feedback from others. But if you see bug reports: Lets get them fixed here.

lib/mutant/matcher/config.rb
```diff
@@ -71,7 +77,9 @@ def format_attribute(attribute_name)
         ATTRIBUTE_FORMAT %
           [
             attribute_name,
-            public_send(attribute_name).map(&:syntax).join(ENUM_DELIMITER)
+            public_send(attribute_name)
+             .map(&PRESENTATIONS.fetch(attribute_name))
```
dkubb: The indentation is off a bit here.

lib/mutant/repository.rb
```diff
@@ -0,0 +1,66 @@
...
+        diff.touches?(subject.source_path, subject.source_lines)
+      end
+
+    end # SubjectFilter
+
+    # Diff between two objects in repository
+    class Diff
+      include Adamantium, Concord.new(:from, :to)
+
+      HEAD              = 'HEAD'.freeze
```
dkubb: I would probably remove the whitespace before the `=`.

spec/unit/mutant/matcher/compiler_spec.rb
```diff
@@ -31,32 +31,47 @@
...
+              Morpher::Evaluator::Predicate::Boolean::Or.new(ignore_expression_predicates)
+            ),
+            Morpher::Evaluator::Predicate::Boolean::And.new(subject_filter_predicates)
+          ]
+        )
+      end
+
+      let(:expected_positive_matcher)    { Mutant::Matcher::Chain.new([matcher_a]) }
+      let(:attributes)                   { { match_expressions: [expression_a] }   }
+      let(:ignore_expression_predicates) { [] }
```
dkubb: Alignment of the `}` is off (assuming you follow the same alignment rules for `let` as in other projects, not sure)

spec/unit/mutant/repository/subject_filter_spec.rb
```diff
@@ -0,0 +1,28 @@
+RSpec.describe Mutant::Repository::SubjectFilter do
+  context '#call' do
+    subject { object.call(mutant_subject) }
+
+    let(:object)       { described_class.new(diff) }
+    let(:diff)         { double('Diff')            }
+    let(:return_value) { double('Return Value') }
```
dkubb: Alignment of `}` is off.

lib/mutant/cli.rb
```diff
@@ -155,6 +155,9 @@ def add_filter_options(opts)
       opts.on('--ignore-subject EXPRESSION', 'Ignore subjects that match EXPRESSION as prefix') do |pattern|
         add_matcher(:ignore_expressions, config.expression_parser.(pattern))
       end
+      opts.on('--since REVISION', 'Only select subjects touched since REVISION') do |revision|
```
dkubb: Is revision the best name for this? I've seen you use the branch name, and I assume the SHA works. Does a tag work too? Would the term "reference" be the more generic term for all of these things?

dkubb:
@mbj I did a quick first-pass over this code.

lib/mutant/cli.rb
mbj: > Is revision the best name for this?

Yes.

The `git log` manpage makes it clear:

```
SYNOPSIS
       git log [<options>] [<revision range>] [[--] <path>...]
```

I think `revision` in the git domain can be sloppily defined as: 

Any string git can parse into an unique commit. Including tags.

This can be confirmed via the `git rev-parse` manpage:

```
SPECIFYING REVISIONS
       A revision parameter <rev> typically, but not necessarily, names a commit object. It uses what is called an extended SHA-1 syntax. Here are various ways to spell object names. The ones listed near the end of this
       list name trees and blobs contained in a commit.
```

>  I've seen you use the branch name, and I assume the SHA works. Does a tag work too?

Yes. Even relative expressions such as `HEAD^1` work or more obscure ones.

lib/mutant/repository.rb
mbj: Refactoring leftover, I had more constants in local working tree, good catch.

mbj:

lib/mutant/cli.rb
dkubb: I brought this up because https://git-scm.com/book/en/v2/Git-Internals-Git-References calls these things references.

mbj: Mhh, I think that git than is a bit inconsistent. Right now as all user fencing documentation calls them `revision`, I think thats the better name.

Actually I like `reference` more as this noun is more closer to the actual meaning. The next time I'll touch this code I'll consider to rename to `refernece`

kbrock:
Thanks. Still wrapping my mind around this.
Feels odd that it would tap into git revisions, but has a lot of potential.

mbj:
> Feels odd that it would tap into git revisions, but has a lot of potential.

@kbrock How this does feel odd? I'm curious, because using git revisions of various forms and their delta is our daily business. Mutant does not add a layer on top of git, it just uses the information of the diff to *any* namable revision to select subjects.

dkubb:
@kbrock it might help if I explain our working process to you.

When @mbj and I work on a project we have something we call "hygenic commits", for lack of a better term. The idea is that every single commit in a feature branch must not break the build. We break doen every commit into the smallest possible unit that still provides value. We split refactoring, whitespace fixes, docs changes, feature additions and bug fixes into separate commits. We then have CI run the tests and some metrics on every build and it has to be green for the commit to eventually make it's way into the master branch.

Here's a typical PR: https://github.com/MountainRoseHerbs/spree/pull/71

What this allows us to do is rearrange commits to tell a story better and run `git bisect` should a bug be found in the future. When we have really complex changes, such as maintaining a spree fork with lots of additional fixes we've found it invaluable.

The addition of the `--since` flag allows us to add mutation testing to every single commit. It's not really practical to run it on every commit due to the length of time it takes; we expect a fast feedback loop for day-to-day development. A full run is still superior, but this new feature will allow us to make sure that no changes are made to the codebase without the new code being fully mutation covered.

mbj:
@kbrock BTW and the `--since` feature is fully optional. You do not have to use it. I've to admit I felt triggered by you using the word "odd", and just want to understand your concerns fully.

@dkubb Thx for your explanation on our strategy.

kbrock:
+1 on bisect and clean commits.

I tend to think of version control and running tests as separate concepts. Neither knowing about each other. Guess that is where I was surprised.

You see, when running the mutant specs, it did something with git and checking out files. I worried that I would loose some of my changes to mutant if I ran tests and hadn't committed my local files. 

So when I saw this referencing a git commit, it triggered that initial concern.


But in thinking about it, this feature is all about comparing code to historical code. And the best way to compare 2 different sets of code is to use git (e.g.: `git log SHA --nameonly`)

Thanks for this feature and sharing some of your reasoning.

mbj:
@kbrock I can confirm mutant does NEVER write to your working tree and or the git database. Mutant only invokes `git log`, which is a readonly command to reduce the manual pain to only select the touched subjects `--since` another state of the working tree.

This allows a trade of, that on big projects significantly speeds up mutation coverage, for a small reduction in confidence. Use with care.

tjchambers:
@mbj Just wanted to weigh in. 

This feature is much appreciated. However as you know I have targeted specs for testing the methods and have constructed a rake job that not only takes into account the changed source (not using git like this does), but also changes to the associated specs for the source in question. So as I am going through my mutation process cycle, I rerun mutant for all source that has changed AND sources where the specs for the source have changed as well. 

My inference from looking at how mutant is architected, and also some of the example <-> source association issues I am having, that this would not be so easy, but perhaps doable - to work from a spec change to what to "re-mutate".

kbrock:
I love that this is a major improvement for integration tests.

I think both @tjchambers and I setup a rule that associates a spec to a source file. Personally, I used the modules and class name:

- `lib/#{modules.join("/")}/#{class_name}.rb`
- `spec/#{modules.join("/")}/#{class_name}_spec.rb`
- run mutant on #{modules.join("::")}::#{class_name}

if either of the file or spec change, run the mutant again.

tjchambers:
@kbrock while there is in my case a *logical* relationship by convention in associating a spec file to a source file, for good reason (and AFAICT also a performance issue yet not understood by me) mutant still evaluates ALL of the spec files when determining which examples to run with rspec when a method is being mutated. So while I (and sounds like you) expect spec file === source file, mutant does not limit itself to that. 

@mbj and I have been discoursing over that, and I expect to have some more time to research later this week. I have issue #374 where I am recording this. For my size project this has ramifications. It is irrespective and unrelated to this major enhancement.

mbj:
@tjchambers Triggering subjects also from test changes sounds reasonable. I created an issue for this in https://github.com/mbj/mutant/issues/387.

PR #376, opened 2015-07-07
Refactor matcher config

mbj:
Extraction from #375.

### DIFF
--- a/lib/mutant/matcher/compiler.rb
+++ b/lib/mutant/matcher/compiler.rb
@@ -13,7 +13,7 @@ module Mutant
       def result
         Filter.new(
           Chain.build(config.match_expressions.map(&method(:matcher))),
-          predicate
+          ignored_subjects
         )
       end
 
@@ -34,32 +34,21 @@ module Mutant
 
     private
 
-      # Predicate constraining matches
+      # Predicate returning false on ignored subject
       #
       # @return [#call]
       #
       # @api private
-      def predicate
-        if subject_rejector
-          Morpher::Evaluator::Predicate::Negation.new(subject_rejector)
+      def ignored_subjects
+        rejectors = config.ignore_expressions.map(&SubjectPrefix.method(:new))
+
+        if rejectors.any?
+          Morpher::Evaluator::Predicate::Boolean::Negation.new(
+            Morpher::Evaluator::Predicate::Boolean::Or.new(rejectors)
+          )
         else
           Morpher::Evaluator::Predicate::Tautology.new
         end
-      end
-
-      # Subject rejector predicate
-      #
-      # @return [#call]
-      #   if there is a subject rejector
-      #
-      # @return [nil]
-      #   otherwise
-      #
-      # @api private
-      def subject_rejector
-        rejectors = config.subject_ignores.map(&SubjectPrefix.method(:new))
-
-        Morpher::Evaluator::Predicate::Boolean::Or.new(rejectors) if rejectors.any?
       end
 
       # Matcher for expression on env
--- a/lib/mutant/matcher/compiler.rb
+++ b/lib/mutant/matcher/compiler.rb
@@ -38,39 +38,13 @@ module Mutant
       #
       # @return [#call]
       #
-      # rubocop:disable MethodLength
-      #
       # @api private
       def predicate
-        if subject_selector && subject_rejector
-          Morpher::Evaluator::Predicate::Boolean::And.new([
-            subject_selector,
-            Morpher::Evaluator::Predicate::Negation.new(subject_rejector)
-          ])
-        elsif subject_selector
-          subject_selector
-        elsif subject_rejector
+        if subject_rejector
           Morpher::Evaluator::Predicate::Negation.new(subject_rejector)
         else
           Morpher::Evaluator::Predicate::Tautology.new
         end
-      end
-
-      # the subject selector predicate
-      #
-      # @return [#call]
-      #   if selector is present
-      #
-      # @return [nil]
-      #   otherwise
-      #
-      # @api private
-      def subject_selector
-        selectors = config.subject_selects.map do |attribute, value|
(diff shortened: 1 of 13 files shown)


mbj:
This was verified via:

```
bundle exec mutant -I lib -r mutant --zombie --use rspec 'Mutant::Reporter::CLI*' 'Mutant::Matcher::Compiler*' 'Mutant::Matcher::Config*'
```

@dkubb Ready for review, assuming all green.

lib/mutant/matcher/compiler.rb
```diff
@@ -34,60 +34,23 @@ def call(subject)
...
-          subject_selector
-        elsif subject_rejector
-          Morpher::Evaluator::Predicate::Negation.new(subject_rejector)
+      def ignored_subjects
+        rejectors = config.ignore_expressions.map(&SubjectPrefix.method(:new))
+
+        if rejectors.any?
+          Morpher::Evaluator::Predicate::Boolean::Negation.new(
+            Morpher::Evaluator::Predicate::Boolean::Or.new(rejectors)
+          )
```
dkubb: I'm not sure how your predicate objects work, but one general pattern that works here without a conditional would be:

```ruby
predicate = rejectors.inject(Contradiction.new) do |predicate, rejector|
  Or.new(predicate, rejector)
end
Negation.new(predicate)
```

Obviously the resulting object graph is a bit larger, but it removes the need for conditional logic.

mbj: @dkubb I can be used that way with a slightly different loop body. Great idea.

I want to do a cycle on morpher, to bring it more in-line with the API you use in the loop body.

lib/mutant/matcher/config.rb
```diff
@@ -22,6 +38,45 @@ def add(attribute, value)
...
 
+    private
+
+      # Present attributes
+      #
+      # @return [Array<Symbol>]
+      #
+      # @api private
+      def present_attributes
+        to_h.reject { |_key, value| value.empty? }.map(&:first)
```
dkubb: What about using `#keys` instead of `.map(&:first)`?

mbj: `Hash#reject` returns an `array` instance.

lib/mutant/matcher/config.rb
```diff
@@ -22,6 +38,45 @@ def add(attribute, value)
...
+      #
+      # @return [String]
+      #
+      # @api private
+      def inspect_attributes
+        attributes = present_attributes
+          .map(&method(:format_attribute))
+          .join(ATTRIBUTE_DELIM)
+
+        return EMPTY_ATTRIBUTES if attributes.empty?
```
dkubb: I think this guard clause doesn't really protect very complex code, I would probably write this as:

```ruby
attributes.empty? ? EMPTY_ATTRIBUTES : attributes
```

mbj: true. I take this suggestion.

lib/mutant/reporter/cli/printer/config.rb
```diff
@@ -11,13 +11,11 @@ class Config < self
...
           #
-          # rubocop:disable AbcSize
-          #
           # @api private
           def run
             info 'Mutant configuration:'
             info 'Matcher:         %s',      object.matcher.inspect
-            info 'Integration:     %s',      object.integration.name
+            info 'Integration:     %s',      object.integration
             info 'Expect Coverage: %0.2f%%', (object.expected_coverage * 100)
```
dkubb: Do you know about the `%g` flag? It will strip trailing zeros from the decimal part.

spec/unit/mutant/matcher/config_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+          .add(:match_expressions, parse_expression('Bar'))
+      end
+
+      it { should eql('#<Mutant::Matcher::Config match_expressions: [Foo,Bar]>') }
+    end
+
+    context 'with match and ignore expression' do
+      let(:object) do
+        described_class::DEFAULT
+          .add(:match_expressions, parse_expression('Foo'))
```
dkubb: Alignment of `parse_expression` statements (?)

lib/mutant/reporter/cli/printer/config.rb
dkubb: Are the parenthesis necessary here?

lib/mutant/matcher/config.rb
dkubb: Are you sure? http://ruby-doc.org/core-2.2.0/Hash.html#method-i-reject

Even in 1.9.3 it returns a Hash: http://ruby-doc.org/core-1.9.3/Hash.html#method-i-reject

dkubb:
@mbj I finished a pass over this code.

lib/mutant/matcher/config.rb
mbj: My 1.8 education strikes again. Good catch! Thx.

lib/mutant/matcher/compiler.rb
mbj: @dkubb I think till I did my cycle on morpher, I'll not touch this. I'll bascially "push this flag down" to the morpher lib.

lib/mutant/reporter/cli/printer/config.rb
mbj: No the parents are not needed. I just find it more clean. This is DSLish code so I outmitted the main parens for the `#info` call. But for the args I wanted to have them present.

Actually I think want to do explicit parens for everywhere. Will do that in a followup pass.

mbj: > Do you know about the %g flag? It will strip trailing zeros from the decimal part.

I'll address that on the followup pass on parens.

mbj:
@dkubb Ready for next pass.

PR #383, opened 2015-07-21
Change coverage on zero results to be 100%

### LINKED ISSUE #381: Error when --since does not select any subjects
When using `--since` with something like `HEAD~1` no subjects are selected and an error results. I think this shouldn't result in an error since you don't always know what the diff is going to contain, and it may occasionally be empty which shouldn't cause the build to break.

mbj:
[fix #381]

### DIFF
--- a/lib/mutant/result.rb
+++ b/lib/mutant/result.rb
@@ -4,6 +4,8 @@ module Mutant
 
     # Coverage mixin
     module Coverage
+      FULL_COVERAGE = Rational(1).freeze
+      private_constant(*constants(false))
 
       # Observed coverage
       #
@@ -11,24 +13,12 @@ module Mutant
       #
       # @api private
       def coverage
-        return Rational(0) if amount_mutation_results.zero?
-
-        Rational(amount_mutations_killed, amount_mutation_results)
+        if amount_mutation_results.zero?
+          FULL_COVERAGE
+        else
+          Rational(amount_mutations_killed, amount_mutation_results)
+        end
       end
-
-      # Hook called when module gets included
-      #
-      # @param [Class, Module] host
-      #
-      # @return [undefined]
-      #
-      # @api private
-      def self.included(host)
-        super
-
-        host.memoize(:coverage)
-      end
-
     end # Coverage
 
     # Class level mixin
@@ -52,6 +42,8 @@ module Mutant
         memoize(name)
       end
     end # ClassMethods
+
+    private_constant(*constants(false))
 
     # Mutant overhead running mutatet tests
     #
--- a/config/flay.yml
+++ b/config/flay.yml
@@ -1,3 +1,3 @@
 ---
 threshold: 18
-total_score: 1196
+total_score: 1204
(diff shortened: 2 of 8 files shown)


mbj:
@dkubb This was validated via:

```
bundle exec mutant --zombie -I lib -r mutant --use rspec --since 'HEAD^1' -- 'Mutant*'
```

And is ready for review.

lib/mutant/result.rb
```diff
@@ -11,24 +11,11 @@ module Coverage
       #
       # @api private
       def coverage
-        return Rational(0) if amount_mutation_results.zero?
+        return Rational(1) if amount_mutation_results.zero?
```
dkubb: Should `Rational(1)` be a constant?

mbj: True. Will change that.

dkubb: Given that there is very little code in this method I wonder if it would be cleaner to write it without a guard clause?

mbj: k

mbj:
@dkubb Under the assumption you finished your last pass, its ready for the next pass, all flags addressed.

dkubb:
@mbj hmm, for some reason this isn't working for me in our work project.

```
RAILS_ENV=test bundle exec mutant -r./config/environment --use rspec --since HEAD~1 -- ApplicationController

...

Mutant configuration:
Matcher:         #<Mutant::Matcher::Config match_expressions: [ApplicationController] subject_filters: [#<Mutant::Repository::SubjectFilter diff=#<Mutant::Repository::Diff from="HEAD" to="HEAD~1">>]>
Integration:     Mutant::Integration::Rspec
Expect Coverage: 100.00%
Jobs:            8
Includes:        []
Requires:        ["./config/environment"]
Subjects:        0
Mutations:       0
Kills:           0
Alive:           0
Runtime:         0.06s
Killtime:        0.00s
Overhead:        Inf%
Coverage:        0.00%
Expected:        100.00%
```

The status code was 0. And just to confirm I was using this branch:

```
$ bundle show mutant
/Users/dkubb/.gem/ruby/2.1.6/bundler/gems/mutant-9088a4b778e2
```

mbj:
@dkubb I checked our work project, and found *ugly* bundler behavior:

Our Gemfile in the branch includes:

```
gem 'mutant',       '~> 0.8.0', git: 'https://github.com/mbj/mutant.git'
gem 'mutant-rspec', '~> 0.8.0', git: 'https://github.com/mbj/mutant.git', branch: 'fix/exit-on-zero-subjects'
```

Some requires will happen against `master`, and some against the branch: When *both* gems reference the `fix/exit-on-zero-subjects` the behavior is absent.

tjchambers:
@mbj As an aside I have been a bit confused why mutant-rspec and mutant are in the same github project. It seems with the intro of minitest support they should be separated, but I am a noob at this and don't know what the ramifications of that are. I am sure that there is a good reason (you folks always have GREAT reasons), but thought I would comment.

mbj:
> @mbj As an aside I have been a bit confused why mutant-rspec and mutant are in the same github project.

Because each repo is a maintenance overhead. And the implementation of `mutant-rspec` is to small to mandate the overhead of another repository.

tjchambers:
I expected a good reason and got one. Thanks.

mbj:
@dkubb this is ready for the next pass.

PR #384, opened 2015-07-21
Fix incremental working set on out of tree subjects

### LINKED ISSUE #382: Out of repository subjects under incremental mutation run
When running with the `--since` flag subjects outside the repository are being flagged as changed.

For example, when running against spree, the `/home/ubuntu/spree/vendor/bundle/ruby/2.1.0/gems/paranoia-2.1.3/lib/paranoia.rb` file is being identified as changed, which is causing an error since it's outside the git repo.

mbj:
[fix #382]

### DIFF
--- a/lib/mutant/repository.rb
+++ b/lib/mutant/repository.rb
@@ -48,6 +48,8 @@ module Mutant
       #
       # @api private
       def touches?(path, line_range)
+        return false unless tracks?(path)
+
         command = %W[
           git log
           #{from}...#{to}
@@ -59,6 +61,26 @@ module Mutant
         fail RepositoryError, "Command #{command} failed!" unless status.success?
 
         !stdout.empty?
+      end
+
+    private
+
+      # Test if path is tracked in repository
+      #
+      # FIXME: Cache results, to avoid spending time on producing redundant results.
+      #
+      # @param [Pathname] path
+      #
+      # @return [Boolean]
+      #
+      # @api private
+      def tracks?(path)
+        command = %W[git ls-files --error-unmatch -- #{path}]
+        Kernel.system(
+          *command,
+          out: File::NULL,
+          err: File::NULL
+        )
       end
 
     end # Diff
--- a/lib/mutant/repository.rb
+++ b/lib/mutant/repository.rb
@@ -49,7 +49,7 @@ module Mutant
       # @api private
       def touches?(path, line_range)
         command = %W[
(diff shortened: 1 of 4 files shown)


mbj:
This was validated via:

```
bundle exec mutant --zombie -I lib -r mutant --use rspec --since master -- 'Mutant*'
```

mbj:

After merging, #383 this branch will be in conflict with the expected flay score, it would be great if you could fix this in case this is the only merge blocker.

mbj:
@dkubb Assuming all green this is also ready for the next pass.

lib/mutant/repository.rb
```diff
@@ -61,6 +63,21 @@ def touches?(path, line_range)
...
+      #
+      # @param [Pathname] path
+      #
+      # @return [Boolean]
+      #
+      # @api private
+      def tracks?(path)
+        command = %W[git ls-files --error-unmatch -- #{path}]
+        _stdout, _stderr, status = Open3.capture3(*command, binmode: true)
+        status.success?
```
dkubb: For consistency, I wonder if the other method in this class should omit blank lines between similar statements as this method?

mbj: The other method has 2 branches. This one not.

I added the blank lines in the other method to make sure the branches are not "buried" in the code.

mbj: I actually think that we should use `Process.spawn` here and not "bend" the `open3` into what we need, via ignoring stdout, and stderr. I'll fix that.

dkubb: Why not just `Kernel.system` since all you care about is the return value?

mbj:
@dkubb This is ready for the next pass.

lib/mutant/repository.rb
```diff
@@ -61,6 +63,26 @@ def touches?(path, line_range)
...
+      # @param [Pathname] path
+      #
+      # @return [Boolean]
+      #
+      # @api private
+      def tracks?(path)
+        command = %W[git ls-files --error-unmatch -- #{path}]
+        Kernel.system(
+          *command,
+          out: File::NULL,
```
dkubb: Oh, this is really nice, I didn't know these options existed.

mbj: It reduces the need to shell out, and build shell commands. Thats great as no escaping is required. (Still beware for option injection).

PR #385, opened 2015-07-23
Change Mutant::Isolation.call to use block form of IO.pipe

dkubb:
This branch changes `Mutant::Isolation.call` to use block form of `IO.pipe` to ensure open file handles are closed when no longer used.

### DIFF
--- a/lib/mutant/isolation.rb
+++ b/lib/mutant/isolation.rb
@@ -38,23 +38,25 @@ module Mutant
       #
       # @api private
       def self.call(&block)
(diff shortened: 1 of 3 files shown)


dkubb:
@mbj assuming all green, this is ready for review.

mbj:
I validated this succesfully via:

```
bundle exec mutant --zombie -I lib -r mutant --since master --use rspec -- 'Mutant*'
```

Need to make it part of the CI.

mbj:
@dkubb thx.

PR #388, opened 2015-07-28
Change --since to only target code under the cwd

dkubb:
This change allows for `--since` to be used in repositories with multiple different projects and only target the code specific to a subdirectory.


dkubb:

PR #389, opened 2015-07-29
Set writer pipe to binary mode

dkubb:


dkubb:
@mbj I've tested this against the code in spree that triggered the original bug and it fixed the issue.

dkubb:

PR #397, opened 2015-08-08
Remove invalild mutation on negating booleans

mbj:
* Fix 287 (read the ticket if you want to complain)


mbj:
@dkubb Ready for review.

PR #399, opened 2015-08-10
Add mutations from `defined?(foo)` to `true`, `false`

backus:
closes #334

PR #401, opened 2015-08-10
Cleanup defined mutations

mbj:
As a followup to #399, fixing issues I did not spotted on review.


mbj:
//cc @backus I discovered this after merging #399, and will merge it after CI passed. Just to let you know.

backus:
Thanks.  Why don't we emit `N_FALSE`?

mbj:
> Thanks. Why don't we emit N_FALSE?

Part of the `emit_singletons` implementation.

backus:
Ah alright makes sense

PR #402, opened 2015-08-10
Add mutation `super` -> `super()`

### LINKED ISSUE #309: Add mutation from `super` to `super()`
To quote @dkubb:

> At first I was going to say no, that super is more primitive than super() but I'm not so sure now.

> I mean, super basically allows you to delegate to the caller what arguments to pass up to the parent method. When you use super() you allow the parent method's own defaults to kick in. If you have no specs that assert different behaviour from the defaults is expected, then I think super() is a perfectly valid mutation -- if you want to use super then you should have specs to assert that giving up control to the caller is expected behaviour.

backus:
closes #309

### DIFF
--- a/lib/mutant/mutator/node/zsuper.rb
+++ b/lib/mutant/mutator/node/zsuper.rb
@@ -16,6 +16,7 @@ module Mutant
         # @api private
         def dispatch
           emit_singletons
+          emit(N_EMPTY_SUPER)
         end
 
       end # ZSuper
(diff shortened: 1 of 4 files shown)


backus:
I wasn't sure if there should be more to this change or not.  I'm still just beginning to familiarize myself with the codebase (evidenced by #401) so LMK if I'm missing anything.

mbj:
@backus Looks fine as is. Merging.

mbj:
@backus & Thx. Keep them coming!

lib/mutant/mutator/node/zsuper.rb
```diff
@@ -16,6 +16,7 @@ class ZSuper < self
         # @api private
         def dispatch
           emit_singletons
+          emit(N_EMPTY_SUPER)
```
mbj: I wounder if we should rename this to `N_ZSUPER` some day to match the AST types better.

PR #405, opened 2015-08-10
Add mutation `A = true` -> `remove_const :A`

### LINKED ISSUE #109: Add Mutation: A = true -> remove_const :A
Like #108 this will expose cases where the same constant is declared with the same value in a parent class.

backus:
closes #109

### DIFF
--- a/lib/mutant/mutator/node/named_value/constant_assignment.rb
+++ b/lib/mutant/mutator/node/named_value/constant_assignment.rb
@@ -19,7 +19,17 @@ module Mutant
           # @api private
           def dispatch
             mutate_name
-            emit_value_mutations if value
+            emit_value_mutations
+            emit_remove_const
+          end
+
+          # Emit remove_const
+          #
+          # @return [undefined]
+          #
+          # @api private
+          def emit_remove_const
+            emit(s(:send, cbase, :remove_const, s(:sym, name)))
           end
 
           # Emit name mutations
--- a/meta/casgn.rb
+++ b/meta/casgn.rb
@@ -4,4 +4,14 @@ Mutant::Meta::Example.add do
   mutation 'A__MUTANT__ = true'
   mutation 'A = false'
   mutation 'A = nil'
+  mutation 'remove_const :A'
+end
+
+Mutant::Meta::Example.add do
+  source 'self::A = true'
+
+  mutation 'self::A__MUTANT__ = true'
+  mutation 'self::A = false'
+  mutation 'self::A = nil'
+  mutation 'self.remove_const :A'
 end


backus:
I don't think this is up to par to be merged right now, but after trying to find similar mutations I could base this off of I wasn't sure how to better approach this one.  Opening this PR so @mbj can see this before I have to go offline for the rest of the day

mbj:
@backus You touched a subject that did not have full mutation coverage. Would be cool if you could backfill that one (or kill the exposed code).

Mutant recently got stable on self hosting, so this kinds of holes where not be detectable early.

mbj:
@backus BTW For consistency lets prefix future feature branch names with `feature/`? (No need to change this on that PR, just mentioning).

backus:
Sure thing.  Don't use the prefixing myself so I wasn't sure what qualified as a feature 



—

mbj:
> Sure thing.  Don't use the prefixing myself so I wasn't sure what qualified as a feature

Its tagged `enchancement` so its a feature. I'd actually love to change the prefix to mirror this. //cc @dkubb Your opinion on this?

dkubb:
@mbj I am ok with using `enhancement` as a prefix instead of `feature`. I do think we should add a `fix` tag though, since I think that's better than tagging a pull request as a `bug`.

backus:
Looking at the conditional 

```ruby
emit_value_mutations if value
```

I'm not sure when value would ever be falsey

backus:
@dkubb would you be able to comment? Looks like you added that line in 1e14b7f7

mbj:
> I'm not sure when value would ever be falsey

It will not, you can safely killl the conditional. An attribute assignment will *always* have a value. The moment hat code was written and reviewed mutant was not targetable at itself `--zombe` flag did not exist. This side effect is a typical find of mutation testing.

dkubb:
> @dkubb would you be able to comment?

Sorry, I have no idea what I was thinking there. It could be a left-over artifact. If I didn't write a spec for it, then I would suggest just removing the conditional. We should discover pretty quickly during usage if the code has an actual purpose.

backus:
CI green @mbj if you want to take another look

backus:
I made this change

```diff
@@ -6,3 +6,12 @@
   mutation 'A = nil'
   mutation 'remove_const :A'
 end
+
+Mutant::Meta::Example.add do
+  source 'self::A = true'
+
+  mutation 'self::A__MUTANT__ = true'
+  mutation 'self::A = false'
+  mutation 'self::A = nil'
+  mutation 'self.remove_const :A'
+end
```

in order to test that `cbase` was being preserved. Previously I had

```ruby
# Emit remove_const
#
# @return [undefined]
#
# @api private
def emit_remove_const
  emit(s(:send, nil, :remove_const, s(:sym, name)))
end
```

which would have broken something like `self::A = true` I believe.

mbj:
@backus Thx.

PR #416, opened 2015-08-14
Add mutation `foo.baz { bar }` -> `foo.bar`

### LINKED ISSUE #271: Add mutation from `receiver.something { single_method_call_with_implicit_self_receiver }` to `receiver.method_call`.
This will catch cases where a block does not add any semantics and should be avoided.

backus:
closes #271

### DIFF
--- a/lib/mutant/mutator/node/block.rb
+++ b/lib/mutant/mutator/node/block.rb
@@ -36,6 +36,22 @@ module Mutant
           return unless body
           emit(body)
           emit_body_mutations
+
+          mutate_body_receiver
+        end
+
+        # Mutate method send in body scope of `send`
+        #
+        # @return [undefined]
+        #
+        # @api private
+        def mutate_body_receiver
+          return unless n_send?(body)
+
+          body_meta = AST::Meta::Send.new(body)
+          send_meta = AST::Meta::Send.new(send)
+
+          emit(s(:send, send_meta.receiver, body_meta.selector, *body_meta.arguments))
         end
 
       end # Block
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,6 +3,7 @@
 * Add mutation from `a != b` to `!a.eql?(b)` and `!a.equal?(b)` #417
 * Add mutation `A.const_get(:B)` -> `A::B` #426
 * Add mutation `def foo(*args); end` into `def foo(*args); args = []; end` #423
+* Add mutation from `foo.baz { bar }` to `foo.bar` #416
 
 # v0.8.3 2015-09-01
 
--- a/config/flay.yml
+++ b/config/flay.yml
@@ -1,3 +1,3 @@
 ---
 threshold: 18
-total_score: 1219
+total_score: 1223
--- a/meta/block.rb
+++ b/meta/block.rb
@@ -69,3 +69,18 @@ Mutant::Meta::Example.add do
   mutation 'foo { |(_a)| }'
   mutation 'foo'
 end
+
+Mutant::Meta::Example.add do
+  source 'self.foo { bar(nil) }'
+
+  singleton_mutations
+  mutation 'self.foo'
+  mutation 'foo { bar(nil) }'
+  mutation 'self.foo { bar }'
+  mutation 'self.foo { nil }'
+  mutation 'self.foo {}'
+  mutation 'self.foo { self }'
+  mutation 'self.foo { raise }'
+  mutation 'bar(nil)'
+  mutation 'self.bar(nil)'
+end


dkubb:
@backus I would suggest rebasing any in-progress branches on top of master when it changes. This will allow us to make sure the in-progress code continues to pass with the latest mutant changes.

backus:
Should be rebased now

backus:
Added changelog entry

mbj:
Thx, I'll look into that one later (much later, will also be busy during the weekend). *think* there are AST constructs that might break the implemented mutation operator. We'll see.

lib/mutant/mutator/node/block.rb
```diff
@@ -36,6 +36,38 @@ def mutate_body
...
+        end
+
+        # Receiver of `send`
+        #
+        # @return [Parser::AST::Node]
+        #   receiver for `send`
+        #
+        # @api private
+        def send_receiver
+          send.children.first
```
mbj: IMO this would better be handled via a `AST::Meta::Send.new(send).receiver`.

lib/mutant/mutator/node/block.rb
```diff
@@ -36,6 +36,38 @@ def mutate_body
...
+
+        # Mutate method send in body scope of `send`
+        #
+        # @return [undefined]
+        #
+        # @api private
+        def mutate_body_receiver
+          return unless body.type.equal?(:send)
+
+          emit(s(:send, send_receiver, body_selector))
```
mbj: This does not forward the arguments. Would turn:

```
foo { bar(baz) }
```

Into:

```
bar
```

Where I think we want:

```
bar(baz)
```

mbj:
@backus Also please rebase ;)

backus:
Rebased and updated to handle for arguments @mbj if you want to give it another pass

lib/mutant/mutator/node/block.rb
```diff
@@ -36,7 +36,35 @@ def mutate_body
...
+          AST::Meta::Send.new(send)
+        end
+        memoize :send_meta
+
+        # Method send from body
+        #
+        # @return [Symbol]
+        #
+        # @api private
+        def body_meta
```
mbj: I see a conceptual problem with exposing this method to the instance namespace: What if someone calls this method without a prior check if body is of type `:send`? 

It would initialize an `AST::Meta::Send` argument with a node of incorrect type. So I think we should keep this as an lvar inside `mutate_body_receiver`, where we also have the guard.

mbj:
@backus I finished another pass.

backus:
Made the relevant changes, rebased, and added in documentation that was missing for `Node::Block#send_meta`.  Ready for another pass @mbj

backus:
Rebased this PR too @mbj

backus:
Rebased this PR too @mbj

mbj:
Do you mind rebasing again? See: https://github.com/mbj/mutant/pull/423#issuecomment-138591285

backus:
Rebased

lib/mutant/mutator/node/block.rb
```diff
@@ -36,7 +36,32 @@ def mutate_body
...
+
+        # AST meta information for `send`
+        #
+        # @return [undefined]
+        #
+        # @api private
+        def send_meta
+          AST::Meta::Send.new(send)
         end
+        memoize :send_meta
```
mbj: This method only has one call side. Memoizing it is overkill.

mbj: I'd argue to in-line the `send_meta` as an lvar to `mutate_body_receiver`.

lib/mutant/mutator/node/block.rb
```diff
@@ -36,7 +36,32 @@ def mutate_body
...
+
+        # Mutate method send in body scope of `send`
+        #
+        # @return [undefined]
+        #
+        # @api private
+        def mutate_body_receiver
+          return unless n_send?(body)
+
+          body_meta = AST::Meta::Send.new(body)
```
dkubb: I find it interesting that you extracted `send_meta` into a separate method, yet `body_meta` is inline within this method, yet it has more call sides.

For consistency I would probably extract `body_meta`, and I'd imagine you'd probably want to memoize it given it's used in more than one place.

mbj: If you follow this one, ignore my comments on `#send_meta`. and leave it as is for consistency.

backus: I believe I had that before: https://github.com/mbj/mutant/pull/416#discussion_r37455898

mbj: OT: I wanted to automate the process of "getting the meta of a given AST node" in a way we:

* Do not duplicate all the `children` DSL calls per node (right now we do in unparser, mutant, mutan-ast-meta)
* Ideally only create ONE meta per node

This is an ongoing refactoring idea, for now lets backfill meta "instances" till I've time to PR it.

dkubb:
> I believe I had that before

@backus ahh ok I don't want to give you contradictory advice to @mbj. I would just say that consistency is probably more important that finding the perfect organization for code. At least if it's consistent when you do discover new abstractions or new ways of organizing code then it's easier to bring all the code up to the new standard.

mbj:
> I believe I had that before
> @backus ahh ok I don't want to give you contradictory advice to @mbj. I would just say that consistency is probably more important that finding the perfect organization for code. At least if it's consistent when you do discover new abstractions or new ways of organizing code then it's easier to bring all the code up to the new standard.

I'm fine you follow either mine or @dkubb style. Just be consistent.

backus:
Rebased and updated. Inlined both `Meta::Send`s

Issue #270, opened 2014-11-11
Add a mutation from `a != b` to `!a.eql?(b)` and `!a.equal?(b)`.


mbj:
This will fix cases where permissive `Object#==` semantics get used unintentionally.

kbrock:
@mbj would you would prefer `eql?` and `equal?` over the simpler `==`?

(sorry if I asked this before - not sure if I submitted it the last time I typed this question)

mbj:
@kbrock `eql?`, `equal?` and `==` are not equal in its semantics, `==` is the operator with most semantics, `eql?` has less and `equal?` the least. Sometimes these are just aliases of each other.

As the fundamental 'Mutation operator design axiom' mutant tries to force the use of the 'Least powerful primitive'. Both for simple expressions (turning an `==` into an `eql?` etc) and for composed expressions (like a sequence `a; b` being turned into `b` etc).

Ruby does not have a build in representation for negations on `eql?` and `equal?`. We tend to use `!=` here. But this is equivalent to `!(a == b)` which uses the "weak" semantics of `==` internally. Hence this ticket exists.

Does all this answer your question?

PR #417, opened 2015-08-14
Add a mutation from `a != b` to `!a.eql?(b)` and `!a.equal?(b)`.

backus:
Opening a WIP PR for this since I'm not sure how to go about implementing #270 at the moment.  The `mutator/node/send.rb` class seems like an inappropriate place to implement this so I wanted to get some input before I invested time on another implementation

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,6 +3,7 @@
 * Remove invalid mutation `super(...)` to `super`
 * Add mutation from `def foo(a = true); end` to `def foo(a = true); a = true; end` #419
 * Add mutation from `def foo; end` to `remove_method :foo` #413
+* Add mutation from `a != b` to `!a.eql?(b)` and `!a.equal?(b)` #417
 
 # v0.8.2 2015-08-11
 
--- a/lib/mutant/mutator/node/send/binary.rb
+++ b/lib/mutant/mutator/node/send/binary.rb
@@ -19,7 +19,7 @@ module Mutant
             emit(left)
             emit_left_mutations
             emit_selector_replacement
-            emit(right) unless n_splat?(right)
+            emit(right)
             emit_right_mutations
           end
 
--- a/lib/mutant/mutator/node/send/binary.rb
+++ b/lib/mutant/mutator/node/send/binary.rb
@@ -21,6 +21,30 @@ module Mutant
             emit_selector_replacement
             emit(right)
             emit_right_mutations
+            emit_not_equality_mutations
+          end
+
+          # Emit mutations for `!=`
+          #
+          # @return [undefined]
+          #
+          # @api private
+          def emit_not_equality_mutations
+            return unless operator.equal?(:'!=')
+
+            emit_not_equality_mutation(:eql?)
+            emit_not_equality_mutation(:equal?)
+          end
+
+          # Emit negated method sends with specified operator
+          #
+          # @param new_operator [Symbol] selector to be negated
+          #
+          # @return [undefined]
+          #
+          # @api private
+          def emit_not_equality_mutation(new_operator)
+            emit(n_not(s(:send, left, new_operator, right)))
           end
 
         end # Binary
--- a/meta/send.rb
+++ b/meta/send.rb
@@ -425,7 +425,7 @@ Mutant::Meta::Example.add do
   mutation 'self[*bar]'
 end
 
-(Mutant::AST::Types::BINARY_METHOD_OPERATORS - %i[<= >= < > == eql?]).each do |operator|
+(Mutant::AST::Types::BINARY_METHOD_OPERATORS - %i[<= >= < > == != eql?]).each do |operator|
   Mutant::Meta::Example.add do
     source "true #{operator} false"
 
@@ -437,4 +437,18 @@ end
     mutation "true  #{operator} true"
     mutation "true  #{operator} nil"
   end
+end
+
+Mutant::Meta::Example.add do
+  source 'a != b'
+
+  singleton_mutations
+  mutation 'nil != b'
+  mutation 'self != b'
+  mutation 'a'
+  mutation 'b'
+  mutation 'a != nil'
+  mutation 'a != self'
+  mutation '!a.eql?(b)'
+  mutation '!a.equal?(b)'
 end


backus:
Gave this mutation a stab @mbj and @dkubb if you would like to take a look.  

I'd be happy to kill the flagged mutations [here](https://github.com/mbj/mutant/blob/master/lib/mutant/mutator/node/send/binary.rb#L22), but I'm not sure I can think of a case where the right hand would be a splat

backus:
I think the splat conditional was introduced because `BINARY_METHOD_OPERATORS` included `:[]` and `:[]=` at the time it was committed: [constants.rb#L12-L13](https://github.com/mbj/mutant/blob/8b7be8a565c0aa221e3a008e64282950c2aa1327/lib/mutant/constants.rb#L12-L13)

dkubb:
> I'd be happy to kill the flagged mutations here, but I'm not sure I can think of a case where the right hand would be a splat

I think @mbj will need to comment on what cases where the right hand would be a splat, but I had one comment about killing the specific mutation. I'd recommend killing it in it's own commit *before* this one, and then add your change in a separate commit; keeping the feature addition separate from "fixing" the mutation coverage.

There's a 50/50 chance you were going to do this anyway, but I figured I'd mention it in case you hadn't thought about it -- I know it's common for some people to commit code once they've got something working. What @mbj and I have been doing for a while is committing the smallest unit of change we possible can that provides an incremental improvement in the project quality, and fits a specific "theme". That is, we try to separate commits that refactor code, from commits that add functionality, and commits that fix a bug, or remove functionality, and so on.

It's often that when adding new functionality you end up refactoring code to make the addition easier. We would tend to split those into (at least) two commits; one to refactor the code and one to add it. In this specific case we have a situation where specs need to be added to fix a lack of mutation coverage on a method, and there's a feature addition; again we'd split them up.

One side effect of this approach is that we find it much easier to review each other's work. We can walk through the commits and kind of understand each individual transformation in isolation. There's nothing worse than looking at a huge diff, and trying to sort out whitespace fixes, from code removal, refactorings and feature additions. When it's all mixed together like that, you can't really make sense of what the author was trying to do (not saying that what you're doing here, your commits are small and awesome, just giving a more common example I think everyone's seen).

backus:
I agree with all of that @dkubb.

> I'd recommend killing it in it's own commit before this one, and then add your change in a separate commit

I committed the removed conditional separately.  I can reorder the commits now

dkubb:
Just to add a bit more to my last comment.. one thing we also do that helps keep things on target with our commits is that we follow the advice in http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html on commit message structure. In particular writing the summary in the recommended imperative style has naturally nudged us in the direction of starting the summary using a small set of keywords that match the theme of the commit and help us keep it single purposed.

One project @mbj and I collaborate on at work on has an interesting set of keywords on the git commit message summaries. Here's the keywords for the last 1000 commits in that project:

```shell
$ git log --pretty=oneline | head -1000 | cut -d' ' -f2 | sort | uniq -c | sort -nr
    298 Merge
    219 Add
    111 Fix
     97 Remove
     54 Change
     48 Upgrade
     37 Refactor
     21 Update
     19 Kill
     16 Revert
     15 Move
     10 Use
      9 Replace
      9 Rename
      7 Improve
      5 Reduce
      3 Make
      2 Prevent
      2 Disable
      2 Allow
      1 Track
      1 Slightly
      1 Setup
      1 Set
      1 Reorganise
      1 Perform
      1 Lift
      1 Introduce
      1 Handle
      1 Enable
      1 Downgrade
      1 Deduplicate
      1 Create
      1 Commenting
      1 Amend
      1 Align
```

I think some of the ones at the last half of the list are questionable, and probably should've been caught during code review, you can see how we structure things more closely.

I don't know if our approaches are better or worse than the norm, but they work for us and I've seen improvements in our code review process and the general quality of our pull requests, so I figured I'd pass them along.

I really need to get busy blogging this stuff ;)

backus:
I'll rewrite the commits!

dkubb:
> I'll rewrite the commits!

Haha, I wasn't saying that, just specifying how we do things. It's more important to keep the commits separated by the kind of transformation they are ;)

backus:
Only takes me a minute though and I respect the conventions.  Markus is definitely asleep so the build time isn't a big deal either :smile:

mbj:
> I think @mbj will need to comment on what cases where the right hand would be a splat, but I had one comment about killing the specific mutation. I'd recommend killing it in it's own commit before this one, and then add your change in a separate commit; keeping the feature addition separate from "fixing" the mutation coverage.

Removing the guard is correct. AFAIK there is no way the `Mutator::Node::Send::Binary` can be hit with an RHS splat as we have a dedicated dispatch for the various "send forms" ruby can have before the mutator gets invoked.

Ack on moving that to a separate commit. But we should make sure we build it also on CI. At the time of posting on the last commit pushed was build.

@backus You might want to look into the utility we wrote `git-push-each` that pushes each commit separately, making sure each one gets a build and is green. Find that utility here: https://github.com/dkubb/git-tools

dkubb:
The need to push each commit individually in order to build them is mostly a side effect of github and CircleCI's integration. I wish it were possible to tell one or both services "build all the commits, not just the last one". The tool I wrote works-around the limitation by pushing each commit individually one at a time, causing a CircleCI build to be triggered for each commit.

mbj:
@dkubb Much better description than mine. Thx for that.

backus:
I like the tools.  Not sure how to run `git-push-each` now that the commits are already up

backus:
Ah I had to rebase anyways.

dkubb:
@backus I plan to add a `--force` option when I figure out a nice clean way to add it, but for now you could temporarily change line 13 to the following:

```bash
commits=$(git rev-list HEAD "^$remote/$parent" --reverse)
```

You can then run it as `git push-each` or `git push-each $branch` if the parent branch is something other than master.

dkubb:
> Ah I had to rebase anyways.

Ahh ok, even better. Once you rebase on top of master `git push-each` should just work as you'd expect.

backus:
Rebased and ready for another pass @mbj

Issue #310, opened 2015-03-24
Add mutation to replace `def foo(arg = default)` with `def foo; arg = default`.


mbj:
This will ensure there is actually a call side that uses the available arity.

PR #419, opened 2015-08-16
Add mutation from `def foo(a = true); end` to `def foo(a = true); a = true; end`

backus:
WIP for #310.  I know this is ugly now. I wasn't sure though if there is a generic way that similar mutations handle this, so I figured I would ask before I invested time into a specific solution.  I'll add a few comments for things that seem strange.

### DIFF
--- a/lib/mutant/mutator/node/define.rb
+++ b/lib/mutant/mutator/node/define.rb
@@ -13,9 +13,38 @@ module Mutant
         # @api private
         def dispatch
           emit_arguments_mutations
+          emit_optarg_body_assignments
           emit_body(N_RAISE)
           emit_body(nil)
           emit_body_mutations if body
+        end
+
+        # Emit mutations with optional arguments as assignments in method
+        #
+        # @return [undefined]
+        #
+        # @api private
+        def emit_optarg_body_assignments
+          arguments.children.each do |argument|
+            next unless n_optarg?(argument) && AST::Meta::Optarg.new(argument).used?
+
+            emit_body_prepend(s(:lvasgn, *argument))
+          end
+        end
+
+        # Emit valid body ASTs depending on instance body
+        #
+        # @param node [Parser::AST::Node]
+        #
+        # @return [undefined]
+        #
+        # @api private
+        def emit_body_prepend(node)
+          if body
+            emit_body(s(:begin, node, body))
+          else
+            emit_body(node)
+          end
         end
 
         # Mutator for instance method defines
--- a/meta/def.rb
+++ b/meta/def.rb
@@ -81,6 +81,22 @@ Mutant::Meta::Example.add do
 end
 
 Mutant::Meta::Example.add do
+  source 'def foo(a, b = nil); true; end'
+
+  mutation 'def foo(_a, b = nil); true; end'
+  mutation 'def foo(a, b = nil); end'
+  mutation 'def foo; true; end'
+  mutation 'def foo(a, b = nil); raise; end'
+  mutation 'def foo(a, b = nil); nil; end'
+  mutation 'def foo(a, b = nil); false; end'
+  mutation 'def foo(a); true; end'
+  mutation 'def foo(a, b = nil); b = nil; true; end'
+  mutation 'def foo(b = nil); true; end'
+  mutation 'def foo(a, _b = nil); true; end'
+  mutation 'def foo(a, b); true; end'
+end
+
+Mutant::Meta::Example.add do
   source 'def foo(_unused); end'
 
   mutation 'def foo(_unused); raise; end'
@@ -113,6 +129,8 @@ Mutant::Meta::Example.add do
   mutation 'def foo(b = 0); end'
   mutation 'def foo(a, b = 0); end'
   mutation 'def foo; end'
+  mutation 'def foo(a = 0, b = 0); a = 0; end'
+  mutation 'def foo(a = 0, b = 0); b = 0; end'
   mutation 'def foo(a = 0, b = 0); raise; end'
 end
 
@@ -125,6 +143,7 @@ Mutant::Meta::Example.add do
   mutation 'def foo(a = nil); end'
   mutation 'def foo(_a = true); end'
   mutation 'def foo(a = true); raise; end'
+  mutation 'def foo(a = true); a = true; end'
 end
 
 Mutant::Meta::Example.add do
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 # v0.8.3 2015-08-xx
 
 * Remove invalid mutation `super(...)` to `super`
+* Add mutation from `def foo(a = true); end` to `def foo(a = true); a = true; end` #419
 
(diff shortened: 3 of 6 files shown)


lib/mutant/mutator/node/define.rb
```diff
@@ -13,11 +13,39 @@ class Define < self
...
+            child.type.equal?(:optarg)
+          end
+
+          return if optargs.empty?
+
+          lvasgns = optargs.map do |optarg|
+            s(:lvasgn, *optarg.children)
+          end
+
+          body_children = [*lvasgns, body].compact
```
backus: Here I'm handling the following two cases:

1. `def foo(a = true); end` meaning `body` is nil
2. `def foo(a = true); true; end` therefore `body` is a node

lib/mutant/mutator/node/define.rb
```diff
@@ -13,11 +13,39 @@ class Define < self
...
+
+          new_body = if body_children.one?
+            body_children.first
+          else
+            s(:begin, *body_children)
+          end
```
backus: I add this conditional to handle for if there is one or multiple nodes in the body.  If there is only one then we don't want to wrap it with `s(:begin, ...)` because it looks like unparser then finds the output to be different in the `Mutant::Meta::Example`.

lib/mutant/mutator/node/define.rb
dkubb: Can you write this as: `body_children = [*lvasgns, *body]`? If body is `nil`, then it's the same as if you had used `Array#compact`.

lib/mutant/mutator/node/define.rb
dkubb: I think that's a fine approach.

OT, but I wonder if there's room for a library where you could hand it a `Parser::AST::Node` and it would do some basic minimization and produce the simplest possible equivalent AST. I wouldn't necessarily want `unparser` to do that work, since it would be unexpected if it changed the structure.

If we had this, then we could build the AST using the most straight-forward way, and then minimize it before emitting it.

@mbj wdyt?

meta/def.rb
```diff
@@ -94,6 +110,7 @@
   mutation 'def foo(_unused = false); end'
   mutation 'def foo(_unused = true); raise; end'
   mutation 'def foo(_unused); end'
+  mutation 'def foo; _unused = true; end'
```
dkubb: Should the mutation just be `def foo; end`? What would be the point in emitting a variable prefixed with an underscore? 

BTW I'm not asking a rhetorical question, I don't know the answer, but I thought maybe it was worth discussing. There could be a good reason why we don't leave it out. It might be a situation where we all agree it should be left out, but we decide to do it later because it brings more value to commit this change now, and then work on optimizing the output in a future PR.

In general our approach is that we attempt to merge a PR if it brings an incremental improvement provided it meets code review and doesn't cause any metrics regressions. There are some changes that are blockers to getting PR merged, but if it's a case where adding it now vs. later makes no difference, then we'd usually defer it to get the improvement in now and then work on a follow-up PR making the additional change.

mbj: > What would be the point in emitting a variable prefixed with an underscore?

I think this is just a badly chosen example that overlaps with the underscore convention.

A mutation from

```
def foo(bar, baz = something)
  body
end
```

to

```
def foo(bar)
  baz = something
  body
end
```

Makes sense. This mutation does not change the semantics of the body, but the arity. It makes sure that there actually is a test call side that does pass in an argument that changes the default behavior.

mbj: Actually what would also make sense is an additional mutation to:

```
def foo(bar, baz = something)
  baz = something
  body
end
```

That would also ensure that the test actually specifies a different than default semantics via the default arguments.

backus: Yeah this line was just added because it is another optarg mutation. I'm not sure how different this mutation could be  since the point of this PR is to just move any optarg into the body.

lib/mutant/mutator/node/define.rb
backus: Unfortunately I think that wont work given the following example:

```ruby
require 'parser'
body = Parser::AST::Node.new(:true) # => (true)
[*body]                             # => []
```

meta/def.rb
backus:
```ruby
def foo(bar, baz = something)
  baz = something
  body
end
```

Shouldn't be too hard to add this mutation too. Should I include it in the PR?

mbj: @backus Seems very adjacent, and I actually think this one is what I initially wanted. :P Mostly because the other you added is redundant with that one.

lib/mutant/mutator/node/define.rb
dkubb: @backus I'm trying to understand what the purpose of the `#compact` is? I assumed that it was meant to filter out a case where the body is `nil`, because I don't see how `lvasgns` can contain any nil values.

backus: So this could alternatively be written as

```ruby
body_children = optargs.map do |optarg|
  s(:lvasgn, *optarg.children)
end

body_children << body if body
```

but I thought that didn't read as nice. Does that clear things up?

meta/def.rb
backus: Ah I'm confused then.  #310 certainly specifies `def foo(arg = default); end` → `def foo; arg = default`. Shall we just go with `def foo(arg = default); arg = default; end` then as the mutation?  that seems like it would be easier actually

backus:
ready for another pass @mbj

lib/mutant/mutator/node/define.rb
```diff
@@ -13,11 +13,40 @@ class Define < self
...
+        # @return [undefined]
+        #
+        # @api private
+        def emit_body_prepend(node)
+          if body
+            emit_body(s(:begin, node, body))
+          else
+            emit_body(node)
+          end
+        end
```
backus: Not sure what the best way to refactor this would be, but I just realized that this is structurally very similar to [`Node::Rescue#emit_concat`](https://github.com/mbj/mutant/blob/master/lib/mutant/mutator/node/rescue.rb#L50)

mbj: Lets push it down to private `Mutator::Node#emit_body_concat` ? And call it from here and the rescue mutator?

mbj: Mhh, does not make too much sense as we here want to prepend, in the rescue case concat. So lets keep it as is for now. I'll do another validation pass then merge. Code looks good to me.

lib/mutant/mutator/node/define.rb
```diff
@@ -13,11 +13,40 @@ class Define < self
         # @api private
         def dispatch
           emit_arguments_mutations
+          emit_body_assignment_mutations
```
mbj: I think this method is better named as `emit_optarg_body_assignments`.

mbj:
I did a minor rename and will merge this once green.

PR #427, opened 2015-08-20
Reorganize and DRY up AST::Meta

backus:
The `AST::Meta` scope has three classes living under it.  All included `include NamedChildren, Concord.new(:node)`.  

1. The first commit changed `Meta` from a module to a class and lifted up the mixins
2. The second commit moved the descendants to their own files in the directory `mutant/ast/meta`.  I figured that the `meta.rb` file only would accept another one or two classes until this reorganization would happen anyways.


mbj:
Nack. LSP violation. Each subclass has a different interface. I had this solution before :P and refactored it to what we have now. Each "meta class" should maybe just mixing a module to reduce duplication, but not be part of an inheritance tree.

There must be a better solution, ideally one where we can re-use the AST-Meta `children` declarations in the `Mutator::Node` ones. I've such a local branch and will push it after rebasing.

backus:
Ah fair enough then.  Should we salvage the second commit at least?  That just splits the classes into separate files.

mbj:
Also my solution will only instantiate ONE meta object per node instance. Reducing a bit of garbade. Not that this is the primary intention. Mutation generator performance was never an issue so far.

@backus Sorry for not telling you that plans, and the prior work I've done locally on our pairing.

mbj:
> Should we salvage the second commit at least? That just splits the classes into separate files.

Good idea. Lets keep that one. Please PR.

backus:
:thumbsup:

backus:
Can't respond right now on twitter DM btw.  On a plane an 2FA requires I receive a text message :/

PR #428, opened 2015-08-20
Move `AST::Meta` classes to their own files

backus:
The AST::Meta was a class or two away from exceeding the rubocop `ClassLength` limit so this seems like a necessary reorganization.


lib/mutant/ast/meta/resbody.rb
```diff
@@ -0,0 +1,15 @@
+module Mutant
+  module AST
+    # Node meta information mixin
+    module Meta
+
+      # Metadata for resbody nods
```
UrsMetz: I think "resbody nods" is a typo and should be "resbody nodes" (the former text is already present in the current HEAD)?

backus: I'll add a commit for the typo

PR #435, opened 2015-08-30
Enhance --fail-fast man documentation

tjchambers:
Provide a more explanatory definition of fail-fast option in mutant CLI

### DIFF
--- a/lib/mutant/cli.rb
+++ b/lib/mutant/cli.rb
@@ -168,7 +168,7 @@ module Mutant
     #
     # @api private
     def add_debug_options(opts)
-      opts.on('--fail-fast', 'Stops immediately on first alive mutation') do
+      opts.on('--fail-fast', 'Stops immediately upon first alive mutation') do
         update(fail_fast: true)
       end
       opts.on('--version', 'Print mutants version') do
--- a/lib/mutant/cli.rb
+++ b/lib/mutant/cli.rb
@@ -168,7 +168,7 @@ module Mutant
     #
     # @api private
     def add_debug_options(opts)
-      opts.on('--fail-fast', 'Fail fast') do
+      opts.on('--fail-fast', 'Stops immediately on first alive mutation') do
         update(fail_fast: true)
       end
       opts.on('--version', 'Print mutants version') do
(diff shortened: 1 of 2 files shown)


mbj:
@tjchambers Chances you can squash all your commits into one? I do not see a reason to have 3 commits for a 2 line change. Also one of your intermediary commits fails the build, which contradicts the "hygenic commit" strategy we follow.

tjchambers:
Ok - will squash (soon as I figure out how). Makes sense.

mbj:
@tjchambers 

```
git checkout $your_branch_name
git fetch origin
git rebase --interactive origin/master
```

Than follow the explanation.

mbj:
@tjchambers Did you meant to close this PR?

tjchambers:
Yes - I can't get it to squash like it should be so I was going to abandon it and start over. I have followed the instructions 6 times and I get merge conflicts and no appropriate single commit. It should be simpler than this. Not your fault. I just cannot get it to work. I don't want to waste your time.

mbj:
@tjchambers I'm happy to get you a pairing showing you on how to do such rebases / sqashing. De-Facto you are "wasting" my time with starting over also ;)

So to improve your effectiveness when contributing getting you know those rebasing / sqashing activities will make you a much better contributor - de-facto: Saving me time.

tjchambers:
@mbj to be honest - I do feel like I know what I am doing in this case. I tried it using the CLI and it makes perfect sense, until I cannot save the content in the editor! 

SO I used SourceTree which is what I use all the time and have done this before and it looks fine but generates merge conflicts because of my choice of wording changes which don't seem to be sequenced in the commits properly. No matter how I re-arranged the commits, my wording change is out of sequence and causes a merge conflict which when resolved creates another commit.  

IMHO when squashing 3 commits into one in the same original sequence should be dead simple. Too many *&^*&^ options. So I will do it right and learn on my own time. I should not have committed this in the first place without running the test suite. SO I made loads of mistakes and need to learn from them. 

If I sounds pissed it is because I am - at myself.

mbj:
@tjchambers The only way to fix this is to show you how its done. I offered you pairings multiple times. Just agree to one ;)

I'm also fine you start over. Just efficiency could be higher. Your help to the project is still net positive, so I cannot force you. And thx for the help!

mbj:
BTW I consider all graphical version managers very leaky abstractions, IMO they result in a solution I call "abcraption".

PR #437, opened 2015-09-01
Revert "Add mutations from `and`/`or` into `if`/`else`"

mbj:
Reverts mbj/mutant#430

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,7 +3,6 @@
 * Remove invalid mutation `super(...)` to `super`
 * Add mutation from `def foo(a = true); end` to `def foo(a = true); a = true; end` #419
 * Add mutation from `def foo; end` to `remove_method :foo` #413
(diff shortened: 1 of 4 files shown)


mbj:
@backus I revert this PR because it results in bad mutations.

It mutates to invalid AST / invalid syntax:

```
 def emit_const_get_mutation
-  unless (selector.equal?(:const_get) && n_sym?(arguments.first))
+  unless if selector.equal?(:const_get)
+    n_sym?(arguments.first)
+  end
     return
   end
   emit(s(:const, receiver, AST::Meta::Symbol.new(arguments.first).name))
 end
```

backus:
Think this is fixable or should we just not bother with this mutation?

mbj:
@backus Lets re-introduce it fixed. I just reverted it to get master clean / releasable. My fault for not spotting this earlier.

PR #439, opened 2015-09-04
Anima update

mbj:


mbj:
@dkubb This is ready for review.

Gemfile
```diff
@@ -1,7 +1,6 @@
-# encoding: utf-8
-
 source 'https://rubygems.org'
 
 gemspec name: 'mutant'
 
 gem 'devtools', git: 'https://github.com/rom-rb/devtools.git'
+eval_gemfile File.expand_path('../Gemfile.shared', __FILE__)
```
dkubb: The `Gemfile.shared` file is empty. What is the reason for evaling it? Why not just remove the file and this line.

lib/mutant/cli.rb
```diff
@@ -207,7 +207,7 @@ def update(attributes)
     #
     # @api private
     def add(attribute, value)
-      update(attribute => config.public_send(attribute).dup << value)
+      with(attribute => config.public_send(attribute).dup << value)
```
dkubb: I know in the past you've preferred to use different constructs for this, like:

```ruby
config.public_send(attribute) + [value]
```

lib/mutant/matcher/config.rb
```diff
@@ -41,7 +41,7 @@ def inspect
       #
       # @api private
       def add(attribute, value)
-        update(attribute => public_send(attribute).dup << value)
+        with(attribute => public_send(attribute).dup << value)
```
dkubb: I have the same comment for this as https://github.com/mbj/mutant/pull/439/files#r38951709

Gemfile
mbj: There should be a commend in that file explaining the reason why it exist. If its not there anymore I rebased it away by accident.

Always when I update mutants dependencies (especially morpher and anima) there is a state where dependencies need to be pulled from git sources.

These git sources need to be reflected in the Gemfiles for:

* The vendored test app, per rspec (and future minitest) integration
* The corpus tests Gemfiles

That would be ~5 Gemfiles (tendency to grow) I'd need to manually add (remove) the git source from/to.

Now instead: I can simply use this (now empty) file to ease upgrades of shared dependencies.

As I said: there was a note in the `Gemfile.shared` trying to explain its presence as (now empty) file.

lib/mutant/cli.rb
mbj: You are correct. This code predates my preference for binary, non destructive operations by default. Even when the frozen values do not leak the context.

Will fix that.

spec/spec_helper.rb
```diff
@@ -33,7 +33,7 @@
 require 'test_app'
 
 module Fixtures
-  TEST_CONFIG = Mutant::Config::DEFAULT.update(reporter: Mutant::Reporter::Trace.new)
+  TEST_CONFIG = Mutant::Config::DEFAULT.with(reporter: Mutant::Reporter::Trace.new)
```
dkubb: I know `#with` was added in Anima, but I wonder if maybe `#merge` would be a better alternative. It implies the object state is being merged with new state and that a new object is being returned. The `#with` method isn't commonly used for this purpose.

spec/support/corpus.rb
```diff
@@ -134,6 +134,7 @@ def install_mutant
         repo_path.join('Gemfile').open('a') do |file|
           file << "gem 'mutant', path: '#{relative}'\n"
           file << "gem 'mutant-rspec', path: '#{relative}'\n"
+          file << "eval_gemfile File.expand_path('#{relative.join('Gemfile.shared')}')\n"
```
dkubb: If you decide to remove `Gemfile.shared` then this can be removed too.

spec/spec_helper.rb
mbj: `#merge` implies a binary operation of two "equally shaped" objects. I especially thought about this one but than realized when we have an `anima` object on the left side, and a `Hash` instance on the right side: The operation is not commutative and should not be called `#merge`.

spec/support/corpus.rb
mbj: Please see my comment about the intentions of `Gemfile.shared` here: https://github.com/mbj/mutant/pull/439#discussion_r38951885

mbj:
@dkubb Ready for the next pass. All flags addressed (by code change or argument).

Gemfile
dkubb: Ahh ok. I read the comment in the file, but the exact reason wasn't clear to me.

dkubb:
@mbj I love that mutant self-mutates. By definition if a change doesn't result in the build breaking then it is ok to merge assuming the code passes review.

PR #442, opened 2015-09-11
Fix mutation to invalid ast on block glueing

mbj:
The mutation operator from #416  was misimplemented. It lifted the inner send to have the receiver of the outer send. Not the outer send AS receiver as intended.

As a byproduct of this fix a related crash to invalid AST was fixed. The exact crash case was NOT included as a spec, because this can only happen on the unwanted mutation.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -2,6 +2,9 @@
 machine:
   ruby:
     version: '2.2'
+dependencies:
+  pre:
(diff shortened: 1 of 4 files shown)


mbj:
@dkubb As this bug is very straightforward to understand / fix. Does not add new infrastructure etc. I'll merge without your review to get a release out.

PR #446, opened 2015-09-22
Fix config attribute primitivity

mbj:
Config attributes should only be primitives, or nested primitives.

Before this changeset they could reference objects that in longterm need to be configurable / parsable from a config file.

So this changeset aims to reduce the "non primitive infection" of the config object to a minimum.


mbj:
Will not be finished in this repo. Closing.

PR #448, opened 2015-10-09
Add mutation `Date.parse` → `Date.iso8601`, `Date.civil`, etc.

backus:


backus:
I've added `rubocop:disable ClassLength` to `Mutator::Node::Send` since it now slightly exceeds the maximum class length.  An easy solution to this I think would be to move send mutation definitions to a config file. Regardless of the refactor though, I would be happy to work on it but I would prefer to defer that to another PR so this doesn't get blocked for too long.

lib/mutant/ast/meta/send.rb
```diff
@@ -56,6 +56,19 @@ def binary_method_operator?
           Types::BINARY_METHOD_OPERATORS.include?(selector)
         end
 
+        # Check if receiver is a top level constant
+        #
+        # @return [Boolean]
+        #
+        # @api private
+        def receiver_top_level_const?
+          return false unless receiver && receiver.type.equal?(:const)
```
mbj: You can include `Mutant::AST::NodePredicates` here and replace `receiver.type.equal?` with `n_const?(receiver)`. This will make the code slightly cleaner, same for `const_parent.type.equal?(:cbase)`.

lib/mutant/ast/meta/send.rb
```diff
@@ -56,6 +56,19 @@ def binary_method_operator?
...
+
+          const_parent = receiver.children.fetch(0)
+
+          const_parent.nil? || const_parent.type.equal?(:cbase)
```
mbj: I wounder if the above two lines would better be factored out into another ast-meta? This way the `receiver.children.fetch(0)` would use the named children DSL, removing the magic constant `0`.

mbj:
@backus I finished a code only pass over this. With two minor flags.

backus:
@mbj should be ready for another pass

lib/mutant/ast/meta/const.rb
```diff
@@ -0,0 +1,24 @@
...
+    # Node meta information mixin
+    module Meta
+
+      # Metadata for const nodes
+      class Const
+        include NamedChildren, Concord.new(:node), NodePredicates
+
+        children :base, :name
+
+        # Test if AST node is a top level constant
```
mbj: The description is IMO wrong. Unless its a cbase you cannot statically know the node is referencing a toplevel constant.

The predicate IMO should be named `possible_toplevel?`. 

Counter example:

```
module Foo
  class String
  end

  String # const node that dynamically returns `Foo::String` but predicate returns `true`.
end

**Comment by mbj** [comment]
@backus I finished another pass.

**Comment by backus** [comment]
Fixed in fcfc35a and ready for another pass @mbj

File: lib/mutant/ast/meta/send.rb [hunk 4]
```diff
@@ -56,6 +56,17 @@ def binary_method_operator?
           Types::BINARY_METHOD_OPERATORS.include?(selector)
         end
 
+        # Check if receiver is a top level constant
```
mbj: Lets name this: `Check if the receiver is likely a top level constant` ;)

mbj: Or `possibly` for sync with implementation.

mbj:
@backus Another pass done, with a tiny (and hopefully in your eyes not nitpicky) flag.

Also: Could you add this to the Changelog?

lib/mutant/ast/meta/send.rb
backus: Woops yeah I forgot to change this one.  Sorry about that

backus:
- [x] fixed wording and related method name
- [x] added changelog entry

@mbj should be ready for another pass

mbj:
@backus Thx.

PR #454, opened 2015-10-19
Add mutation from `@foo` to `foo`

backus:
In the first commit I've reconfigured the AUOM integration to not expect full mutation coverage since this commit adds mutations which AUOM does not cover

### DIFF
--- a/lib/mutant/mutator/node/named_value/access.rb
+++ b/lib/mutant/mutator/node/named_value/access.rb
@@ -6,7 +6,7 @@ module Mutant
         # Mutation emitter to handle named value access nodes
         class Access < Node
 
-          handle(:gvar, :cvar, :ivar, :lvar, :self)
+          handle(:gvar, :cvar, :lvar, :self)
 
         private
 
@@ -17,6 +17,45 @@ module Mutant
           # @api private
           def dispatch
             emit_singletons
+          end
+
+          # Named value access emitter for instance variables
+          class Ivar < Access
+            NAME_RANGE = (1..-1).freeze
+
+            handle(:ivar)
+
+            children :name
+
+            # Emit mutations
+            #
+            # @return [undefined]
+            #
+            # @api private
+            def dispatch
+              emit_attribute_read
+              super()
+            end
+
+            private
+
+            # Emit instance variable as attribute send
+            #
+            # @return [undefined]
+            #
+            # @api private
+            def emit_attribute_read
+              emit(s(:send, nil, attribute_name))
+            end
+
+            # Variable name without leading '@'
+            #
+            # @return [Symbol]
+            #
+            # @api private
+            def attribute_name
+              name.slice(NAME_RANGE).to_sym
+            end
           end
 
         end # Access
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,7 @@
 
 * Add mutation from `Date.parse` to more strict parsing methods #448
 * Add mutation from `foo.to_i` to `Integer(foo)` #455
+* Add mutation from `@foo` to `foo` #454
 
 # v0.8.5 2015-09-11
 
--- a/config/flay.yml
+++ b/config/flay.yml
@@ -1,3 +1,3 @@
 ---
 threshold: 18
-total_score: 1244
+total_score: 1253
--- a/meta/ivar.rb
+++ b/meta/ivar.rb
@@ -1,0 +1,6 @@
+Mutant::Meta::Example.add do
+  source '@foo'
+
+  singleton_mutations
+  mutation 'foo'
+end
--- a/meta/op_assgn.rb
+++ b/meta/op_assgn.rb
@@ -2,6 +2,7 @@ Mutant::Meta::Example.add do
   source '@a.b += 1'
 
   singleton_mutations
+  mutation 'a.b += 1'
   mutation '@a.b += -1'
   mutation '@a.b += 2'
   mutation '@a.b += 0'


backus:
Here are the mutations in AUOM revealed by this PR: 

```diff
@@ -1,15 +1,15 @@
 def pretty_unit
   if unitless?
     return
   end
-  numerator = Inspection.prettify_unit_part(@numerators)
+  numerator = Inspection.prettify_unit_part(numerators)
   denominator = Inspection.prettify_unit_part(@denominators)
   if numerator.empty?
     numerator = "1"
   end
   if denominator.empty?
     return " #{numerator}"
   end
   format(UNIT_FORMAT, numerator, denominator)
 end
```

```diff
@@ -1,15 +1,15 @@
 def pretty_unit
   if unitless?
     return
   end
   numerator = Inspection.prettify_unit_part(@numerators)
-  denominator = Inspection.prettify_unit_part(@denominators)
+  denominator = Inspection.prettify_unit_part(denominators)
   if numerator.empty?
     numerator = "1"
   end
   if denominator.empty?
     return " #{numerator}"
   end
   format(UNIT_FORMAT, numerator, denominator)
 end
```

```diff
@@ -1,13 +1,13 @@
 def initialize(scalar, numerators, denominators)
   @scalar = scalar
   [numerators, denominators].permutation do |left, right|
     left.delete_if do |item|
       right.delete_at(right.index(item) || right.length)
     end
   end
   @numerators = numerators.freeze
   @denominators = denominators.freeze
-  @unit = [@numerators, @denominators].freeze
+  @unit = [numerators(), @denominators].freeze
   @scalar.freeze
 end
```

```diff
evil:AUOM::Unit#initialize:/Users/johnbackus/Dropbox/coding/auom/lib/auom/unit.rb:248:f1c53
@@ -1,13 +1,13 @@
 def initialize(scalar, numerators, denominators)
   @scalar = scalar
   [numerators, denominators].permutation do |left, right|
     left.delete_if do |item|
       right.delete_at(right.index(item) || right.length)
     end
   end
   @numerators = numerators.freeze
   @denominators = denominators.freeze
-  @unit = [@numerators, @denominators].freeze
+  @unit = [@numerators, denominators()].freeze
   @scalar.freeze
 end
```

```diff
evil:AUOM::Unit#initialize:/Users/johnbackus/Dropbox/coding/auom/lib/auom/unit.rb:248:e489b
@@ -1,13 +1,13 @@
 def initialize(scalar, numerators, denominators)
   @scalar = scalar
   [numerators, denominators].permutation do |left, right|
     left.delete_if do |item|
       right.delete_at(right.index(item) || right.length)
     end
   end
   @numerators = numerators.freeze
   @denominators = denominators.freeze
   @unit = [@numerators, @denominators].freeze
-  @scalar.freeze
+  scalar().freeze
 end
```

dkubb:
@mbj I'm curious if it would be possible to emit these methods without the empty parens?

backus:
@mbj this is ready for review

mbj:
> @mbj I'm curious if it would be possible to emit these methods without the empty parens?

No. Because unparser is lvar scope aware and needs to use the empty parents here to disambiguate the private getter with the lvar.

By default argument less getter calls are emitted without parens in unparser.

mbj:
> Here are the mutations in AUOM revealed by this PR:

@backus The build will not pass unless `auom` is fully mutation covered. We need to fix auom first. I'll do this before merging that one.

mbj:
Interesting it *still* passed, it should not.

backus:
> Interesting it still passed, it should not. 

Note https://github.com/backus/mutant/commit/05de96219351b892ac8501675a6996b69b3f8abd

mbj:
> Interesting it still passed, it should not.
> Note backus@05de962

Hah, was not looking at the diff yet. Did not expected you'd change this. Lets fix auom before introducing a "regression" to the corpus test.

backus:
> Lets fix auom before introducing a "regression" to the corpus test.

:thumbsup: I'll drop the commit in that case

backus:
Dropped the config change

lib/mutant/mutator/node/named_value/access.rb
```diff
@@ -19,6 +19,44 @@ def dispatch
...
+            end
+
+            private
+
+            # Emit instance variable as attribute send
+            #
+            # @return [undefined]
+            #
+            # @api private
+            def emit_attribute
```
mbj: The name of the method is very generic. Lets name it something like `emit_attribute_read`?

lib/mutant/mutator/node/named_value/access.rb
```diff
@@ -19,6 +19,44 @@ def dispatch
...
+              emit(s(:send, nil, attribute_name))
+            end
+
+            # Variable name without leading '@'
+            #
+            # @return [Symbol]
+            #
+            # @api private
+            def attribute_name
+              name.slice(1..-1).to_sym
```
mbj: I think the range literal should be moved to a constant.

We typically use constants for all mutable literals, unless the propose of the literal is to create a mutable value.

mbj: Ranges are monkeypatchabe, thus mutable :(

mbj:
@backus I did a small code only pass, fix for auom is in progress.

lib/mutant/mutator/node/named_value/access.rb
dkubb: Another nice side effect of moving literals to constants is you become more aware of their usage in code. Over-usage of string literals or other promitives in code is often a sign of other code smells being present, such as [Primitive Obsession] (http://c2.com/cgi/wiki?PrimitiveObsession) or [Stringly Typed](http://c2.com/cgi/wiki?StringlyTyped) coding.

Not saying that's the case here, just something to be aware of.

backus:
@mbj fixed flags.  Should be ready for another pass

mbj:
Closing in favor of #463 as I've direct control over this branch. @backus Your commit authorship is keept intact.

mbj:
@backus I just realize I could have pushed to your fork as you gave me access, sorry for the noise.

PR #455, opened 2015-10-19
Add mutation `foo.to_i` → `Integer(foo)`

backus:


backus:
@mbj this is ready for review

backus:
Note: I realize the send node mutator is getting bloated.  I would be open to extracting it into a few different classes in the future.  I tried to ignore this while adding this mutation since it would be out of scope for the PR

mbj:
> Note: I realize the send node mutator is getting bloated. I would be open to extracting it into a few different classes in the future. I tried to ignore this while adding this mutation since it would be out of scope for the PR

The send mutator was already broken up into several specific classes. See `Send::Arguments`, `Send::Binary`, `Send::Unary`. I think we can have subclasses for receiver mutations etc.

Right now lets merge this as is and do a dedicated pass for breaking it up.

mbj:
@backus Do you mind the next time you avoid non ASCII characters in commit messages? I spotted the `→` too late.

backus:
> @backus Do you mind the next time you avoid non ASCII characters in commit messages? I spotted the → too late.

Which commit? The `→` should only be in the issue title

mbj:
> Which commit? The `→` should only be in the issue title

True. I assumed commit == PR title. My fault.

PR #466, opened 2015-10-29
Improve README

backus:
There are a few things I think could be improved with the readme.  The lowest hanging fruit is the organization.  I think code examples and getting started information should come first in the readme.


README.md
```diff
@@ -100,6 +48,47 @@ mutant --include lib --require virtus --use rspec Virtus::Attribute.build
 mutant --include lib --require virtus --use rspec Virtus::Attribute#type
 ```
 
+Configuration
+-------------
+
+Occasionally mutant will produce a mutation with an infinite runtime.  When this happens
```
mbj: You are using two whitespaces here after `.` is this intentional?

dkubb: This is probably a hold-over habit from [typing on typewriters](http://www.quickanddirtytips.com/education/grammar/two-spaces-after-a-period). I catch myself doing the same thing sometimes even though I never really typed on a typewriter much as a kid.

README.md
```diff
@@ -100,6 +48,47 @@ mutant --include lib --require virtus --use rspec Virtus::Attribute.build
...
+```
+
+which will fail specs which run for longer than 5 seconds.
+
+Rails
+-------
+
+Assuming you are using rspec, you can mutation test Rails models by adding the following lines to your Gemfile:
+
+```ruby
```
mbj: Can we move this gemfile addition outside of the `Rails` context? Its not rails specific.

README.md
mbj: I'd prefer we go for one space after `.`. For the reasons:

* Its more usual in documentation I read these days
* I'd never remind myself to use two spaces for consistency.

mbj:
@backus There are 2 flags I'd like to see addressed before merge. The two spaces after fullstop one and the gemfile instructions that are not rails specific.

README.md
dkubb: > I'd prefer we go for one space after `.`.

I agree. The 2 space after period thing isn't really recommended anymore.

backus:
@mbj flags addressed.  Added a few additional fixes

mbj:
@backus thx.

PR #468, opened 2015-10-30
Reduce matcher interface

mbj:
* New primary interface #call makes specs and implementations much
  easier
* We had #each mostly for historical reasons that are not relevant
  anymore
* Kills mutations in `Mutant::Matcher*` namespace that had been present because it was written before `--zombie` was functional.

Extraction from #446


mbj:

Sorry this change is so big. I initially thought about constraining it into the `Matcher` namespace and the call side in `Env::Bootstrap` but I found a blatand LSP violation that created a much bigger backpressure to also touch `Env::Bootstrap`.

Lets say it this way: The size of this change is paying legacy debt.

mbj:
This can be validated for:

* Mutation coverage of the change
* Not breaking mutant

Via:

```
bundle exec mutant --zombie -I lib -r mutant --use rspec 'Mutant::Env::Bootstrap*' 'Mutant::Expression*' 'Mutant::Matcher*' 
```

This is more coarse grained than the check on CI that does via `--since` a more fine grained validation.

lib/mutant/env/bootstrap.rb
```diff
@@ -95,7 +95,7 @@ def scope_name(scope)
       # @api private
       def infect
         config.includes.each(&$LOAD_PATH.method(:<<))
-        config.requires.each(&method(:require))
+        config.requires.each(&Kernel.method(:require))
```
backus: Curious.  Why did you add `Kernel`?

lib/mutant/env/bootstrap.rb
```diff
@@ -137,7 +137,7 @@ def expression(scope)
         name = scope_name(scope) or return
 
         unless name.instance_of?(String)
-          warn("#{scope.class}#name from: #{scope.inspect} returned #{name.inspect}. #{SEMANTICS_MESSAGE}")
+          warn("#{scope.class}#name from: #{scope} returned #{name}. #{SEMANTICS_MESSAGE}")
```
backus: since this is duplicated

`warn("#{scope.class}#name from: #{scope} ` + ` #{SEMANTICS_MESSAGE}")`

it might read better if you had a `semantic_warning` method

mbj: Next iterations. I saw that but deliberately ignored the possible improvement just to focus on the PRs intentions. I touched this one only for the reason of mutation coverage.

lib/mutant/matcher/method.rb
```diff
@@ -1,133 +1,154 @@
...
+              Adamantium::Flat,
+              Concord::Public.new(:scope, :target_method, :evaluator)
 
       # Methods within rbx kernel directory are precompiled and their source
       # cannot be accessed via reading source location. Same for methods created by eval.
       BLACKLIST = %r{\A(kernel/|\(eval\)\z)}.freeze
 
-      # Enumerate matches
+      SOURCE_LOCATION_WARNING =
+        '%s does not have valid source location, unable to emit subject'.freeze
```
tjchambers: I think this should say 'does not have a valid source location'

lib/mutant/matcher/method.rb
```diff
@@ -1,133 +1,154 @@
...
-      def each
-        return to_enum unless block_given?
+      def call(env)
+        evaluator.call(scope, target_method, env)
+      end
+
+      # Abstract method match evaluator
+      #
+      # Present to avoid passing the env argument around in case the
+      # logic would be implemnented directly on the Matcher::Method
```
tjchambers: implemented is misspelled here

lib/mutant/matcher/method.rb
mbj: @tjchambers Good catch. My non native speaking mind needed multiple seconds to identify the difference between code and your suggestion.

Just to help me: Can you next time highlight the diff a bit? "does not have *a* valid source location" would make it more obvious.

lib/mutant/env/bootstrap.rb
mbj: Short: To be able to spec the global side effect this code should trigger.

Long:

When specifying requires, its IMO okay to rely on the fact `Kernel#require` will do the right thing when called. So its IMO fine to place message expectations to "specify" certain requires happen.

I could actually have setup a real file require it and observe a constant was added, but that would defeat the idea of a unit spec. A unit spec under execution should not mutate global state. (OT: Thats a nice way to try defining unit spec vs integration spec!)

Now when setting up message expectations (that are temporal monkey patches effectively) we need an object to target this monkey patches on.

2 thoughts arise from that:

* The primary API of this class is `Bootstrap.call` that returns `Env`. No instance of `Bootstrap` is visible to the API client. The whole intention is to cause side effects and produce a "stable" environment object. Tests should only use the API of the class and do not "fuck around" with their implementation details. As I do not have access on an instance of `Bootstrap` via the API I do not have an object to set the message expectation up against.

* If I'd even have an API where I could reach instances of `Bootstrap` setting up a message expectation on the object under test would be a very bad idea. Here I actually have to blame ruby because very important APIs with global side effects such as `Kernel#require` are available as private methods everywhere. I'd prefer it if `require 'foo'` would only work toplevel, or at class level, but not at instance level per default. This would reduce lots of confusion.

In summary: By changing the code to another construct, causing the same side effect. Its now more testable.

lib/mutant/matcher/method.rb
tjchambers: understood. thanks for the feedback.

lib/mutant/env/bootstrap.rb
mbj: Another maybe more useful instance of the same pattern is available here: https://github.com/mbj/mutant/blob/master/bin/mutant#L13-L29

This is the call side for mutants self-zombification. It explicitly passes kernel into the zombifier API, for testing the zombifier I could implement a "fake Kernel" (that also does some message expectations, but not via rspec) to specify the behavior of Zombification.

Here is the test support class: https://github.com/mbj/mutant/blob/master/spec/support/ruby_vm.rb

This example is an instance of the same idea: When multiple APIs that produce the same effect exist (`#require` vs `Kernel.require`) use the one that is easier to unit test.

lib/mutant/matcher/method.rb
mbj: @tjchambers And to make it explict: Big thx for that comments! Even I ping @dkubb explicitly, you are very much invited to add your flags to anything you see. Especially for questions.

lib/mutant/env/bootstrap.rb
backus: Very interesting!

mbj: Actually I think next refactorings will pass in Kernel, $stderr, $stdout and other global friends explicitly into the mutant namespace. I had great success with that on the `Mutant::Zombifier` sub-project. So it'll help everywhere.

Each time I have to place a message expectation via rspec on a global constant something in my heart dies.

lib/mutant/matcher/method.rb
```diff
@@ -1,133 +1,154 @@
...
       # Methods within rbx kernel directory are precompiled and their source
       # cannot be accessed via reading source location. Same for methods created by eval.
       BLACKLIST = %r{\A(kernel/|\(eval\)\z)}.freeze
 
-      # Enumerate matches
+      SOURCE_LOCATION_WARNING =
+        '%s does not have valid source location, unable to emit subject'.freeze
+
+      CLOSURE_WARNING =
+         '%s is dynamically defined in a closure, unable to emit subject'.freeze
```
backus: I'm not quite sure what this warning means which makes me think this error message should probably be a bit longer and explain more.

mbj: In summary it detected that case:

```
class Foo
end

klass = some_cool_dynamic_computation_resulting_in_foo
klass.class_eval do
   def bar
   end
end
```

If this `Foo#bar` candidate for subject emitting on the expression `Foo#bar` mutant cannot do this, as per current runtime mutation insertion: Mutant would not know how to reproduce the constant scope `Foo#bar` is defined under. There is no monkeypatch syntax mutant could generate that by 100% would reproduce the same constant scope.

lib/mutant/matcher/method.rb
```diff
@@ -1,133 +1,154 @@
...
         end
 
-        self
-      end
+      private
 
-    private
+        # Test if method should be skipped
+        #
+        # @return [Truthy]
```
backus: Would it make sense to change this to 

```ruby
# @return [Array] if valid source location
# @return [nil] otherwise
```

seems more accurate and, if you used something in the future which tried to use YARD to do type checks, this wouldn't introduce errors.

lib/mutant/matcher/method.rb
backus: Interesting.  I think you should add an example like the code you specified to the warning.  It will likely end up as a question on github issues otherwise.

lib/mutant/matcher/method.rb
mbj: I do not like that:

* Its not correct. Because it actually returns `Env` as `Env#warn` returns `Env`. If you found an implenentation of `Env#warn` that returns `Array` its a bug.
* There is no intention to return `Env` hence it should not be documented. Nobody should rely on that.

lib/mutant/matcher/method.rb
mbj: I think all warnings should be short and technically accurate. Single lines. BUT link to an exanation arthicle in the wiki / Readme section etc.

Mostly for the reason I do not like long warnings in the code, because there they are buried. The long explanation texts you where asking for are actually part of the missing documentation.

lib/mutant/matcher/method.rb
backus: Ah I didn't realize it returned that. Fair enough

mbj: BTW this is all from the fact that `nil` and `false` are indistinguishable under truthiness. So mutant kills explicit boolean returns because the call side is a conditional :(

mbj: Its actually a mirror of the conceptual weakness of ruby. One we have to accept.

lib/mutant/matcher/method.rb
backus: Well in this case you could then add a link to this thread in the comment.  Shortened url: http://git.io/vlCMF

mbj: Why not take the chance and lift this to a README.md section?

backus: Good point.  I could add that

lib/mutant/matcher/method/instance.rb
```diff
@@ -3,62 +3,68 @@ class Matcher
...
-          def source_location
-            scope.unmemoized_instance_method(method_name).source_location
+          def match?(node)
+            node.type.equal?(:def)                 &&
+            node.location.line.equal?(source_line) &&
+            node.children.fetch(NAME_INDEX).equal?(method_name)
           end
 
-        end # Memoized
+          # Evaluator specialized for memoized instance mthods
```
backus: Should be

> Evaluator specialized for memoized instance *methods*

lib/mutant/matcher/method.rb
backus: https://github.com/mbj/mutant/pull/473

lib/mutant/matcher/method.rb
```diff
@@ -1,133 +1,154 @@
...
+        #
+        # @return [Enumerable<Subject>]
+        #
+        # @api private
+        def call
+          return EMPTY_ARRAY if skip?
 
-        if !skip? && subject
-          yield subject
+          [subject].compact
```
backus: `skip?` checks `matched_node_path.any?(&method(:n_block?))` which should return false when `matched_node_path` is empty.  This is also the only case where `subject` should return `nil`.  Could this just be `[subject]` then?

lib/mutant/matcher/method/instance.rb
```diff
@@ -3,62 +3,68 @@ class Matcher
...
+          # @param [Parser::AST::Node] node
           #
-          # @return [Array{String,Fixnum}]
+          # @return [Boolean]
           #
           # @api private
-          def source_location
-            scope.unmemoized_instance_method(method_name).source_location
+          def match?(node)
+            node.type.equal?(:def)                 &&
```
backus: This could be `n_def?` right?

mbj: Yes. And should.

lib/mutant/matcher/method.rb
mbj: True. But as this code is self covered mutant would have taken out the `#compact`, and told me that case was redundant. As it does not there is likely a reason we missed. I'll check it out.

mbj: As expected, for the singleton case removing the compact results in:

```
  1) Mutant::Matcher::Method::Singleton#call on singleton methods when also defined on lvar warns about definition on non const/self
     Failure/Error: expect(subject).to eql([])
       
       expected: []
            got: [nil]
       
       (compared using eql?)
       
       Diff:
       @@ -1,2 +1,2 @@
       -[]
       +[nil]
```

Because in that case `skip?` can return `true` and the node path can still be empty.

lib/mutant/env/bootstrap.rb
mbj: @backus Because this class was touched so heavily your flag triggered me to go the extra mile and include a nicer way to print out formatted semantics warnings.

PR #469, opened 2015-10-30
Correctly exclude evaled code

mbj:
* The previous regexp was ment to only match `(eval)` literally.


backus:
:thumbsup: Fixed my use case

PR #470, opened 2015-10-31
Drop Subject#public?

mbj:
* A relict from the days where test selection depended on this
  information
* Kill mutations in Mutation#insert, also very old code.


mbj:
@dkubb Assuming all green this is also ready for review.

PR #471, opened 2015-10-31
Kill mutations in Zombifier and Reporter

mbj:


mbj:

lib/mutant/reporter/cli/format.rb
```diff
@@ -42,7 +42,7 @@ class Output
           #
           # @api private
           def tty?
-            @tty
+            tty
```
backus: Could this just be an `alias_method` instead?

mbj: It could. But I do not like it for its inheritance behavior. While I'm not using inheritance here I actually have no reason to avoid it ;)

mbj: Fixed. thx.

PR #474, opened 2015-11-01
Add more metrics to env progress reporter

mbj:


mbj:
@dkubb Assuming all green this is also ready for review.

PR #477, opened 2015-11-07
Remove & inline Mutant::Delegator with its only call side

mbj:


mbj:
Closing that one, will re-open it later.

Issue #370, opened 2015-07-06
Add machine readable report


mbj:
For various cases a machine readable output in a defined JSON (or YAML) format will be handy.

There already exist private non open-sourced implementations of these reports we can maybe opensource.

References #369.

kbrock:
SimpleCov has a standard output, would it be possible to generate that?

You could hookup to
https://codeclimate.com/repos/53e7c2c3e30ba01d44007b35/coverage_setup

Looks like [CodeClimate](http://blog.codeclimate.com/blog/2015/06/19/code-climate-platform/) may be their view of the future, and they probably a good standard to follow. Wonder how that changes their current status of code coverage.

Tools like [rubocop][https://github.com/bbatsov/rubocop#json-formatter] do have a formatter that looks usable. Not sure if the other tools in the [metric_fu](https://github.com/metricfu/metric_fu/) suite have similar formatters. I'm betting that code climate is probably getting these tools to produce a common output. Either that or they have given up on them.

mbj:
@kbrock The simplecov output is made for line based coverage tools, it does not feature the amount of granularity mutant supports. Both for actual coverage and metadata.

I see myself supporting that format as an aggregate, a lossfull conversation to plug into existing ecosystems and also code climate. But I consider this format a bad choice for mutants primary machine readable report format.

kbrock:
@mbj agreed - it is lossy. but it is a very common format for others. Would be great to get mutant code coverate into CodeClimate and/or travis.

I'll say this again, take a look at [CodeClimate]'s format. The [spec] describes column/offset and length. Their issue construct seems to fit mutant's use cases.

Specifically, I'd like the final output to be aggregated when multiple mutations point to the same problem. E.g. This code block could be self, nil, deleted, ... - which all seem to say, this code is useless or not tested.

[CodeClimate]: http://blog.codeclimate.com/blog/2015/06/19/code-climate-platform/
[spec]: https://github.com/codeclimate/spec/blob/master/SPEC.md

I keep throwing around the idea of using sqlite as the interim format. I guess that is kinda the same thing as using json. But I guess the concept of aggregating similar errors together is high on my mind. (including errors / code coverage across multiple reporting engines)

kbrock:
I keep throwing around the idea of using sqlite as the interim format. I guess that is kinda the same thing as using json. But I guess the concept of aggregating similar errors together is high on my mind. (including errors / code coverage across multiple reporting engines).

[...]

PR #479, opened 2015-11-11
Adds support for dumping JSON formatted summary of execution.

jheth:
We participated in Rails Rumble over the weekend and built a product based on your mutant gem.  Instead of parsing standard out I implemented a printer for collecting results into a hash and then dumping them to a file as JSON format.  Hopefully this is useful for you and others using the tool.  I'm happy to make any changes you suggest.  I did not write any tests for this, is that something you would like?

I also realize there's a bit of duplication between the CLI printer. If you have ways to consolidate that I'm open to it.

### DIFF
--- a/lib/mutant/reporter/hash/printer.rb
+++ b/lib/mutant/reporter/hash/printer.rb
@@ -1,0 +1,52 @@
+module Mutant
+  class Reporter
+    class Hash
+      # Hash runner status printer base class
+      class Printer
+        include AbstractType, Delegator, Adamantium::Flat, Concord.new(:object), Procto.call(:run)
+
+        private_class_method :new
+
+        delegate :success?
+
+        NL = "\n".freeze
+
+        # Run printer
+        #
+        # @return [self]
+        #
+        # @api private
+        abstract_method :run
+
+      private
+
+        # Visit a collection of objects
+        #
+        # @return [Class::Printer] printer
+        # @return [Enumerable<Object>] collection
+        #
+        # @return [undefined]
+        #
+        # @api private
+        def visit_collection(printer, collection)
+          collection.map do |object|
+            visit(printer, object)
+          end
+        end
+
+        # Visit object
+        #
+        # @param [Class::Printer] printer
+        # @param [Object] object
+        #
+        # @return [undefined]
+        #
+        # @api private
+        def visit(printer, object)
+          printer.call(object)
+        end
+
+      end # Printer
+    end # CLI
+  end # Reporter
+end # Mutant
--- a/lib/mutant/reporter/hash/printer/config.rb
+++ b/lib/mutant/reporter/hash/printer/config.rb
@@ -1,0 +1,36 @@
+module Mutant
+  class Reporter
+    class Hash
+      class Printer
+        # Printer for mutation config
+        class Config < self
+
+          # Report configuration
+          #
+          # @param [Mutant::Config] config
+          #
+          # @return [undefined]
+          #
+          # @api private
+          def run
+            matcher = object.matcher.inspect
+            # #<Mutant::Matcher::Config match_expressions: [Rumble::Loop]>
+            # Parse out expression [...]
+            expression = matcher.scan(/\[([\w:,\*]+)\]/).flatten.first
+            expression ||= matcher
+
+            {
+              matcher: expression,
+              integration: object.integration,
+              expect_coverage: (object.expected_coverage * 100).to_f,
+              jobs: object.jobs,
+              includes: object.includes,
+              requires: object.requires
+            }
+          end
+
+        end # Config
+      end # Printer
+    end # CLI
+  end # Reporter
+end # Mutant
--- a/lib/mutant/reporter/hash/printer/env_progress.rb
+++ b/lib/mutant/reporter/hash/printer/env_progress.rb
@@ -1,0 +1,64 @@
+module Mutant
+  class Reporter
+    class Hash
+      class Printer
+        # Env progress printer
+        class EnvProgress < self
+          delegate(
+            :coverage,
+            :amount_subjects,
+            :amount_mutations,
+            :amount_mutations_alive,
+            :amount_mutations_killed,
+            :runtime,
+            :killtime,
+            :overhead,
+            :env
+          )
+
+          # Run printer
+          #
+          # @return [undefined]
+          #
+          # rubocop:disable AbcSize
+          #
+          # @api private
+          def run
+            {
+              config: visit(Config, env.config),
+              subjects: amount_subjects,
+              mutations: amount_mutations,
+              kills: amount_mutations_killed,
+              alive: amount_mutations_alive,
+              runtime: runtime,
+              killtime: killtime,
+              overhead: overhead_percent.to_f,
+              coverage: coverage_percent.to_f,
+              expected: (env.config.expected_coverage * 100).to_f
+            }
+          end
+
+        private
+
+          # Coverage in percent
+          #
+          # @return [Float]
+          #
+          # @api private
+          def coverage_percent
+            coverage * 100
+          end
+
+          # Overhead in percent
+          #
+          # @return [Float]
+          #
+          # @api private
+          def overhead_percent
+            (overhead / killtime) * 100
+          end
+        end # EnvProgress
+      end # Printer
+    end # CLI
+  end # Reporter
+end # Mutant
--- a/lib/mutant/reporter/hash/printer/env_result.rb
+++ b/lib/mutant/reporter/hash/printer/env_result.rb
@@ -1,0 +1,26 @@
+module Mutant
+  class Reporter
+    class Hash
+      class Printer
+        # Full env result reporter
+        class EnvResult < self
+          delegate(:failed_subject_results)
+          delegate(:subject_results)
+
+          # Run printer
+          #
+          # @return [undefined]
+          #
+          # @api private
+          def run
+            {
+              failed_subject_results: visit_collection(SubjectResult, failed_subject_results),
+              success_subject_results: visit_collection(SubjectResult, subject_results.select(&:success?)),
+              env_progress: visit(EnvProgress, object)
+            }
+          end
+        end # EnvResult
+      end # Printer
+    end # Hash
+  end # Reporter
+end # Mutant
--- a/lib/mutant/reporter/hash/printer/mutation_result.rb
+++ b/lib/mutant/reporter/hash/printer/mutation_result.rb
@@ -1,0 +1,149 @@
+module Mutant
+  class Reporter
+    class Hash
+      class Printer
+        # Reporter for mutation results
+        class MutationResult < self
+
+          delegate :mutation, :test_result
+
+          DIFF_ERROR_MESSAGE =
+            'BUG: Mutation NOT resulted in exactly one diff hunk. Please report a reproduction!'.freeze
+
+          MAP = {
+            Mutant::Mutation::Evil    => :evil_details,
+            Mutant::Mutation::Neutral => :neutral_details,
+            Mutant::Mutation::Noop    => :noop_details
+          }.freeze
+
+          NEUTRAL_MESSAGE =
+            "--- Neutral failure ---\n" \
+            "Original code was inserted unmutated. And the test did NOT PASS.\n" \
+            "Your tests do not pass initially or you found a bug in mutant / unparser.\n" \
+            "Subject AST:\n" \
+            "%s\n" \
+            "Unparsed Source:\n" \
+            "%s\n" \
+            "Test Result:\n".freeze
+
+          NO_DIFF_MESSAGE =
+            "--- Internal failure ---\n" \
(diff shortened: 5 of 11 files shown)


lib/mutant/reporter/hash/printer.rb
```diff
@@ -0,0 +1,52 @@
+module Mutant
+  class Reporter
+    class Hash
+      # Hash runner status printer base class
+      class Printer
```
mbj: This base class duplicates a lot with `Mutant::Reporter::CLI::Printer`, I suggest deduplicating that into `Mutant::Reporter::Printer`.

lib/mutant/reporter/hash/printer/config.rb
```diff
@@ -0,0 +1,36 @@
...
+          #
+          # @return [undefined]
+          #
+          # @api private
+          def run
+            matcher = object.matcher.inspect
+            # #<Mutant::Matcher::Config match_expressions: [Rumble::Loop]>
+            # Parse out expression [...]
+            expression = matcher.scan(/\[([\w:,\*]+)\]/).flatten.first
+            expression ||= matcher
```
mbj: This will not work for complex matchers that are made off many expressions, as happens on projects without a namespace.

I suggest to serialize the full `Matcher::Config` object at the granularity it has instead of the first match expression.

lib/mutant/reporter/hash/printer/config.rb
```diff
@@ -0,0 +1,36 @@
...
+
+            {
+              matcher: expression,
```
mbj: I'd align the values in this hash.

lib/mutant/reporter/hash/printer/env_progress.rb
```diff
@@ -0,0 +1,64 @@
...
+          # Run printer
+          #
+          # @return [undefined]
+          #
+          # rubocop:disable AbcSize
+          #
+          # @api private
+          def run
+            {
+              config: visit(Config, env.config),
```
mbj: I'd align the values in this hash.

lib/mutant/reporter/hash/printer/mutation_result.rb
```diff
@@ -0,0 +1,149 @@
...
+            end
+          end
+
+          # Print no diff message
+          #
+          # @return [undefined]
+          #
+          # @api private
+          def print_no_diff_message
+            NO_DIFF_MESSAGE % [
```
jheth: Is there a way to make these methods protected.  This file is largely duplicated and I could extend and override if these weren't private.

lib/mutant/reporter/hash/printer/env_progress.rb
```diff
@@ -0,0 +1,64 @@
...
+          end
+
+        private
+
+          # Coverage in percent
+          #
+          # @return [Float]
+          #
+          # @api private
+          def coverage_percent
```
mbj: These are duplicated with the CLI reporter, lets push them down to the result objects.

lib/mutant/reporter/hash/printer/env_progress.rb
```diff
@@ -0,0 +1,64 @@
...
+            coverage * 100
+          end
+
+          # Overhead in percent
+          #
+          # @return [Float]
+          #
+          # @api private
+          def overhead_percent
```
mbj: Instance of: https://github.com/mbj/mutant/pull/479/files#r44568570

lib/mutant/reporter/hash/printer/env_progress.rb
```diff
@@ -0,0 +1,64 @@
...
+              config: visit(Config, env.config),
+              subjects: amount_subjects,
+              mutations: amount_mutations,
+              kills: amount_mutations_killed,
+              alive: amount_mutations_alive,
+              runtime: runtime,
+              killtime: killtime,
+              overhead: overhead_percent.to_f,
+              coverage: coverage_percent.to_f,
+              expected: (env.config.expected_coverage * 100).to_f
```
mbj: Lets push the float version projection down to the config object instead of duplicating the rational * 100 to_f everywhere.

lib/mutant/reporter/hash/printer/env_result.rb
```diff
@@ -0,0 +1,26 @@
+module Mutant
+  class Reporter
+    class Hash
+      class Printer
+        # Full env result reporter
+        class EnvResult < self
+          delegate(:failed_subject_results)
+          delegate(:subject_results)
```
mbj: Lets combine the 2 delegate calls into one.

lib/mutant/reporter/hash/printer/mutation_result.rb
```diff
@@ -0,0 +1,149 @@
...
+            "Mutated AST:\n" \
+            "%s\n".freeze
+
+          NOOP_MESSAGE    =
+            "---- Noop failure -----\n" \
+            "No code was inserted. And the test did NOT PASS.\n" \
+            "This is typically a problem of your specs not passing unmutated.\n" \
+            "Test Result:\n".freeze
+
+          FOOTER = '-----------------------'.freeze
```
mbj: All the messages here are duplicated, lets push them down to the proposed printer class till we find a better place.

lib/mutant/reporter/hash/printer/mutation_result.rb
```diff
@@ -0,0 +1,149 @@
...
+
+          # Run report printer
+          #
+          # @return [undefined]
+          #
+          # @api private
+          def run
+            {
+              identification: mutation.identification,
```
mbj: I'd align the hash values here.

lib/mutant/reporter/hash/printer/mutation_result.rb
```diff
@@ -0,0 +1,149 @@
...
+            "---- Noop failure -----\n" \
+            "No code was inserted. And the test did NOT PASS.\n" \
+            "This is typically a problem of your specs not passing unmutated.\n" \
+            "Test Result:\n".freeze
+
+          FOOTER = '-----------------------'.freeze
+
+          # Run report printer
+          #
+          # @return [undefined]
```
mbj: Return type is `Hash` not undefined.

lib/mutant/reporter/hash/printer/env_result.rb
```diff
@@ -0,0 +1,26 @@
...
+
+          # Run printer
+          #
+          # @return [undefined]
```
mbj: Return type is `Hash` not `undefined`.

lib/mutant/runner.rb
```diff
@@ -33,6 +33,12 @@ def initialize(*)
     def run_mutation_analysis
       @result = run_driver(Parallel.async(mutation_test_config))
       reporter.report(@result)
+
+      if env.config.json_dump
+        require 'json'
```
mbj: This require should be in `lib/mutant.rb` no dynamic requires.

lib/mutant/reporter/hash/printer/config.rb
jheth: Does  `Matcher::Config` support to_hash or to_json?  What type of serializer are you thinking of?

mbj: An external one, just like you did for everything else here.

mbj:
@jheth I did some style only flags to dig a bit into the code.

While I can confirm it "works" I do not like the implementation approach. Mostly because the `CLI::Printer` thing is filling a StringIO and you create a version that returns someting from `#run`. This is not LSP compatible.

So the base class I wanted to see is not gonna happen.

I think a much easier way to define the serialization of is using [morpher](https://github.com/mbj/morpher). That sadly has not enough documentation for me to recommend to use by anyone else than me.

The benefit is that we get an inverse transformation for free that would allow the combination of partial runs into a bigger result. This would benefit especially @tjchambers use cases.

What we can do with this PR is extracting parts of it, the ones pushing down various projections into the result objects.

lib/mutant/reporter/hash/printer.rb
mbj: I revert that one, see: https://github.com/mbj/mutant/pull/479#issuecomment-155875500

mbj:
I'll close this PR for now, as I think its features cannot be pulled into master as-is without adding significant future development friction due LSP violations, and other design imperfections (that just mirror the design imperfection of the existing CLI reporter, so no blame to your side).

I totally see how it make sense to develop the code that way for a Hackaton that is time constrainted.

To bring this feature to mutant we should do smaller changes like pushing down the redundant projections to result objects. Adding tests for each step.

mbj:
@jheth Thx for this PR it highlighted the need to fix a few things before implementing #370.

lib/mutant/runner.rb
```diff
@@ -33,6 +33,12 @@ def initialize(*)
...
+        hash = Reporter::Hash::Printer::EnvResult.call(@result)
+        File.open(env.config.json_dump, 'w') {|f| f.write(hash.to_json) }
```
mbj: `File.write` is the better option here, does not need a block.

lib/mutant/reporter/hash/printer/mutation_result.rb
```diff
@@ -0,0 +1,149 @@
...
+          end
+
+          # Evil mutation details
+          #
+          # @return [String]
+          #
+          # @api private
+          def evil_details
+            diff = Diff.build(mutation.original_source, mutation.source)
+            diff = false ? diff.colorized_diff : diff.diff
```
mbj: I wounder if it makes sense to colorize the diff reported in the JSON. The webservice consuming this can colorize much better HTML aware.

lib/mutant/runner.rb
```diff
@@ -33,6 +33,12 @@ def initialize(*)
     def run_mutation_analysis
       @result = run_driver(Parallel.async(mutation_test_config))
       reporter.report(@result)
+
+      if env.config.json_dump
```
mbj: I think instead of this branch, there should be a `Reporter::Sequence` that wraps any number of active reporters, calling `#start`, `#progress`, `#result` on each one. 

On the CLI you simply build up a `Reporter::Sequence` object by default containing an instance of `Reporter::CLI` but when the json dumping is active add an `Reporter::JSON` to the collection.

This would give you the same feature without infecting the runner on the special case.

lib/mutant/reporter/hash/printer/mutation_result.rb
mbj: Why do you need to transport this messages to your API? Only theminimal source of truth must be submitted, not the projection.

This question also applies to all other cases where you submit projected values.

mbj:
@jheth And closing this PR does not mean we should stop discussing improvements.

lib/mutant/runner.rb
mbj: To make my point clear, the call to `reporter.report(@result)` just above would cover your use case: https://github.com/jheth/mutant/blob/eab3de25ead132e5d876c09b8570260dbc40c61b/lib/mutant/runner.rb#L35

mbj:
@jheth We have a mutant Slack channel for such discusions, in case you had approached me before I could have saved you significant amounts of time. Do you want an invite for you and your team? If so send me your email addreses.

backus:
@jheth Just to be clear, this **is** something we want to get into mutant and this PR is certainly appreciated.  At the same time mutant has very high standards (checkout [mbj/devtools](https://github.com/mbj/devtools/) at some point). All new PRs (especially from new contributors) take some time and discussion.

Can @mbj send a slack invite to the email listed on your profile?

tjchambers:
@jheth I am just a heavy user of mutant, not much of a contributor. But I am thrilled to have this intellectual focus and contribution. It may not be something that can be PR'ed in as is, but the work that was done is much appreciated by me, and it surfaces some areas that may need prepped before movement can be made in this area (one near to my heart - reporting). 

Just want to add my thanks!

jheth:
Yes, you can send a slack invite to my profile email.  Is there anything you want me to do or are you going to take this over and implement yourself?    This work was done for rails rumble and I put it together quick to serve our purpose of having a consumable format.

mbj:
@jheth Its your decision on if  and how you contribute.

Even if you just throw together this PR that could not be merged as is, but you triggered this discussion you effectively contributed already.

Do not feel obligated to do anything here, I'm happy with what you did so far, basically because it forces me to finally fix this part of mutant.

I think we need to discuss the scope of the feature in channel before deciding about the actual implementaion. 

You dumped out lots of redundant data, like the coverage percentage. Where you could have dumped out the number of mutations and the number of mutations killed.

I see this was "only" for a hackaton and you simply PRed as is. But if we want to have this as a main mutant feature it needs more thought.

Feel invited and thx alread!

mbj:
@jheth In #504 I begun the work to add the sequence reporter we had been talking about, allowing us to avoid adding a json reporter specific call side to the runner.

mbj:
@jheth The sequence reporter is now in master, feel free to base your work on that one if you continue your project. You should be able to plug "any" custom/private/properitary format without having to patch into the released mutant.

PR #483, opened 2015-11-13
Reduce globals from Cache

mbj:
A first commit in an upcoming series of commits where all IO globals like `File`, `Kernel`, `Dir` etc are passsed into the call sides. Allowing much easier specs that do not hook these globals.

The smell moves from `Mutant::Cache` to `Mutant::CLI` next refactoring will lift it to the CLI call sides.


mbj:

mbj:
@dkubb Ping.

lib/mutant/cache.rb
```diff
@@ -21,7 +24,7 @@ def initialize
     # @api private
     def parse(path)
       @cache.fetch(path) do
-        @cache[path] = Parser::CurrentRuby.parse(File.read(path))
+        @cache[path] = Parser::CurrentRuby.parse(file.read(path))
```
dkubb: So this `file` object is actually a `::File` class? Why not have your `path` object be a `Pathname`? That's better than passing around `String` objects to represent a file path.

IMHO I would find wherever these `String` objects are created and turn them into `Pathname` objects. A `String` primitive is rarely the correct object; except maybe where you are dealing with opaque text that is getting logged, certainly not the case here.

lib/mutant/cache.rb
```diff
@@ -1,14 +1,17 @@
...
     #
+    # @param [Class<File>]
+    #
     # @return [undefined]
     #
     # @api private
-    def initialize
+    def initialize(*)
+      super
       @cache = {}
```
dkubb: I realize this isn't part of the changed code, but does this `Hash` need to be accessed from multiple threads? If so you might be better off with a `ThreadSafe::Cache` instance here instead.

dkubb: Also your `#parse` method isn't under review, but you could probably push the cache population into the constructor, eg:

```ruby
def initialize(*)
  super
  @cache = ThreadSafe::Cache.new
end

def parse(path)
  @cache.fetch_or_store(path) do |path|
    Parser::CurrentRuby.parse(file.read(path))
  end
end
```

lib/mutant/cache.rb
dkubb: Also if you use `ThreadSafe::Cache` you can use it's `#fetch_or_store` method to simplify #parse:

```ruby
def initialize(*)
  super
  @cache = ThreadSafe::Cache.new
end

# ...

def parse(path)
  @cache.fetch_or_store(path) do |path|
    Parser::CurrentRuby.parse(file.read(path))
  end
end
```

dkubb: If you do decide to make `path` a `Pathname` object (as I think you should), this whole `Cache` object becomes unnecessary. You can just use a `ThreadSafe::Cache` instance directly and inject that into the Bootstrap instance.

lib/mutant/cache.rb
mbj: @dkubb No all parsing happens in single threaded bootup code. Mutant loads all code upfront.

lib/mutant/cache.rb
mbj: > So this file object is actually a ::File class? Why not have your path object be a Pathname? That's better than passing around String objects to represent a file path.

Its the `::File` singleton, I want to inject it for easier testing as setting up a message expectation on an IO global is a bad idea.

Pathname does not have `Pathname.read(a_file)` interface.

dkubb: > Pathname does not have Pathname.read(a_file) interface.

No, but you can use `path.read`. This code will become:

```ruby
def parse(path)
  @cache.fetch_or_store(path) do |path|
    Parser::CurrentRuby.parse(path.read)
  end
end
```

The expectation you'd have to add to the specs will be even simpler than the current iteration.

dkubb:
@mbj I have one question about `#parse`. It seems odd it's on a class named `Cache`. I would think that the class would be better named as `Parser`. The fact that it has a cache internally seems more like an implementation issue rather than it's core responsibility.

mbj:
> @mbj I have one question about #parse. It seems odd it's on a class named Cache. I would think that the class would be better named as Parser. The fact that it has a cache internally seems more like an implementation issue rather than it's core responsibility.

It should be called `Parser#call` instead, and the parser has caching behavior. Correct.

mbj:
@dkubb Assuming all green this is ready for review again.

lib/mutant/parser.rb
```diff
@@ -12,18 +12,18 @@ def initialize
...
+    # @param [Pathname] path
     #
     # @return [AST::Node]
     #
     # @api private
-    def parse(path)
+    def call(path)
       @cache.fetch(path) do
-        @cache[path] = Parser::CurrentRuby.parse(File.read(path))
+        @cache[path] = ::Parser::CurrentRuby.parse(path.read)
```
dkubb: Why not use `@cache[path] ||= ::Parser::CurrentRuby.parse(path.read)`? What advantage does using `Hash#fetch` bring in this case?

mbj: `||=` is a generic branching construct. Where `Hash#fetch` is a special cased one.

Hence my preference for the latter.

dkubb: Yeah, but the whole AST for this method could be simplified to:

```ruby
def call(path)
  @cache[path] ||= ::Parser::CurrentRuby.parse(path.read)
end
```

I can't think of any metric where that is not better than the current code.

mbj: Yeah, I think the elimination of redundant portion of `@cache[k]` and `@cache[k]=` weights more than my argument above. Changing to your suggestion.

spec/unit/mutant/parser_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+      expect(path).to receive(:read)
+        .with(no_args)
+        .and_return(':source')
+    end
+
+    it 'returns parsed source' do
+      expect(subject).to eql(s(:sym, :source))
+    end
+
+    it 'returns cached parsed source' do
```
dkubb: What about changing the description to something like `'is idempotent'`?

mbj: True.

mbj:
@dkubb This is ready for review again. I needed to add a pass adjusting the mutant configuration a bit.

dkubb:
@mbj this can be merged once the build is green.

mbj:
@dkubb Thx, sorry for not rebasingy by myself after #485. We should have a bot soon ;)

PR #484, opened 2015-11-15
Fix emitter duplication

mbj:


mbj:
@dkubb Ready for review after #483

mbj:
Will repopen that one later.

PR #488, opened 2015-11-15
Drop support for rspec-3.2

mbj:
I do not like to carry around supports for things I do not use by myself. And enlarge CI runtime for no real gain. Updating rspec is trivial.


mbj:
@dkubb Ready for review.

PR #489, opened 2015-11-15
Kill mutations in Mutant::Runner

mbj:
* Prefer attribute reader over ivar
* Reduce intermediary unused abstract class
* Do not build a Mutant::Result::Subject in Hash default
  where an empty mutation result Array is sufficient.

### DIFF
--- a/lib/mutant/runner/sink.rb
+++ b/lib/mutant/runner/sink.rb
@@ -1,107 +1,72 @@
 module Mutant
   class Runner
-    # Abstract base class for computation sinks
     class Sink
-      include AbstractType
+      include Concord.new(:env)
 
-      # sink status
+      # Initialize object
       #
-      # @return [Object]
+      # @return [undefined]
       #
       # @api private
-      abstract_method :status
+      def initialize(*)
+        super
+        @start           = Time.now
+        @subject_results = {}
+      end
 
-      # Test if computation should be stopped
+      # Runner status
+      #
+      # @return [Result::Env]
+      #
+      # @api private
+      def status
+        Result::Env.new(
+          env:             env,
+          runtime:         Time.now - @start,
+          subject_results: @subject_results.values
+        )
+      end
+
+      # Test if scheduling stopped
       #
       # @return [Boolean]
       #
       # @api private
-      abstract_method :stop?
+      def stop?
+        env.config.fail_fast && !status.subject_results.all?(&:success?)
+      end
 
-      # Consume result
+      # Handle mutation finish
       #
-      # @param [Object] result
+      # @param [Result::Mutation] mutation_result
       #
       # @return [self]
       #
       # @api private
-      abstract_method :result
+      def result(mutation_result)
+        subject = mutation_result.mutation.subject
 
-      # Mutation result sink
-      class Mutation < self
-        include Concord.new(:env)
+        @subject_results[subject] = Result::Subject.new(
+          subject:          subject,
+          mutation_results: previous_mutation_results(subject) + [mutation_result],
+          tests:            mutation_result.test_result.tests
+        )
 
-        # Initialize object
-        #
-        # @return [undefined]
-        #
-        # @api private
-        def initialize(*)
-          super
-          @start           = Time.now
-          @subject_results = Hash.new do |_hash, subject|
-            Result::Subject.new(
-              subject:          subject,
-              tests:            [],
-              mutation_results: []
-            )
-          end
-        end
+        self
+      end
 
-        # Runner status
-        #
-        # @return [Status]
-        #
-        # @api private
-        def status
-          env_result
-        end
+    private
 
-        # Test if scheduling stopped
-        #
-        # @return [Boolean]
-        #
-        # @api private
-        def stop?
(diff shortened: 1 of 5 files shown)


mbj:
@dkubb Ready for review.

lib/mutant/runner/sink.rb
```diff
@@ -1,107 +1,72 @@
...
+      end
+
+      # Test if scheduling stopped
       #
       # @return [Boolean]
       #
       # @api private
-      abstract_method :stop?
+      def stop?
+        env.config.fail_fast && !status.subject_results.all?(&:success?)
```
backus: Since `env` is reachable within the `status` object and `status.subject_results.all?(&:success?)` is operating on `status`'s dependencies could this be extracted to a simpler method in `Result::Env`?

lib/mutant/runner/sink.rb
```diff
@@ -1,107 +1,72 @@
...
+      # Initialize object
       #
-      # @return [Object]
+      # @return [undefined]
       #
       # @api private
-      abstract_method :status
+      def initialize(*)
+        super
+        @start           = Time.now
```
backus: Does this go against your desire to pass in top level constants?

mbj: This refactoring was not globally executed yet. I decided to do a global style fix before, that found the uncoverage this PR is addresssing.

mbj: The original code also had this smell. The smell is touched but preserved unmodified.

lib/mutant/runner/sink.rb
mbj: Thats true. Backpressure for this decision is: `Mutant::Result::Env` should not be concerned with any interface a reporter is not interested in.

The `#stop?` only is of concern of the runner.

lib/mutant/runner/sink.rb
dkubb: In the future I would suggest passing in some kind of timer object that you can start, and get the elapsed time.

mbj: @dkubb Not this iteration, out of focus. Its planned to reradicate all direct use of IO globals. But than I found global style fixes are needed before, than I found uncoverage, so this PR exists :P

lib/mutant/runner/sink.rb
dkubb: I wonder if it would be cleaner if you could write this as:

```ruby
env.config.fail_fast && status.subject_results.any?(&:fail?)
```

I don't know if there's a `#fail?` method or equivalent, but it seems a bit simpler.

mbj: There is no `fail?`, I wanted to reach consistency and only have "positive" APIs. I had `#fail?` before on all `Result::*` objects, but it lead to constant churn on using the positive or the negative API at the call sides.

lib/mutant/runner/sink.rb
```diff
@@ -1,107 +1,72 @@
...
-        # @return [Status]
-        #
-        # @api private
-        def status
-          env_result
-        end
+        @subject_results[subject] = Result::Subject.new(
+          subject:          subject,
+          mutation_results: previous_mutation_results(subject) + [mutation_result],
+          tests:            mutation_result.test_result.tests
```
dkubb: Any chance you can change this to `mutation_result.tests`? Even though it'd be delegated through the test result, I think it makes sense to be able to call it direction from the mutation result.

mbj: I'd for now keep it as is, as `Result::Test` and `Result::Mutation` objects are gonna be merged. They have an 1:1 relationship.

mbj: I mainly want to keep the smell "visible", and do not bury it through easier indirection.

mbj:
@dkubb This is ready for the next pass.

lib/mutant/runner/sink.rb
```diff
@@ -1,107 +1,72 @@
...
-            subject_results: @subject_results.values
-          )
-        end
+      # Return previous results
+      #
+      # @param [Subject]
+      #
+      # @return [Array<Result::Mutation>]
+      def previous_mutation_results(subject)
+        subject_result = @subject_results.fetch(subject) { return EMPTY_ARRAY }
```
dkubb: What if instead of this you had a null object to represent empty mutation results? Then you could change this to something like:

```ruby
@subject_results.fetch(subject, Result::Subject::NULL).mutation_results
```

I know how this short circuit works, but it just seems really odd to me.

dkubb: You could even do something like this:

```ruby
def initialize(*)
  # ...
  @subject_results = Hash.new(Result::Subject::NULL)
end 

# ...

def previous_mutation_results(subject)
  @subject_results[subject].mutation_results
end
```

mbj: I thought about that. But I could not find a meaningfull null value for all fields.

Subjects reference:

* A node (to mutate)
* A context (to insert into, AST wise)
* An expresssion (for test selection)
* A source location

For all these I could not imagine good values populating a null object.

PR #490, opened 2015-11-16
Fix the use of instance doubles globally

### LINKED ISSUE #361: Switch to `instance_double` from `double`.

mbj:
[fix #361]

### DIFF
--- a/spec/unit/mutant/context_spec.rb
+++ b/spec/unit/mutant/context_spec.rb
@@ -31,27 +31,5 @@ RSpec.describe Mutant::Context::Scope do
 
       it { should eql(expected) }
     end
-
-    context 'with Class as scope' do
-      let(:scope) { Mutant::Context }
-
-      let(:expected) do
-        s(:class,
-          s(:const, nil, :Context),
-          nil,
-          s(:str, 'test')
-        )
-      end
-
-      it { should eql(expected) }
-    end
-
-    context 'with anything else as scope' do
-      let(:scope) { double(name: 'Foo') }
-
-      it 'raises exception' do
-        expect { subject }.to raise_error(RuntimeError, "Cannot wrap scope: #{scope.inspect}")
-      end
-    end
   end
 end
--- a/spec/unit/mutant/integration/rspec_spec.rb
+++ b/spec/unit/mutant/integration/rspec_spec.rb
@@ -3,8 +3,8 @@ require 'mutant/integration/rspec'
 RSpec.describe Mutant::Integration::Rspec do
   let(:object) { described_class.new(Mutant::Config::DEFAULT) }
 
-  let(:options) { double('options') }
-  let(:runner)  { double('runner')  }
+  let(:rspec_options) { instance_double(RSpec::Core::ConfigurationOptions) }
+  let(:rspec_runner)  { instance_double(RSpec::Core::Runner)               }
 
   let(:example_a) do
     double(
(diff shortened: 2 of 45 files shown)


mbj:
@dkubb Extraction from #486. Ready for review.

spec/unit/mutant/context_spec.rb
```diff
@@ -31,27 +31,5 @@
 
       it { should eql(expected) }
     end
-
-    context 'with Class as scope' do
```
dkubb: Should these spec removals be broken up into a separate commit? I assume they were removed for some other reason?

spec/unit/mutant/integration/rspec_spec.rb
```diff
@@ -115,10 +115,20 @@
...
+    expect(RSpec::Core::ConfigurationOptions).to receive(:new)
+      .with(%w[spec --fail-fast])
+      .and_return(rspec_options)
+
+    expect(RSpec::Core::Runner).to receive(:new)
+      .with(rspec_options)
+      .and_return(rspec_runner)
+
+    expect(RSpec).to receive(:world)
+      .with(no_args)
```
dkubb: Can the `#world` method receive any arguments? If not, then I think specifying `with(no_args)` is redundant.

spec/unit/mutant/context_spec.rb
mbj: Without this removal there is a un-reachable mutation covered branch in code that was selected via `--since` for this commit.

spec/unit/mutant/integration/rspec_spec.rb
```diff
@@ -115,10 +115,20 @@
...
+      .and_return(world)
+
+    allow(Time).to receive(:now)
```
dkubb: In situations where you are receiving and returning a value, the receive_messages method can sometimes be cleaner:

```ruby
allow(Time).to receive_messages(now: Time.now)
```

__NOTE:__ this may apply to other specs in this PR.

spec/unit/mutant/integration/rspec_spec.rb
mbj: Ohh, I thought its required else rspec would silenty throw away arguments.

spec/unit/mutant/integration/rspec_spec.rb
mbj: Will this assert `now` is called with zero arity?

spec/unit/mutant/integration/rspec_spec.rb
dkubb: No, it's mainly for use in methods that receive optional arguments and you want to specify the behaviour of the method when it receives nothing. I've overused it in the past too, and recently begun only asserting it in cases where a method takes optional args only.

mbj: We need inverse mutation testing so we are forced to use the smallest possible set of semanics in the rspec API.

spec/unit/mutant/integration/rspec_spec.rb
dkubb: I don't think it asserts that `Time#now` is called with zero arity, but the method doesn't receive any arguments either. If some code actually attempted to pass an argument to the method it would result in a runtime exception.

IMHO, as mentioned above, it's not necessary to specify `with(...)` except in cases where the method could reasonably accept an argument.

mbj: > If some code actually attempted to pass an argument to the method it would result in a runtime exception.

Thats what I wanted to know. Thx.

dkubb: To further express my point, if we begin to asset a method does not receive arguments when it already cannot, then it's not really different than asserting that a method doesn't receive a block. Technically, I guess you could assert these things, but I think it's redundant.

mbj:
@dkubb Ready for the next pass, flags addressed by code change or comment.

PR #491, opened 2015-11-16
Kill mutations in env

mbj:


mbj:
@dkubb Ready for review after #490. Sorry for PR dependency, but its better to include the global fix here so a simple rebase after merge of dependant PR is sufficient.

mbj:
@dkubb Ready for review.

spec/unit/mutant/env/boostrap_spec.rb
```diff
@@ -155,14 +152,11 @@ def name
...
-        expected_warnings = [
+      before do
+        expected_warning =
           "Class#name from: #{invalid_class.inspect} returned Object. #{Mutant::Env::SEMANTICS_MESSAGE}"
-        ]
 
-        expect { subject }
-          .to change { config.reporter.warn_calls }
-          .from([]).to(expected_warnings)
+        expect(config.reporter).to receive(:warn).with(expected_warning).and_return(config.reporter)
```
dkubb: The formatting here is inconsistent with other changes in this file above.

spec/unit/mutant/env/boostrap_spec.rb
```diff
@@ -92,16 +91,14 @@ def name
         end
       end
 
-      it 'warns via reporter' do
-        expected_warnings = [
+      before do
+        expected_warning =
```
dkubb: I would probably move `expected_warning` into a let statement.

spec/unit/mutant/env/boostrap_spec.rb
```diff
@@ -155,14 +152,11 @@ def name
...
```
dkubb: Instance of https://github.com/mbj/mutant/pull/491/files#r44890780

spec/unit/mutant/env/boostrap_spec.rb
dkubb: Given there are 3 placed in this spec that do the same thing, what about creating a single method that creates the expectation given `expected_warnings` and use it here and elsewhere?

spec/unit/mutant/env_spec.rb
```diff
@@ -45,18 +45,38 @@
...
+          .and_return(mutation_subject)
+
+        expect(context).to receive(:root)
+          .with(s(:nil))
+          .and_return(wrapped_node)
+
+        expect(Mutant::Loader::Eval).to receive(:call)
+          .ordered
+          .with(wrapped_node, mutation_subject)
+          .and_return(nil)
```
dkubb: Is the return value used? If not, why not return a `double`? If it gets used it'll explode. I usually prefer to do this when defining expectations for methods called in "void context".

spec/unit/mutant/env/boostrap_spec.rb
mbj: Good catch. Fixed both.

spec/unit/mutant/env_spec.rb
mbj: This is just a leftover. I replaced it with returning the receiver to mirror what the real `Mutant::Loader::Eval` does.

mbj:

mbj:
@dkubb Assuming all green this is ready for the next pass.

spec/unit/mutant/env/boostrap_spec.rb
```diff
@@ -41,12 +41,19 @@
...
+      .and_return(config.reporter)
+  end
+
   before do
     expect(integration_class).to receive(:new)
       .with(config)
       .and_return(integration)
 
-    expect(integration).to receive(:setup).and_return(integration)
+    expect(integration).to receive(:setup)
```
dkubb: You can use `receive_messages` here.

spec/unit/mutant/env/boostrap_spec.rb
```diff
@@ -146,6 +143,10 @@ def self.name
         end
       end
 
+      let(:expected_warning) do
+        "Class#name from: #{invalid_class.inspect} returned Object. #{Mutant::Env::SEMANTICS_MESSAGE}"
```
dkubb: You wrapped the `expected_warning` above, but not here. I'd recommend wrapping.

spec/unit/mutant/env_spec.rb
```diff
@@ -64,10 +84,18 @@
...
+          .and_raise(Mutant::Isolation::Error, 'test-error')
       end
 
-      let(:test_result) { Mutant::Result::Test.new(tests: tests, output: 'test-error', passed: false, runtime: 0.0) }
+      let(:test_result) do
+        Mutant::Result::Test.new(
+          tests:  tests,
+          output: 'test-error',
+          passed: false,
+          runtime: 0.0
```
dkubb: The hash keys should be aligned.

mbj:
@dkubb Flags addressed, ready for the next pass.

spec/unit/mutant/env/boostrap_spec.rb
```diff
@@ -75,6 +76,11 @@
     context 'when Module#name calls result in exceptions' do
       let(:object_space_modules) { [invalid_class] }
 
+      let(:expected_warning) do
+        "Class#name from: #{invalid_class} raised an error: " \
```
dkubb: Below you use `invalid_class.inspect` but not here. Any reason for the difference?

mbj: Just unaccounted variance. Reducing the other case to the minimum.

mbj:

PR #492, opened 2015-11-16
Prefer method access over ivar

tjchambers:


mbj:
I'm fine with this diff, when the build is green I'll merge.

tjchambers:
BTW I checked for other occurrences of `{@` and found none other than this.

PR #499, opened 2015-11-21
Fix style

mbj:


mbj:
@dkubb This is ready for review after #491

lib/mutant.rb
```diff
@@ -193,22 +192,22 @@ class Config
...
         Expression::Methods,
         Expression::Namespace::Exact,
         Expression::Namespace::Recursive
-      ])
+      ]),
+      fail_fast:         false,
+      includes:          EMPTY_ARRAY,
+      integration:       Integration::Null,
+      isolation:         Mutant::Isolation::Fork,
+      jobs:              Mutant.ci? ? CI_DEFAULT_PROCESSOR_COUNT : ::Parallel.processor_count,
```
dkubb: I'm curious, why bother to limit the processor count in CI? Why not just let it determine how many processors there are and use everything available?

spec/support/rb_bug.rb
```diff
@@ -1,5 +1,6 @@
 require 'ffi'
 
+# @api private
```
dkubb: Do you mean `@private` or `@api private` ?

lib/mutant.rb
dkubb: I'm just trying to understand the scenario where a CI system would make a processor available, but you would explicitly choose to not make use of it. Wouldn't many of the off-the-shelf CI systems have limited numbers of processors anyway? I would imagine most them would have 1 or 2 processors, max. It makes no sense to spin up a container on a CI server with like 8 processors, when most build systems won't properly utilize them anyway.

dkubb:
@mbj I finished a pass over this.

lib/mutant.rb
mbj: We discussed this in the past and the fact you do not recall it tells us we should have an exmplaining comment here.

CI systems run mostly in containerized environments where the host has many more CPUs available than the containers itself can utilize, enforced by resource limits. As there is no standard way to detect the resource limits of a container, and only the hosts resources can be read: Mutant would spin up many more threads above the container quota, trashing and eventually be killed by the CI host.

We simply choose the default of 4 reflecting our usual CircleCI projects. I wounder if we should remove this CI specific default and let everyon figure out by himself the correct jobs setting, only defaulting to the detected CPI count?

As this code is part of the diff, but was not semantically changed, I do not consider this a blocker. I made a dedicated issue to continue the discussion. https://github.com/mbj/mutant/issues/501

spec/support/rb_bug.rb
mbj: @dkubb I meant `@api private`, I assume this is the best way to set a inherited default for `@api` YARD tags on the full mutant namespace. Setting this allows me to remove all the redundant `@api private` tags for individual methods / classes.

Mutant intentionally has no specified public API. The only stable thing is the CLI.

mbj:
@dkubb I adressed all flags, ready for the next pass.

lib/mutant.rb
dkubb: > I wounder if we should remove this CI specific default and let everyon figure out by himself the correct jobs setting, only defaulting to the detected CPI count?

@mbj I like the idea of just setting this to `::Parallel.processor_count` and then allowing people to override it in configuration. It doesn't seem necessary to me to have a CI specific handler for this.

mbj: I remember we both liked the CI specific overwrite in the past, but I've changed my mind recently. I'll add a PR to kill it.

spec/support/rb_bug.rb
mbj: Ohh, I see your flag is not on `lib/mutant.rb` my above comment assumed this location. Still my intention is to set the helpers api to private here.

lib/mutant.rb
mbj: I added #502 to kill this default.

dkubb: > I remember we both liked the CI specific overwrite in the past, but I've changed my mind recently. I'll add a PR to kill it.

Yeah, I weigh the extra complexity vs our needs. Maybe when our CI was "static" it made sense, or maybe I just said "yeah sure" because we had other concerns to focus on. Now that we are moving CI elsewhere, to something a bit more dynamic, I would like to revisit this and make things simpler.

I *think* most of the time our new systems won't "lie" about how many cores are available, and when we can't make it report accurately we can just override it with explicit configuration.

PR #500, opened 2015-11-22
Reduce reference of IO globals

mbj:


mbj:
@dkubb Ready for review after #499, sorry for this chain of dependent PRs, but it makes more sense to base this change on top of the global style fix in #499.

mbj:
> IMHO this approach seems like a mistake. Naming the option with the same name as the core class should be a clear signal this is less than ideal.

I agree that the state this PR is resulting in is clearly showing a smell. Still I think this state is better than before:

The PR addresses the problem that mutant at various location interacted with IO globals, and the only way to spec these was A): Running the IO operation and observing the mostly global state change B) hooking the IO global via a message expectation, resulting in observable inconsistencies under parallelization (rspec does not do this, but xspec will!)

> I think a better approach, here and below, would be to make a simple object that provides a limited interface and uses the core class.

Exactly. That is planned at the next step. But before I want to "lift" the IO globals into the config object, to get a clear "index" on what the wrapping class needs to provide. See this PR as a small intermediary. 

> Sure, this design allows you to inject the core class in specs, but I think this is taking dependency injection too far. I could be wrong about this, but I think you'll find this approach is a dead-end, and a limited scope wrapper is a much cleaner approach long term.

Its much easier to go from "config object holds globals" to "we have a mutant specific IO object we can pass around" than from "IO globals are everywhere" to the desired state.

There will be some other PRs that reduce more IO global use, the code base still references `$stderr` and `File` from classes that do not yet keep state. Fixing these to use the config is the next step, than coming up with the "mutant specific IO wrapper" we discussed.

With that arguments: @dkubb Ready for the next pass.

lib/mutant.rb
```diff
@@ -204,7 +204,11 @@ class Config
       integration:       Integration::Null,
       isolation:         Mutant::Isolation::Fork,
       jobs:              Mutant.ci? ? CI_DEFAULT_PROCESSOR_COUNT : ::Parallel.processor_count,
+      kernel:            Kernel,
```
dkubb: So you've audited the code to make sure it doesn't refer to anything outside of the `Mutant` namespace, besides here of course?

lib/mutant/repository.rb
```diff
@@ -20,17 +20,9 @@ def call(subject)
 
     # Diff between two objects in repository
     class Diff
-      include Adamantium, Concord.new(:from, :to)
+      include Adamantium, Anima.new(:config, :from, :to)
```
dkubb: Do you think the threshold for switching from Concord to Anima is 3 parameters?

I normally wouldn't have a problem with 3 params in a constructor, 4 and I would probably use Anima though (or rethink my object because 4 is probably pushing it).

lib/mutant.rb
mbj: No direct references to `Kernel` exist inside the mutant namespaces anymore, correct. But other IO globals are still referenced and need future iterations, this PR is only the first step.

I extracted this PR out of a bigger change, as other changes need refactorings to be applied first.

lib/mutant/runner.rb
```diff
@@ -37,12 +37,13 @@ def run_mutation_analysis
     #   the last returned status payload
     def run_driver(driver)
       status = nil
+      sleep = env.config.kernel.method(:sleep)
```
dkubb: I would suggest aligning the `=` operators on adjacent lines.

spec/unit/mutant/cli_spec.rb
```diff
@@ -237,7 +237,13 @@
       let(:expected_matcher_config) do
         default_matcher_config.with(
           subject_filters: [
-            Mutant::Repository::SubjectFilter.new(Mutant::Repository::Diff.new('HEAD', 'master'))
+            Mutant::Repository::SubjectFilter.new(
+              Mutant::Repository::Diff.new(
+                config: Mutant::Config::DEFAULT,
+                from: 'HEAD',
```
dkubb: The Hash value alignment is off here.

dkubb:
@mbj I finished a pass over this code.

lib/mutant/repository.rb
mbj: I also base my decision also on: Is there a natural positional order, if not use anima, as arguments are named.

Here I could not imagine how I could infer the correct positioning of arguments if I just knew: "`Diff.new` takes 3 arguments".

spec/unit/mutant/cli_spec.rb
dkubb: Side note, isn't there a rubocop for this? It seems like such an obvious test.

mbj: I've to check rubocop, they ship some cops that are disabled by default, maybe our case is covered.

mbj:

lib/mutant/cli.rb
```diff
@@ -153,14 +164,14 @@ def add_debug_options(opts)
       end
       opts.on('--version', 'Print mutants version') do
         puts("mutant-#{VERSION}")
-        Kernel.exit(EXIT_SUCCESS)
+        config.kernel.exit(EXIT_SUCCESS)
```
dkubb: Isn't the default status code for `Kernel.exit` success? If so, then the `EXIT_SUCCESS` argument isn't required.

lib/mutant/cli.rb
```diff
@@ -153,14 +164,14 @@ def add_debug_options(opts)
...
       end
       opts.on('-d', '--debug', 'Enable debugging output') do
         with(debug: true)
       end
       opts.on_tail('-h', '--help', 'Show this message') do
         puts(opts.to_s)
-        Kernel.exit(EXIT_SUCCESS)
+        config.kernel.exit(EXIT_SUCCESS)
```
dkubb: This is an instance of https://github.com/mbj/mutant/pull/500/files#r45697893

lib/mutant/repository.rb
dkubb: Good point. With things like `from` and `to` as positional arguments sometimes the order would be different depending on the "direction" you are modelling. Having them as Hash arguments make sense.

Although I could argue that `config`, assuming it is required, might be better as the first required argument to the constructor.

dkubb: One other thing to keep in mind. If `config` is a required argument to this, does it make sense for the `Config` object to have helper methods that can be used to create a diff, passing itself into `Diff.new`? I don't know if it does, but I figured I should ask.

dkubb:
@mbj I finished a pass over this PR.

lib/mutant/cli.rb
mbj: true, good catch

dkubb: Is it possible to remove the `EXIT_SUCCESS` constant altogether?

lib/mutant/repository.rb
mbj: The config object is a pure value distribution object. It should not know how to interpret the configured values. The reminder of the Mutant namespace is the "evaluator for the config".

mbj:
Depends on #508

backus:
I noted remaining globals as I looked through mutant. I don't think this is exhaustive and it doesn't cover implicit usage like with Kernel:

- `Mutex` and `ConditionVariable` in mailbox.rb
- `File` and `IO` references in isolation.rb
- `Object` references in expression/method.rb, expression/methods.rb, expression/namespace.rb
- `Marshall` in isolation.rb
- `ObjectSpace` in env/bootstrap.rb
- `Open3` in cli/tput.rb
- `OptionParser` in cli.rb
- `Regexp` in expression/namespace and mutant/mutator/node/named_value/variable_assignment.rb
- `Digest::SHA1` in mutation.rb
- `Set` in mutator, parallel/master, and zombifier.rb
- `StringIO` in reporter/cli/format
- `Time` in env.rb, integration/rspec.rb, and runner/sink.rb

mbj:
@backus Yeah, I'm ware this PR is not exhaustive. Its intend to slice small portions each time I find motivations. In `xspec` I do not want to support setting message expectations on anything else than doubles hence over the time we need to eradicate all the globals we use and set message expectations on.

So not all of the globals in your list have to be changed, as there are AFAIK no message expectations on for example `Set`.

backus:
> Yeah, I'm ware this PR is not exhaustive.

Ah yeah I understood that. I just wanted to have some idea of what is still left and I also wanted to know if you are actually planning to kill all of these constant references too.

mbj:
> I also wanted to know if you are actually planning to kill all of these constant references too.

Longterm yes.Currently I cut the iteration into a small one to get it merged as an iterative improvement before it grew to big. I'll do followups to focus on the globals I've to place message expectations on. This is the blocker for making mutant hosted on `xspec`.

> I just wanted to have some idea of what is still left

NP

dkubb:
@mbj I will be ready to merge this once #508 is merged.

lib/mutant/config.rb
```diff
@@ -13,7 +13,11 @@ class Config
       :includes,
       :isolation,
       :jobs,
+      :kernel,
```
dkubb: @mbj not a blocker, but should this Array of Symbols be defined using `%i[...]`?

PR #502, opened 2015-11-23
Remove ci jobs default

mbj:


mbj:
@dkubb Ready for review after #499

Rakefile
```diff
@@ -6,18 +6,23 @@ Rake.application.load_imports
 
 task('metrics:mutant').clear
 namespace :metrics do
-  task :mutant => :coverage do
-    success = Kernel.system(*%w[
+  task mutant: :coverage do
+    jobs = 4 if ENV.key?('CIRCLE_CI')
```
dkubb: This appears redundant given the code below.

mbj: Yeah, refactoring leftover. Good catch.

Rakefile
```diff
@@ -6,18 +6,23 @@ Rake.application.load_imports
...
       --since HEAD~1
-      --ignore-subject Mutant::Meta*
-      --
-      Mutant*
-    ]) or fail 'Mutant task is not successful'
+      --require mutant
+      --use rspec
+      --zombie
+    ]
+    arguments.concat(%W[--jobs 4]) if ENV.key?('CIRCLE_CI')
```
dkubb: Have you considered having a default of `0` jobs indicate the number of cores should be chosen? This is the vonention used by parallel. If you adopted those conventions, then maybe you could have `--jobs #{ENV.key?('CIRCLE_CI') ? 4 : 0}`

mbj: I do not like this inline signalling. Assinging `0` a special behavior seems to be weird. If you right now boot mutant with `--jobs 0` the system behaves like you'd expect: Boots but does not make progress, as no workers where spawned.

mbj: I simply do not like "C-Style" APIs where integers heavily get abused for control flow triggers. Maybe mutant could support `--jobs detect` or something instead that makes it super clear.

PR #503, opened 2015-11-23
Fix API usage in corpus tests

mbj:
* I HATE DYNLANGS!

### DIFF
--- a/spec/support/corpus.rb
+++ b/spec/support/corpus.rb
@@ -131,7 +131,7 @@ module MutantSpec
         end
         lockfile = repo_path.join('Gemfile.lock')
         lockfile.delete if lockfile.exist?
-        system('bundle install')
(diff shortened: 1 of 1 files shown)


mbj:
@dkubb Ready for review.

spec/support/corpus.rb
```diff
@@ -131,7 +131,7 @@ def install_mutant
         end
         lockfile = repo_path.join('Gemfile.lock')
         lockfile.delete if lockfile.exist?
-        system('bundle install')
+        system(%w[bundle install])
```
dkubb: The `install` comment is optional. You could just say `%w[bundle]` if you wanted.

mbj: Good catch.

mbj:

PR #504, opened 2015-11-23
Add sequence reporter

mbj:
Will allow to reporters together, allowing to add reporters without adding new call sites to the runner code. Related to #479.


spec/unit/mutant/reporter/null_spec.rb
```diff
@@ -2,21 +2,11 @@
...
-
-    it_should_behave_like 'a command method'
-  end
-
-  describe '#progress' do
-    subject { object.progress(value) }
-
-    it_should_behave_like 'a command method'
+      it_should_behave_like 'a command method'
+    end
```
backus: This feels too DRY to me. I think I prefer the previous form where the same form was duplicated for each described method. Part of the reason I feel strongly about this being *too* DRY is that you're using a shared context inherited from devtools. After I parsed what the heck was going on here I felt silly because the tests are making fairly trivial assertions.

mbj: There is no to DRY, my code is ideally reduced to desert. RSpec triggers too redundant syntax anyway. The shorter I can get it the easier to maintain.

This was triggered by the fact I had to copy and paste a block to cover a missed method, while just changeing the message being send to the object. Copy and pasting something that is by 95% equivalent to 3 other instances is not a development / maintenance strategy I like.

The DRY-er the better.

backus: :thumbsup:

backus:
I noted the different in the `Null` reporter interface in slack. Otherwise this PR looks good to me.

mbj:

PR #508, opened 2015-12-19
Fix repo location of rubyspec

mbj:


mbj:
@dkubb Ready for review.

spec/support/corpus.rb
```diff
@@ -134,15 +138,33 @@ def install_mutant
...
       end
 
-      # Not in the docs. Number from chatting with their support.
-      CIRCLE_CI_CONTAINER_PROCESSES = 2
+      # The effective ruby file paths
+      #
+      # @return [Array<Pathname>]
+      def effective_ruby_paths
+        paths = Pathname
+          .glob(repo_path.join('**/*.rb'))
```
backus: Think this glob string could be hoisted up into a constant?

lib/mutant/result.rb
```diff
@@ -203,6 +203,11 @@ def runtime
         test_result.runtime
       end
 
+      # The time spend on killing
```
backus: Typo. Should be "time spent" not "time spend"

spec/support/corpus.rb
mbj: Yes. Just like the other ones.

spec/support/corpus.rb
```diff
@@ -134,15 +138,33 @@ def install_mutant
...
+        end
+      end
 
       # Number of parallel processes to use
       #
       # @return [Fixnum]
       def parallel_processes
-        if ENV['CI']
-          Mutant::Config::DEFAULT.jobs
+        if ENV.key?('CI')
```
backus: Could you instead use `Mutant.ci?` here

backus:
@mbj I finished a pass over this PR

spec/support/corpus.rb
mbj: I could, but `Mutant.ci?` should go away. Mutant itself should not be CI aware, for various reasons. It was left out in https://github.com/mbj/mutant/pull/502, as there is currently no direct option to steer the CLI progress formatter. I'll have to add a CLI option.

A spec suite is allowed to be CI aware, so I did not use this API. Maybe its time to add `MutantSpec.ci?` instead.

backus: I'd be interested in hearing why mutant shouldn't be CI aware. I would agree for a generic application but mutant is specifically used for testing so I would think that continuous integration would be something it could know about.

Regardless, `MutantSpec.ci?` sounds like a good solution if you don't like `Mutant.ci?`

spec/integrations.yml
```diff
@@ -1,14 +1,14 @@
 ---
 - name: rubyspec
   namespace: Rubyspec
-  repo_uri: 'https://github.com/rubyspec/rubyspec.git'
+  repo_uri: 'https://github.com/ruby/rubyspec.git'
```
dkubb: Has development on the `rubyspec/rubyspec` repo stopped while the `ruby/rubyspec` one continued?

mbj: The repo disappeared. Yes. And I found `ruby/rubyspec` to be the successor.

spec/support/corpus.rb
mbj: > I'd be interested in hearing why mutant shouldn't be CI aware. I

Mutant should not read globals. Unless being told so or making this behavor (de-)configurable.

Before #502 mutant was setting very restrictive resource consumption details on detecting the `CI` environment variable. Which is a far to broad decision, as the previous resource consumption decision was based on one possible CI environment (CircleCI).

There is no way such a hard - coded default could possibly be correct for all possible CI environments. So its better to drop it.

Context for CI resource consumption problems: Docker and friends do not provide a way to introspect the *real* amount of threads allocated per container. `/proc/cpuinfo` reflect the hosts CPU without restrictions applied.

So mutant on CircleCI used to spawn 32 threads which quickly ran into resource barriers causing the build to be terminated by their watchdogs.

Also the previous hard limit on jobs under CI detection was set to `4` which depending on domain can be FAR too much or FAR to less, depending on domain under test.

So setting concurrency on CI is really the job of the dev, nothing mutant should do.

Same for everything else that should be environment aware.

mbj:

spec/support/corpus.rb
```diff
@@ -134,15 +138,33 @@ def install_mutant
...
-      # Not in the docs. Number from chatting with their support.
-      CIRCLE_CI_CONTAINER_PROCESSES = 2
+      # The effective ruby file paths
+      #
+      # @return [Array<Pathname>]
+      def effective_ruby_paths
+        paths = Pathname
+          .glob(repo_path.join(RUBY_FILE_PATTERN))
+          .sort_by(&:size)
+          .reverse
```
backus: I know this was in the previous code but I'm curious, why do you sort and reverse here?

spec/support/corpus.rb
backus: Interesting! Makes sense.

spec/support/corpus.rb
mbj: In the following explanation I'll refer to the files in the corpus as item.

On large parallel, blocking (as in all parallel computations have to finish before we can move on) computations we ideally want to have the work most evenly distributed between the concurrent threads and they should *all terminate the same moment*.

So in the worst case one thread out of N is just, before all others finish their last itemstarting to process the *item with the longest computation time*. Adding significantly to the total runtime of the parallel computation in the most inefficient way.

We can avoid this worst case scenario (and also improve the average case!) via making sure that only small items are left when the queue drains, via starting to produce the heavy items early.

Now as part of processing an item in this specific computation we have the problem that we cannot easily estimate the weight of an item. A big file could also be made out of comments that cause no mutations.

But it turns out that using the file size as a cheap to check property is a good approximation to use this optimization.

You can try this for yourself and remove the optimization noticing that we slice some seconds of the overall runtime with this simple thing.

spec/support/corpus.rb
```diff
@@ -10,10 +10,16 @@ module MutantSpec
   #
   # rubocop:disable MethodLength
   module Corpus
+    # Not in the docs. Number from chatting with their support.
+    # 2 processors allocated per container, 4 processes works well.
+    CIRCLE_CI_CONTAINER_PROCESSES = 4
+    TMP                           = ROOT.join('tmp').freeze
```
dkubb: Does the temporary folder need to be under the current root path? Why not just rely on something like `Dir.tmpdir`?

spec/support/corpus.rb
```diff
@@ -134,15 +138,33 @@ def install_mutant
...
+
+        paths - excluded_paths
+      end
+
+      # The excluded file paths
+      #
+      # @return [Array<Pathname>]
+      def excluded_paths
+        exclude.flat_map do |pattern|
+          Pathname.glob(repo_path.join(pattern))
```
dkubb: Would it be possible to join the patterns together into a single string, eg: (untested)

```ruby
Pathname.glob(repo_path.join("{#{pattern.join(',')}}")
```

This would also have the advantage of de-duping the list automatically.

spec/support/corpus.rb
backus: Very interesting. Thanks for explaining

spec/support/corpus.rb
mbj: Initially I wanted to avoid this, as there is no `PathGlob.union` thing, but now as I see it I like it more. As interpolating via `#{}` next to literal `{}` is uneasy to read I think we should go for: `'{%s}' % exclude.join(',')` as this avoids the confusion when parsing the code by brain. Than move the format into a constant.

spec/support/corpus.rb
mbj: My decision criteria for where to place the corpus checkouts are:

a) Persists after a failed corpus test (so I can manually target it for debugging)
b) Easily reachable for manual debugging
c) Easy to implement

`Dir.tmpdir` points to a shared resource, so we have to add code path to deal with the case the most canonical name for a checked out corpus repo is already taken. Defeating b) and c). Creating a unique name is easy via `Dir.mktmpdir` but that violates a) somewhat and also contradicts b).

So I think having the checkouts inside the repo gives us isolation, and easy addressing for manual tests for free so is the overall better choice.

dkubb: Ok, that sounds good.

dkubb:
@mbj I will be ready to merge this once https://github.com/mbj/mutant/pull/508/files#r48119998 is done, and the build is green.

mbj:
> @mbj I will be ready to merge this once https://github.com/mbj/mutant/pull/508/files#r48119998 is done, and the build is green.

I hope I get to it today, all under the "on vacation" indeterminism.

dkubb:
> I hope I get to it today, all under the "on vacation" indeterminism.

Of course, no pressure.

mbj:

spec/support/corpus.rb
```diff
@@ -10,10 +10,18 @@ module MutantSpec
   #
   # rubocop:disable MethodLength
   module Corpus
+    TMP                 = ROOT.join('tmp').freeze
+    EXCLUDE_GLOB_FORMAT = '{%s}'.freeze
+    RUBY_GLOB_PATTERN   = '**/*.rb'
```
backus: Can we freeze this?

mbj: Yes.

mbj:

backus:
You can also remove `core/regexp/shared/new.rb` from the excludes in the rubyspec integration since that file was deleted in https://github.com/ruby/rubyspec/commit/9614f253688c0e2093ed799a00763354b34f330b

spec/support/corpus.rb
```diff
@@ -134,15 +140,31 @@ def install_mutant
...
+      #
+      # @return [Array<Pathname>]
+      def excluded_paths
+        Pathname.glob(repo_path.join(EXCLUDE_GLOB_FORMAT % exclude.join(',')))
+      end
 
       # Number of parallel processes to use
       #
       # @return [Fixnum]
       def parallel_processes
```
dkubb: I realize this isn't changed in this diff, but I wondered why this needs to be an instance method?

spec/support/corpus.rb
```diff
@@ -58,18 +66,16 @@ def verify_mutation_coverage
...
       # @raise [Exception]
       #   otherwise
-      #
-      # rubocop:disable AbcSize
       def verify_mutation_generation
         checkout
         start = Time.now
-        paths = Pathname.glob(repo_path.join('**/*.rb')).sort_by(&:size).reverse
+
         options = {
```
dkubb: Also not changed, but I wondered why the `options` isn't a singleton method, and why it's not frozen.

spec/support/corpus.rb
mbj: Its a helper method, and I do not want to make it part of the public API of the corpus class just to avoid having it as a helper method.

No Public API > "Bad" private API.

spec/support/corpus.rb
mbj: I think for the same reason not an instance method as https://github.com/mbj/mutant/pull/508#discussion_r48656994, not frozen because the last time I tried `parallel` did mutate the input, I'm happy to retry freezing in another iteration, but as I want to use mutants parallel interface in the future likely this code will go away before.

mbj:
> You can also remove core/regexp/shared/new.rb from the excludes in the rubyspec integration since that file was deleted in ruby/rubyspec@9614f25

I wounder if the code should raise when an exclude does NOT trigger. But will delay that for the next iteration.

mbj:
@backus Redundant exclude removed.

Issue #415, opened 2015-08-14
Shared meta examples


backus:
#413 required adding the same example 9 times to `meta/def.rb`: [diff](https://github.com/mbj/mutant/pull/413/files#diff-1d7a1c09650337b70e04322c549fbf6c).

It would be nice if we could specify something like `subset` in the example below:

```ruby
Mutant::Meta::Example.add do
  source 'def foo; end'

  mutation 'def foo; raise; end'
end

Mutant::Meta::Example.add do
  source 'def foo; foo; rescue; end'
  subset 'def foo; end'

  mutation 'def foo; nil; rescue; end'
  mutation 'def foo; self; rescue; end'
  mutation 'def foo; end'
  mutation 'def foo; foo; end'
end
```

where `subset` indicates that the mutations from the first example should also show up in the second example.

Issue #395, opened 2015-08-05
Criteria/Classification for Mutations?


tjchambers:
I was reading through the source code in an effort to better understand the classes of mutations and I got to wondering what the philosophical criteria was for choosing/implementing mutations. Basically the question that I ended up having is what makes a suitable mutation?

In my experience with mutant I synthesized some categories (incomplete):

- Promotion to greater *strength* (== -> eql? or equal?)
- Omission (remove an element in a parameter list or array and deduce if it is critically tested)
- Boundary checking (>= -> == or >)
- Ordering (reverse_map => map)
- Numeric negation (1 -> -1)
- Predicate replacement (predicate -> false, self, nil)

I am certain this is a extremely rough and surely incomplete taxonomy. 

One area where I fail to see mutation types (at least so far) is what I would call *opposition*. Not being schooled on mutation testing per se, I don't see many replacements of one method with it's diametric opposite. For example (and vice versa):

> to < or <=
#select to #reject
#odd? to #even? 

My inference is that most of these would perhaps NOT catch bugs (hopefully), and some could exacerbate infinite looping. 

Thoughts?

dkubb:
Could another term for this be an "inverse function"?

mbj:
I call them: Semantically reducing operators. And Orthogonal operators.

`#[]` to `#fetch` is an instance of "reduction operator".
`>` to `<` is an orthogonal one.

mbj:
At the source code level there is no distinction being made. As I do not see a reason to do so. It *could* be included in the report, but I doubt it would help anyone.

mbj:
The classes @tjchambers made up are more fine grained, but I think they all fit into `reduction` vs `orthogonal`.

tjchambers:
@dkubb Sounds clearer to me that my term. 

@mbj Orthogonal is good too. 

I guess my question is really - is there a reason to NOT perform these orthogonal mutations?

mbj:
@tjchambers Orthogonal operators are the ones where most equivalent mutations are coming from. But they are very helpful for example turning literals into others, example `true` to `false`.

dkubb:
I think a clear set of terms to describe the different classifications would benefit mutation testing since all authors have a common vocabulary in which to share ideas. Right now there is some agreement, but it's never been formalized.

mbj:
[...]

PR #510, opened 2015-12-25
Add mutation from `Hash#values_at` → `Hash#fetch_values`

backus:
Depends on #508

### DIFF
--- a/meta/send.rb
+++ b/meta/send.rb
@@ -224,6 +224,22 @@ Mutant::Meta::Example.add do
 end
 
 Mutant::Meta::Example.add do
+  source 'foo.values_at(a, b)'
+
+  singleton_mutations
+  mutation 'foo.fetch_values(a, b)'
+  mutation 'foo'
+  mutation 'self.values_at(a, b)'
+  mutation 'foo.values_at(a)'
+  mutation 'foo.values_at(b)'
+  mutation 'foo.values_at(nil, b)'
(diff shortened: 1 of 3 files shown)


backus:
@mbj this is ready for review

meta/send.rb
```diff
@@ -224,6 +224,22 @@
 end
 
 Mutant::Meta::Example.add do
+  source 'foo.values_at(a, b)'
```
dkubb: @backus In the tests for #511 I noticed you used `self.` instead of `foo.`. Should we use one form for all new mutations consistently?

Also, it's probably outside of scope, but a future PR should probably make sure it's consistent across the board.

backus: True. This should probably be `self` too then.

Yes. In general I think that we can significantly improve meta by making it more consistent and incorporating https://github.com/mbj/mutant/issues/415 and https://github.com/mbj/mutant/issues/395

mbj:
@backus I'm ready to merge after rebase.

backus:
@mbj rebased

meta/send.rb
mbj: I think that we can reduce the verbosity of the meta DSL a bit, so that you do not have to re-specify "adjacent mutations" that have to be present to make the source of the "mutation under specification" valid.

I avoided doing this until its painful enough, IMO the "its painful enough" state is reached.

PR #511, opened 2015-12-25
Add mutations from `Hash#[]` and `Hash#fetch` to `Hash#key?`

backus:
Depends on #508

### DIFF
--- a/lib/mutant/mutator/node/send.rb
+++ b/lib/mutant/mutator/node/send.rb
@@ -26,7 +26,7 @@ module Mutant
           to_i:          %i[to_int],
           to_a:          %i[to_ary],
           to_h:          %i[to_hash],
-          at:            %i[fetch],
+          at:            %i[fetch key?],
           fetch:         %i[key?],
           :[] =>         %i[at fetch key?],
           :== =>         %i[eql? equal?],
--- a/lib/mutant/mutator/node/send.rb
+++ b/lib/mutant/mutator/node/send.rb
@@ -27,7 +27,7 @@ module Mutant
           to_a:          %i[to_ary],
           to_h:          %i[to_hash],
           at:            %i[fetch],
-          :[] =>         %i[at fetch],
+          :[] =>         %i[at fetch key?],
           :== =>         %i[eql? equal?],
           :>= =>         %i[> == eql? equal?],
           :<= =>         %i[< == eql? equal?],
--- a/lib/mutant/mutator/node/send.rb
+++ b/lib/mutant/mutator/node/send.rb
@@ -27,6 +27,7 @@ module Mutant
           to_a:          %i[to_ary],
           to_h:          %i[to_hash],
           at:            %i[fetch],
+          fetch:         %i[key?],
           :[] =>         %i[at fetch key?],
           :== =>         %i[eql? equal?],
           :>= =>         %i[> == eql? equal?],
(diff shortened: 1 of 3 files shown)


backus:
@mbj this is ready for review

lib/mutant/mutator/node/send.rb
```diff
@@ -27,7 +27,8 @@ class Send < self
           to_a:          %i[to_ary],
           to_h:          %i[to_hash],
           at:            %i[fetch],
```
dkubb: I realize this is probably outside of the scope, but should this also include `%i[fetch key?]` to be consistent with below?

backus: Doesn't seem out of scope. I agree I think if we're mutating `:[]` and `:fetch` that means `:at` should be mutated as well

lib/mutant/mutator/node/send.rb
```diff
@@ -26,8 +26,9 @@ class Send < self
           to_i:          %i[to_int],
           to_a:          %i[to_ary],
           to_h:          %i[to_hash],
-          at:            %i[fetch],
-          :[] =>         %i[at fetch],
+          at:            %i[fetch key?],
+          fetch:         %i[key?],
+          :[] =>         %i[at fetch key?],
```
dkubb: @mbj I wonder if in the future that instead of defining a hash mapping the method to possible alternatives, we just defined a list of lists where you'd list a series of methods, ranked from most to least powerful.

So in this case, instead of having entries for `#at`, `#fetch` and #[]` we could just write:

```ruby
[
  # ...
  `%i[ [] at fetch key? ]`
  # ...
]
```

Of course under the hood you could have some kind of index to make it efficient to lookup a specific method and find less powerful alternatives, but it would eliminate duplicate information like we have now.

backus: @dkubb You could even maybe just define the mapping in terms of one method name to another method name like so

```ruby
{
  reverse_map:   :map,
  map:           :each,
  kind_of?:      :instance_of?,
  is_a?:         :instance_of,
  reverse_each:  :each,
  reverse_merge: :merge,
  send:          :__send__,
  __send__:      :public_send,
  gsub:          :sub,
  eql?:          :equal?,
  to_s:          :to_str,
  to_i:          :to_int,
  to_a:          :to_ary,
  to_h:          :to_hash,
  at:            :fetch,
  fetch:         :key?,
  :[] =>         :at,
  :== =>         :eql?,
  :>= =>         :>,
  :>  =>         :==,
  :<= =>         :<,
  :<  =>         :==
}
```

with that you can then infer that a set of mappings like this

```ruby
  :[] =>         :at,
  at:            :fetch,
  fetch:         :key?,
```

means that `:[]` is semantically simplified to `:at`, `:at` to `:fetch`, and `:fetch` to `:key?`.

mbj: @backus Ideally we do a "bigger" refactoring where a mutation operator only goes "one level", and the mutation engine re-injects the mutant to generate higher levels.

For now I only do higher levels on certain explicit cases, like when expanding `#[]` to `#at` and `#fetch`.

For now I think we should keep the explicitness in higher level, and think about global higher levels at the emitter root level.

mbj: @dkubb ^^ In this particular case I trade verbosity for the time it takes to work out a "real" high level mutation strategy. Lets keep the verbosity for now and do not implement a half backed high level strategy for a specific emitter, instead later remove the explicitness when a global strategy exists.

PR #514, opened 2016-01-09
Fix coverage of Mutant::Result namespace

mbj:


mbj:

mbj:
Note ready for review, as there is uncoverage exposed in something I touched I did not explicitly manually tested for. `--since` is great! Gonna fix that one and re-request the review.

mbj:
This is ready for review again.

lib/mutant/mutation.rb
```diff
@@ -40,11 +40,11 @@ def original_source
 
     # Test if mutation is killed by test reports
     #
-    # @param [Array<Report::Test>] test_reports
+    # @param [Result::Test] test_result
     #
     # @return [Boolean]
-    def self.success?(test_result)
-      self::TEST_PASS_SUCCESS.equal?(test_result.passed)
+    def success?(test_result)
```
dkubb: Why promote this to an instance method? It doesn't use any instance method state so should probably remain as a class method.

mbj: It is de-facto using instance state `self.class::` for looking up a constant relative to the instances class.

That we factored this out in a constant is just implementation detail and should not leak into an ugly public interface: https://github.com/mbj/mutant/pull/514/files#diff-e4388dd312e80786e0cc9aaa0272e00bL217

An ugly public interface that was uneasy to spec because it would have required many hops in method expectation chains to do so.

Also conceptually its the mutation *instances* decision if a given test result kills that mutation. Its just by accident we do not *yet* use other information than the `Result::Test#pass`. In other experiments I want to base of this code (some of these are not in this repo) also the time spend on the mutation is relevant relative to the subjects (reachable from the mutation instance) configuration.

dkubb: > It is de-facto using instance state self.class:: for looking up a constant relative to the instances class.

Well, sort of. It doesn't require instance state. You could make due with class state.

Also I wasn't factoring in the constant at all. That isn't relevant imho. My primary consideration is if actual instance state is used or not. If it's not, then the method should be a singleton method.

My point in flagging this is that when you could rewrite the method as the following and still have it pass, then maybe it shouldn't be an instance method:

```ruby
def success?(test_result)
  self.class.success?(test_result)
end

def self.success?(test_result)
  self::TEST_PASS_SUCCESS.equal?(test_result.passed)
end
```

You could also just use `delegate :success?, to: self`if you had a delegation helper.

spec/unit/mutant/mutation_spec.rb
```diff
@@ -1,10 +1,16 @@
...
+  let(:mutation_class) do
+    Class.new(Mutant::Mutation) do
+      const_set(:SYMBOL, 'test'.freeze)
+      const_set(:TEST_PASS_SUCCESS, true)
+    end
   end
 
-  let(:object)  { TestMutation.new(mutation_subject, Mutant::AST::Nodes::N_NIL) }
-  let(:context) { instance_double(Mutant::Context)                              }
+  let(:context)        { instance_double(Mutant::Context) }
```
dkubb: The formatting around the curly braces looks a bit off I think.

lib/mutant/mutation.rb
mbj: > Also I wasn't factoring in the constant at all. That isn't relevant imho. My primary consideration is if actual instance state is used or not. If it's not, then the method should be a singleton method.

As a general guideline, yes. But there are great examples to derivate from the general rule.

I care more about the clean-ness of the public API than the clean-ness of the implementation of the public API!

Hence I prefer a call side: `mutation.success?(test_result)` over `mutant.class.success?(test_result)`.

Since having to remember (and document / test) the 2nd one is a nightmare.

From the call sides POV I want to know if "this test result kills that specific mutation" and I'm not at all interested in the detail if for all instances of that mutation class the "killed mutation" decision is done the same.

Also as I noted I've a private branch that uses instance specific state to determine the kill decsion, one that depends on this change.

There are many good examples "algebraic OO components" as they are building an algebra need to violate the "has to use instance state rule".

Cases like:

```ruby
def success?(_input)
  false
end
```

Are well justified when part of an LSP compatible class hierarchy where "dynamically (with instance state) computated" versions exit.

Examples of the "keep the instance API well formed, but delegate to singleton for deduplication" are to be found in another popular library that implements something more complex than this cases small "mutant result algebra":

* https://github.com/dkubb/axiom/blob/master/lib/axiom/aggregate.rb#L36-L46

dkubb: > Also as I noted I've a private branch that uses instance specific state to determine the kill decision, one that depends on this change.

This is all the argument you needed :P

PR #515, opened 2016-01-09
Fix coverage of Mutant::AST namespace

mbj:


mbj:

spec/unit/mutant/ast/named_children_spec.rb
```diff
@@ -42,8 +42,18 @@
...
       describe '#bar' do
-        it 'returns named child bar' do
-          expect(instance.bar).to be(node_bar)
+        context 'when node is present' do
+          it 'returns named child bar' do
+            expect(instance.bar).to be(node_bar)
+          end
+        end
+
+        context 'when node is NOT present' do
```
backus: Does 'NOT' need to be all caps here?

mbj: I reflect how I typically use to communicate negations in a development context, to make sure these are not missed. 

Maybe it something specific to the fact I'm not a native speaker, and fear to miss writing or reading negations wrong.

Also I realize how I'm inconsistent in the existing specs. I'm fine to do a followup to bring it in sync as soon we decided about the best strategy.

backus: :thumbsup: yeah I don't think it is a big deal either way, just probably good to be consistent

dkubb: In general I don't think it's necessary to emphasize `NOT` in this way but I don't feel that strongly about it.

I do with there was a convention to emphasize things like this in a spec description. Perhaps wrapping the word in asterisks?

dkubb:
The one outstanding question in this PR isn't a blocker so I am going to merge this now.

PR #516, opened 2016-01-09
Fix coverage of Mutant::Context namespace

mbj:
* Fix Mutant::Context::Scope as only subclass of abstract
  Mutant::Context via onlining Scope semantics into Context.


mbj:
@dkubb, @backus this is ready for review.

lib/mutant/context.rb
```diff
@@ -1,19 +1,92 @@
...
+        s(:module, name, node)
+      end
+    end
+
+    # Nesting of scope
+    #
+    # @return [Enumerable<Class,Module>]
+    def nesting
+      const = Object
+      name_nesting.each_with_object([]) do |name, nesting|
```
dkubb: Would it be possible to use `Enumerable#map` here instead?

lib/mutant/context.rb
```diff
@@ -1,19 +1,92 @@
...
+      name_nesting.last
+    end
+
+    # Match expressions for scope
+    #
+    # @return [Enumerable<Expression>]
+    def match_expressions
+      name_nesting.each_index.reverse_each.map do |index|
+        Expression::Namespace::Recursive.new(
+          scope_name: name_nesting.take(index.succ).join(NAMESPACE_DELIMITER)
```
dkubb: Do you have any line length limitations specified in rubocop? This seems to exceed 79 chars.

spec/unit/mutant/context_spec.rb
```diff
@@ -32,4 +32,88 @@
...
+  describe '#identification' do
+    subject { object.identification }
+
+    it { should eql(scope.name) }
+  end
+
+  describe '#root' do
+    subject { object.root(node) }
+
+    let(:node)  { s(:sym, :node)   }
```
dkubb: The spacing around the curly braces looks a bit off here.

lib/mutant/context.rb
mbj: We have, but for historic reasons the line length limit is quite large. Rubocop was added "late" to this project.

I'll reduce the line length here and also the rubocop limit in case this is possible without adding tons of excludes.

lib/mutant/context.rb
mbj: We can. And I acutally thought about this and threw the idea away because it feels counter intuitive to rely on iteration `N - 1` to produce result for `N`. AKA the body mutates an lvar.

But once I wrote it down as per your suggestion, it does not feel so weird so I'm okay with `#map`.

lib/mutant/context.rb
mbj: This is at 77chars ;)

mbj:

lib/mutant/context.rb
dkubb: I agree that the body mutating an ivar is a bit weird, but that was already happening within the `#each_with_object` block, so I didn't see changing to `Enumerable#map` as a decrease in code quality.

lib/mutant/context.rb
dkubb: > This is at 77chars ;)

Hah. I must admit I eyeballed this one. It just *looked* much longer than 79 chars.

lib/mutant/context.rb
mbj: yeah, I only spotted this after "seeing" the code. BTW: Its an lvar not an ivar, if we'd modify an ivar from `#map` I'd really hate it.

We might be able to combine map with reduce here if we where on Haskell ;)

lib/mutant/context.rb
mbj: This is because compared to the reminder falls out. I also eyballed into agreeing you ;)

lib/mutant/context.rb
```diff
@@ -1,19 +1,91 @@
...
+        s(:module, name, node)
+      end
+    end
+
+    # Nesting of scope
+    #
+    # @return [Enumerable<Class,Module>]
+    def nesting
+      const = Object
+      name_nesting.map do |name, nesting|
```
dkubb: The `nesting` variable isn't going to be used in this block.

spec/unit/mutant/context_spec.rb
```diff
@@ -32,4 +32,88 @@
...
+  describe '#identification' do
+    subject { object.identification }
+
+    it { should eql(scope.name) }
+  end
+
+  describe '#root' do
+    subject { object.root(node) }
+
+    let(:node)  { s(:sym, :node) }
```
dkubb: Formatting looks a bit off here.

lib/mutant/context.rb
mbj: Mutant took it out already hence this branch is failing :P

PR #517, opened 2016-01-09
Remove Mutant::Delegator

mbj:
* And inline its semantics into the only inclusion side
* Add mutation covered specs


mbj:

lib/mutant/reporter/cli/printer.rb
```diff
@@ -3,10 +3,34 @@ class Reporter
...
+                Adamantium::Flat,
+                Concord.new(:output, :object),
+                Procto.call(:run)
 
         private_class_method :new
 
+        # Create delegators to object
+        #
+        # @return [undefined]
+        def self.delegate(*names)
```
dkubb: Are there other places that use delegation within the codebase? If so should this be extracted into a module and used here?

Even if it's not used anywhere else, this delegation plumbing code might be better off separated from the system domain.

dkubb: Heh, I realize now you removed `Mutant::Delegator` for this PR. Why did you do that?

I think it's fine to separate plumbing code from domain knowledge.

mbj: This change is about to in-line the `Mutant::Delegator` with its only call sides. It was a module as I thought delegation to private `#object` would be a more common pattern. But it only turned out to exist in the `Mutant::Reporter::CLI::Printer*` namespace so I decided to remove `Mutant::Delegator` with this PR.

mbj: > Heh, I realize now you removed Mutant::Delegator for this PR. Why did you do that?
> 
> I think it's fine to separate plumbing code from domain knowledge.

Short reason: It was easier to mutation cover at the call side than mutation covering the extracted module.

dkubb: Oh I see. It's delegating to the `object`. You *could* make this an explicit parameter to `delegate` and generalize it.

mbj: Once I've got a 2nd call side outside the `Mutant::Reporter::CLI::Printer*` namespace I'll do that.

PR #518, opened 2016-01-09
Fix Mutant::Diff coverage

mbj:


mbj:

lib/mutant/diff.rb
```diff
@@ -74,19 +72,15 @@ def hunks
...
-      tail.each_with_object([head]) do |right, aggregate|
-        left = aggregate.last
-        if right.overlaps?(left)
-          right.merge(left)
-          aggregate.pop
-        end
-        aggregate << right
+      tail.reduce(head) do |left, right|
+        right.merge(left)
+        right
```
mbj: Proactive comment: Yes `Diff::LCS::Hunk#merge` returns `nil` but modifies receiver in-place.

lib/mutant/diff.rb
```diff
@@ -74,19 +72,15 @@ def hunks
...
       head, *tail = hunks
 
-      tail.each_with_object([head]) do |right, aggregate|
-        left = aggregate.last
-        if right.overlaps?(left)
-          right.merge(left)
-          aggregate.pop
-        end
-        aggregate << right
+      tail.reduce(head) do |left, right|
```
dkubb: I would suggest rewriting this as:

```ruby
def minimize_hunk
  hunks.reduce do |left, right|
    left.merge(right)
    left
  end
end
```

PR #519, opened 2016-01-10
Fix integration coverage

mbj:


mbj:

lib/mutant/integration.rb
```diff
@@ -4,38 +4,23 @@ module Mutant
...
-    # @param [String] name
+    # @param kernel [Kernel]
+    # @param name [String]
     #
-    # @return [Integration]
-    #   if found
-    def self.lookup(name)
-      REGISTRY.fetch(name)
+    # @return [Class<Integration>]
+    def self.setup(kernel, name)
```
backus: I think as long as you are doing dependency injection this way, things like `kernel` should always be passed as keyword arguments. Otherwise it seems like `kernel` being a positional argument will often just be surprising.

spec/unit/mutant/integration_spec.rb
```diff
@@ -10,6 +10,20 @@
     subject { object.setup }
     it_should_behave_like 'a command method'
   end
+
+  describe '.setup' do
+    subject { described_class.setup(kernel, name) }
+
+    let(:name)   { 'null'                                  }
+    let(:kernel) { instance_double(Kernel.singleton_class) }
```
backus: is `instance_double(Kernel.singleton_class)` functionally different from `class_double(Kernel)`?

lib/mutant/integration.rb
```diff
@@ -4,38 +4,23 @@ module Mutant
   class Integration
     include AbstractType, Adamantium::Flat, Concord.new(:config)
 
-    REGISTRY = {}
-
     # Setup integration
     #
-    # @param [String] name
+    # Integrations are supposed to define a constant under
+    # Mutant::Integration named after tha capitalized +name*
```
backus: Typo: should be "the" not "tha"

backus:
@mbj I finished doing a pass over this PR

spec/unit/mutant/integration_spec.rb
mbj: AFAIK no. I simply was not aware of `class_double`.

lib/mutant/integration.rb
mbj: I do not use kwargs yet in this code base. Also this is an internal API right now a positional argument is fine for me, given this only takes 2 args.

spec/unit/mutant/integration_spec.rb
dkubb: :+1: for `class_double(Kernel)`. I didn't know about that one either.

mbj:
@backus Ready for the next pass.

mbj:
I lost an important commit in this that fixed the require highjacker to also catch requires on `Kernel` with explicit receiver.

We need to get a rebase bot so I do not "crapify" PRs on rebasing.

backus:
> I lost an important commit in this that fixed the require highjacker to also catch requires on Kernel with explicit receiver.

Can you not recover it with reflog?

mbj:
> Can you not recover it with reflog?

yeah, I meant lost in sense of not present here anymore. I wanted to explain the build failure.

backus:
> yeah, I meant lost in sense of not present here anymore. I wanted to explain the build failure.

Ah ok that makes more sense

mbj:
@dkubb This is ready for review after #532

lib/mutant/integration.rb
```diff
@@ -4,38 +4,23 @@ module Mutant
...
+    # @param kernel [Kernel]
+    # @param name [String]
     #
-    # @return [Integration]
-    #   if found
-    def self.lookup(name)
-      REGISTRY.fetch(name)
+    # @return [Class<Integration>]
+    def self.setup(kernel, name)
+      kernel.require("mutant/integration/#{name}")
```
ch1c0t: Is `kernel` really necessary here? Why not just use `Kernel#require` with implicit receiver?

mbj: For dependency injection reasons. I do not want to refer to any "IO global" in the mutant code base for various reasons:

* Its harder to test
* I can inject metric tracer objects on debugging
* I do want to inject objects that do something fundamentally different in future, these are in private branches right now, but that behavior depends on "perfect" hickacks of interaction with `Kernel`.

PR #531, opened 2016-02-07
Add `#dig` mutation

backus:


mbj:
@backus You probably know alreay, but as there is no indication of this branch is a WIP, the build fails with alive mutation:

```
 def emit_dig_mutation
-  unless (selector.equal?(:dig) && arguments.any?)
+  unless (selector.equal?(:dig) && arguments)
     return
   end
   fetch_mutation = s(:send, receiver, :fetch, arguments.first)
   if arguments.one?
     return emit(fetch_mutation)
   end
   emit(s(:send, fetch_mutation, :dig, *arguments.drop(1)))
 end
-----------------------
evil:Mutant::Mutator::Node::Send#emit_dig_mutation:/home/ubuntu/mutant/lib/mutant/mutator/node/send.rb:127:c0fdd
@@ -1,11 +1,11 @@
 def emit_dig_mutation
-  unless (selector.equal?(:dig) && arguments.any?)
+  unless (selector.equal?(:dig) && self)
     return
   end
   fetch_mutation = s(:send, receiver, :fetch, arguments.first)
   if arguments.one?
     return emit(fetch_mutation)
   end
   emit(s(:send, fetch_mutation, :dig, *arguments.drop(1)))
 end
-----------------------
evil:Mutant::Mutator::Node::Send#emit_dig_mutation:/home/ubuntu/mutant/lib/mutant/mutator/node/send.rb:127:dfca1
@@ -1,11 +1,11 @@
 def emit_dig_mutation
-  unless (selector.equal?(:dig) && arguments.any?)
+  unless selector.equal?(:dig)
     return
   end
   fetch_mutation = s(:send, receiver, :fetch, arguments.first)
   if arguments.one?
     return emit(fetch_mutation)
   end
   emit(s(:send, fetch_mutation, :dig, *arguments.drop(1)))
 end
```

I think you have to add some counter examples for these mutations. As `foo.dig` case is not covered right now.

backus:
@mbj addressed flags. Ready for another pass

mbj:
Assuming all green I'm ready to merge.

lib/mutant/mutator/node/send.rb
```diff
@@ -117,6 +118,22 @@ def emit_receiver_selector_mutations
...
+        # Emit mutation for `#dig`
+        #
+        # - Mutates `foo.dig(a, b)` to `foo.fetch(a).dig(b)`
+        # - Mutates `foo.dig(a)` to `foo.fetch(a)`
+        #
+        # @return [undefined]
+        def emit_dig_mutation
+          return unless selector.equal?(:dig) && arguments.any?
+
+          fetch_mutation = s(:send, receiver, :fetch, arguments.first)
```
dkubb: What about:

```ruby
def emit_dig_mutation
  return unless selector.equal?(:dig) && arguments.any?
  head, *tail = arguments
  fetch_mutation = s(:send, receiver, :fetch, head)
  return emit(fetch_mutation) if tail.empty?
  emit(s(:send, fetch_mutation, :dig, *tail))
end
```

lib/mutant/mutator/node/send.rb
```diff
@@ -117,6 +118,22 @@ def emit_receiver_selector_mutations
...
         end
 
+        # Emit mutation for `#dig`
+        #
+        # - Mutates `foo.dig(a, b)` to `foo.fetch(a).dig(b)`
+        # - Mutates `foo.dig(a)` to `foo.fetch(a)`
+        #
+        # @return [undefined]
+        def emit_dig_mutation
+          return unless selector.equal?(:dig) && arguments.any?
```
mbj: I wounder if for composed conditions `unless` is the best choice?

`return if !selector.equal? || arguments.none?`

might be easier to read.

lib/mutant/mutator/node/send.rb
backus: Heh I actually had exactly this code with different naming. The flay score was a bit higher and I didn't feel like it bought me much. I'm currently ambivalent so I'd be interested in hearing an argument for why this code is better.

lib/mutant/mutator/node/send.rb
backus: Hm. Seems logical. The guard is currently (effectively) `return if !(a && b)` so the transformation to `return if (!a || !b)` seems preferable.

lib/mutant/mutator/node/send.rb
dkubb: One reason I think it's better is that it replaces method calls with language features. Instead of calling methods to find the head and tail of the list it's using pattern matching to extract the values. I tend to prefer using built-in syntax over method calls.

backus: Hm ok I think I can agree with that preference. As devils advocate though: I think multiple left hand assignment might pack a bit more potential for surprise since it depends on the type being an array in this case. Even if arguments became an enumerable instance which behaved like an array this syntax would change:

```ruby
require 'concord'

class Arguments
  include Enumerable, Concord.new(:nodes)

  def each(&blk)
    nodes.each(&blk)
  end
end

arguments = [1,2,3]

# interface
arguments.first   # => 1
arguments.drop(1) # => [2, 3]

# mlhs
a1, *b1 = arguments
a1 # => 1
b1 # => [2, 3]

arguments = Arguments.new([1,2,3])

# interface
arguments.first   # => 1
arguments.drop(1) # => [2, 3]

# mlhs
a2, *b2 = arguments
a2 # => #<Arguments nodes=[1, 2, 3]>
b2 # => []
```

lib/mutant/mutator/node/send.rb
mbj: Its not entirely pushing the negation down literally, since the negation can be consumed by the operator turning `arguments.any?` into `arguments.none?`. Where I do think we should use `arguments.empty?` as this one has less semantics (does not take a block, should also be a mutation).

The preference comes from that parsing an `unless` when the condition is composed is IMO very hard to read, especially when the negations can be pushed down into using not literally negated sub-conditions.

I think that for that reason `unless` should only be used in guard form, and than with a conditional that is not composed.

backus: Agree

mbj:
@backus Do you mind to rebase and fix the last flag? I'd like to include this PR into the next release.

I really wounder why github does not make it easy to change the branch behind a PR, so I could take yours, fix the last flag and re-use this PR with the branch I'd push up. Not burning the history of this branch via closing it and opening another one.

backus:
@mbj should be rebased and updated.

backus:
@mbj all green

PR #532, opened 2016-02-14
Fix require highjack to also work on singleton

mbj:


mbj:
@dkubb Ready for review.

lib/mutant/require_highjack.rb
```diff
@@ -10,10 +10,14 @@ module RequireHighjack
     # @return [#call]
     #   the original implementation on singleton
     def self.call(target, callback)
-      target.method(:require).tap do
+      target.public_method(:require).tap do
         target.module_eval do
+          undef_method(:require)
           define_method(:require, &callback)
-          public :require
+          class << self
```
dkubb: Are you able to use `singleton_class.undef_method(:require)` here instead?

mbj: I did not assume this is public. If its public its better to use this form.

dkubb: > I did not assume this is public

I *think* it is. The Ruby documentation doesn't *usually* have any entries for private methods afaik.

mbj: Counter example: 

```
     NoMethodError:
       private method `undef_method' called for #<Class:#<Module:0x0056106133bf90>>
```

mbj: Should we prefer:

`singleton_class.send(:undef_method, :require)` over the `class << self` here?

dkubb: No, I dislike `#send`. At best we could use `singleton_class.instance_eval { undef_method(:require) }`, but that's probably not better than `class << self` which is a built-in language construct.

mbj: Yeah, lets go with the buildin for now.

mbj:
As there are no unaddressed flags left I'll merge this one.

PR #533, opened 2016-02-14
Fix coverage for Mutant::Subject* namespace

mbj:


mbj:
@dkubb Ready for review.

lib/mutant/subject.rb
```diff
@@ -9,11 +9,10 @@ class Subject
     # @return [Enumerable<Mutation>]
     # @return [undefined]
     def mutations
-      mutations = [neutral_mutation]
-      Mutator.each(node) do |mutant|
-        mutations << Mutation::Evil.new(self, wrap_node(mutant))
+      [neutral_mutation] +
+      Mutator.each(node).map do |mutant|
```
dkubb: I don't think there's any strong reason to have this on it's own line. In fact. you can probably use `Array#concat` to save a tiny bit of memory allocation.

mbj: I was focusing on killing the mutation. I like your suggestion. OT: we need to make sure the neutral is present before the evils.

I'll have to refactor it later so that only evils are in `Subject#mutations` and the neutral gets injected explicitly, with special handling (mandatory fail fast).

mbj:
Merging this as all flags are addressed.

PR #534, opened 2016-02-14
Fix/setup error handling

mikeweaver:
I ran into a problem using mutant to test my rails application that uses SQLite as the DB. The mutant workers were fighting over the test DB and getting "SQLite3::ReadOnlyException: attempt to write a readonly database" exceptions. It took me a LONG time to figure out why the tests were failing because the SQLite exception was being eaten and a marshaling error, "marshal data too short" was being displayed instead. If you would like to reproduce the problem behavior, you may do so by running "bundle exec mutant -I ./lib/git --use rspec Git::GitBranch" on this repo: https://github.com/mikeweaver/git-conflict-detector

This pull request is a proposed fix for handling errors that occur during the setup inside mutant before running the actual test. It will cause the actual error that occurred to be shown to the user. If you agree with the fix, I will write some unit tests for it.

One concern I have: With this fix it won't be possible for mutant to differentiate between tests that failed because they were supposed to and tests that failed during mutant setup. Perhaps we should add a param to TestResult that could be use to differentiate these two cases?

### DIFF
--- a/lib/mutant/integration/rspec.rb
+++ b/lib/mutant/integration/rspec.rb
@@ -65,8 +65,15 @@ module Mutant
         Result::Test.new(
           output:   @output.read,
           passed:   passed,
-          runtime:  Time.now - start,
+          runtime:  calculate_run_time(start),
           tests:    tests
+        )
+      rescue => exception
+        Result::Test.new(
+            tests:    tests,
+            output:   exception.to_s,
+            runtime:  calculate_run_time(start),
+            passed:   false
         )
       end
 
@@ -142,6 +149,21 @@ module Mutant
       def filter_examples(&predicate)
         @world.filtered_examples.each_value do |examples|
           examples.keep_if(&predicate)
+        end
+      end
+
+      # Calculate the run time of a test
+      #
+      # @param [#Time] start_time
+      #
+      # @return [Time]
+      #
+      # @api private
+      def calculate_run_time(start_time)
+        if start_time
+          Time.now - start_time
+        else
+          Time.at(0)
         end
       end
 
--- a/lib/mutant/isolation.rb
+++ b/lib/mutant/isolation.rb
@@ -47,7 +47,11 @@ module Mutant
             end
 
             writer.close
-            Marshal.load(reader.read)
+            data = reader.read
+            if data.nil? || data.empty?
+              raise Error.new('Mutation returned no data!')
+            end
+            Marshal.load(data)
           ensure
             Process.waitpid(pid) if pid
           end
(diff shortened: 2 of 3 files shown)


mbj:
I had equivalents of such code before, and the problem is that rescuing *all* exceptions and mapping them into a non passed Test result hides bugs in mutant and rspec. 

If the layer that determines successful or failed tests itself permits arbitrary exceptions and maps them to one of these, something is wrong.

If an exception bubbles from your code through rspec and terminating the mutant integration with an exception: Its the fault of rspec. It should turn the exception into a false test result.

What we need is a 3rd state, `crashed`. This one would NOT be counted as killed, and reported explicitly.

Sadly I do not have the time to make this happen anytime soon. My opensource time is very limited.

As this PR in the current state introduces a regression: I'll close it for now. Feel free to discus.

> One concern I have: With this fix it won't be possible for mutant to differentiate between tests that failed because they were supposed to and tests that failed during mutant setup. Perhaps we should add a param to TestResult that could be use to differentiate these two cases?

This is already existent, mutant interprets the `#passed` depending on the type of the mutation, for self / environment verification reasons mutant internally emits neutral/noop mutations that propose is to test the signalling.

mbj:
> I ran into a problem using mutant to test my rails application that uses SQLite as the DB. The mutant workers were fighting over the test DB and getting "SQLite3::ReadOnlyException:

Did you saw noop errors? 

Also instead of mapping these kind of errors to not passed test result (and so basically get "false coverage assumption) better run with `-j1` or a more production relevant database with concurrency support?

Also transactions (you never commit) might help even on SQLite (but will on production DBs).

mikeweaver:
Thanks for the quick responses! I might have time to add the "crashed" state. Were you thinking of changing the "passed" boolean in TestResult to an enum with values "passed", "failed" and "crashed"?

Regarding running with -j1, yes, that is what I did to work around the concurrency problem with SQLLite. It works fine when there is only one worker.

I'd like to add some information to the readme about both of these issues. (How to diagnose a "marshal data too short" error and concurrency issues with SQLLite.) Are you ok with me updating the readme?

mbj:
@mikeweaver I think adding a readme section that elaborates about: Concurrency introduced problems, how they are signalled by noops etc.

I thought about this a bit longer, and I do not think a crashed state is the best idea. Better a sum type like:

```haskell
data Result = TestResult a | Exception e
```

I realize this is close to the `Either` type.

Because a crashed state would make most of the fields in `TestResult` unpopulated / invalid. Its better to wrap this whole thing.

> Thanks for the quick responses! I might have time to add the "crashed" state.

Its not a quick task to do it end-to-end. Since not only the IPC needs to change, but also reporting needs to be added. And the presence of an `Exception` IPC result needs IMO to terminate the runner loop.

This can be done in small iterations, fell free to pick one.

We have a Slack channel, if you want to discuss this in details feel free to send me your email address so I can invite you.

mbj:
Another thought: The explicit "IPC result" should not be added to the integration, it should be added to the isolation, and simply be re-raised in the master process when the isolation reads the result.

This way all possible integrations benefit.

PR #536, opened 2016-02-14
Fix coverage of Isolation namespace

mbj:


mbj:

I realize you might dislike the injection of so many globals, it looks like I'm "taking dependency injection too far". 

There are 2 reasons for this, one is mentioned in the code comments itself:

* Its easier to specify this way, and mutations to "non mocked side effectful operations" cannot occur, a problem that depending on timing mutant turned into a fork bomp when mutating this subject is gone away.

The other, a more long term goal:

* There is a *lot* to tune in mutants use of systemcalls, but before tuning we need to be able to gather fine grained statistics. I do want to consolidate all IO globals into a central place I can later replace with fine grained logging. Also I can add jitter etc. This is a step towards this goal. I realize that in initialization of `Mutant::Config::DEFAULT` many IO globals are referenced twice at different layers. But I'm fine with this for now as its an improvement, later passes will factor away these globals from the `Mutant::Config` and add them to the (currently not existend) `Mutant::World` or `Mutant::IO` object that will wrap all these and serve as the central injection point for metrics / jitter.

lib/mutant.rb
```diff
@@ -200,7 +200,14 @@ class Config
       fail_fast:         false,
       includes:          EMPTY_ARRAY,
       integration:       Integration::Null,
-      isolation:         Mutant::Isolation::Fork,
+      isolation:         Mutant::Isolation::Fork.new(
+        devnull: ->(&block) { File.open(File::NULL, File::WRONLY, &block) },
```
dkubb: I wonder if you could also open it with `File::APPEND`?

mbj: Likely, but as you cannot seek the `/dev/null` it would not make a difference.

But as: `File::APPEND` is the more "strict" version of `File::WRONLY` I think append is the better choice overall. If it works I'll use it. Good catch.

dkubb: What about doing something like:

```ruby
DEVNULL = Pathname.new(File::NULL).freeze

# ...

devnull: ->(&block) { DEVNULL.open(File::WRONLY | File::APPEND, &block) }
```

lib/mutant/isolation.rb
```diff
@@ -1,31 +1,35 @@
...
+    class None < self
       # Call block in isolation
       #
       # @return [Object]
       #
       # @raise [Error]
       #   if block terminates abnormal
-      def self.call(&block)
+      def call(&block)
         block.call
```
dkubb: Instead of assigning `&block` explicitly, what about just using `yield` within this method? It's not like you use `block` elsewhere here.

lib/mutant.rb
mbj: AFAIK you do not need `WRONLY` when you use `APPEND`. But I've to test this.

I personally prefer not to extract a constant in initialization code, because it stretches the intention about multiple lines.

lib/mutant/isolation.rb
mbj: True, I think this is the better pattern. I think I mentally prefer `&block` because it advertises "this method takes a block" something ruby should have enforced in the first place.

But I cannot / should not compensate this via redundant syntax.

spec/unit/mutant/isolation/fork_spec.rb
```diff
@@ -0,0 +1,272 @@
...
+      DEFAULTS = IceNine.deep_freeze(
+        arguments: [],
+        return:    nil,
+        yields:    nil,
+        exception: nil
+      )
+
+      private_constant(*constants(false))
+
+      def self.new(arguments)
```
dkubb: Are any of the arguments required? It looks like `:receiver` and `:selector` are required, so I would probably make them explicit. You can also then define the defaults within the method signature too, eg:

```ruby
def initialize(receiver:, selector:, arguments: [], return: nil, yields: nil, exception: nil)
  super
end
```

dkubb: Also, I realize that this method signature is gross, but I think it's better to make it explicit than to hide it behind another ruby construct.

spec/unit/mutant/isolation/fork_spec.rb
```diff
@@ -0,0 +1,272 @@
...
+
+      def call(observation)
+        Verifier.new(self, observation).call
+      end
+
+      class Verifier
+        include Concord.new(:expectation, :observation)
+
+        def call
+          %i[receiver selector arguments].each do |name|
```
dkubb: I would probably move this list of symbols into a constant.

dkubb: I would also move this check into an `assert_*` style method to match the design of the rest of the method.

spec/unit/mutant/isolation/fork_spec.rb
```diff
@@ -0,0 +1,272 @@
...
+        end
+
+      private
+
+        def error(message)
+          fail "#{message},\n observation:\n #{observation.inspect}\n expectation:\n #{expectation.inspect}"
+        end
+
+        def trigger_exception
+          exception = expectation.exception or return
```
dkubb: I think it would be clearer to use `fail exception if exception`.

spec/unit/mutant/isolation/fork_spec.rb
mbj: I like this form much more. Better than maintaining the `DEFAULTS` thing, I need to get used to kwards more, we are not constrained anymore by kwargs incompatible backwards compat.

spec/unit/mutant/isolation/fork_spec.rb
```diff
@@ -0,0 +1,272 @@
...
+
+    class MessageObservation
+      include Anima.new(:receiver, :selector, :arguments, :block)
+    end
+
+    class ExpectationVerifier
+      include Concord.new(:expectations)
+
+      def call(observation)
+        expectation = expectations.slice!(0) or fail "No expected message but observed #{observation}"
```
dkubb: Would `Array#shift` be more clear than `Array#slice!(0)`?

mbj: > Would Array#shift be more clear than Array#slice!(0)?

Sure, dunny why I forgot about `Array#shift` existance here.

spec/unit/mutant/isolation/fork_spec.rb
mbj: thats better

dkubb:
@mbj I finished a pass over this PR.

lib/mutant.rb
mbj: As we later do `$stderr.repoen(devnull)` the flags have to be match up with the state of `$stderr` so we have to open with `w`.

```
/home/mbj/devel/mutant/lib/mutant/isolation.rb:68:in `reopen': <STDERR> can't change access mode from "w" to "a" (ArgumentError)
```

mbj:
@dkubb Ready for the next pass

lib/mutant.rb
dkubb: Ahh good point. I don't think I'd ever seen that error. I do think it should be possible to change the access mode if you are restricting it further but I can see why they chose to do that.

lib/mutant/isolation.rb
```diff
@@ -1,31 +1,35 @@
...
       #
       # @return [Object]
       #
       # @raise [Error]
       #   if block terminates abnormal
-      def self.call(&block)
-        block.call
+      def call
+        yield
       rescue => exception
```
dkubb: Do you want to ignore exceptions that don't inherit from `StandardError`? If you want to catch all exceptions then you probably want `rescue Exception => exception`.

lib/mutant/isolation.rb
```diff
@@ -33,30 +37,41 @@ module Fork
...
+      # @param [IO] writer
+      #
+      # @return [undefined]
+      def child(reader, writer, &block)
+        reader.close
 
+        devnull.call do |null|
+          stderr.reopen(null)
+          stdout.reopen(null)
+          writer.write(marshal.dump(block.call))
```
dkubb: You could probably just do `yield` here.

lib/mutant/isolation.rb
mbj: Right now I've no counter example where I wanted to capture an exception that does NOT inherit from standard error. 

This iteration is only to get this into a verifiable state. I'll do a followup that does a much better job for propagating the exceptions captured in the child, I'll revisit your point here.

lib/mutant/isolation.rb
dkubb: Just to verify, the `write` command is atomic correct? I thought it had some internal buffering in place. Maybe `syswrite` might be a safer primitive to use?

mbj: True, will do. We should have a mutation from `block.call -> yield`

mbj: We do close the writer below, buffers will be flushed than.

spec/integration/mutant/isolation_spec.rb
```diff
@@ -0,0 +1,7 @@
+RSpec.describe Mutant::Isolation::Fork, mutant: false do
+  specify do
+    a = 1
+    Mutant::Config::DEFAULT.isolation.call { a = 2 }
```
dkubb: What about something a bit more intention revealing, eg:

```ruby
specify do
  a = 1
  expect { Mutant::Config::DEFAULT.isolation.call { a = 2 } }
    .to_not change(a).from(1)
end
```

spec/unit/mutant/isolation/fork_spec.rb
```diff
@@ -0,0 +1,264 @@
...
+      def assert_done
+        expectations.empty? or fail "unconsumed expectations:\n#{expectations.map(&:inspect).join}"
+      end
+
+      # rubocop:disable MethodLength
+      def self.verify(rspec_context, expectations)
+        verifier = new(expectations)
+
+        hooks = expectations
+          .group_by { |expectation| [expectation.receiver, expectation.selector] }
```
dkubb: What about:

```ruby
hooks = expectations.map(&:receiver).to_set
```

spec/integration/mutant/isolation_spec.rb
mbj: It feels to be abit redundant to re-specify the start state where it is specified in the same block. the `.from(1)` does not add anything to the strictness of the statement. Also the value of `a` does not matter at all as the only important semantic here is the unchanged value.

spec/unit/mutant/isolation/fork_spec.rb
dkubb: Actually, maybe I am misinterpreting what `.map(&:first)` is doing here.

mbj: > What about:
> 
> hooks = expectations.map(&:receiver).to_set

This is not equivalent as I need each unique pair of `receiver, selector`.

spec/unit/mutant/isolation/fork_spec.rb
```diff
@@ -0,0 +1,264 @@
...
+          .map(&:first)
```
dkubb: Would you get the same result by replacing this with `.keys` ?

spec/unit/mutant/isolation/fork_spec.rb
mbj: > Actually, maybe I am misinterpreting what .map(&:first) is doing here.

yeah, `Enumerable#group_by` produces not only the unique "projection" but also the values that fall under the projected group identity. I throw the latter away as I do not need them.

spec/unit/mutant/isolation/null_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+      expect(object.call { :foo }).to be(:foo)
+    end
+
+    it 'wraps *all* exceptions' do
+      expect { object.call { fail 'foo' } }.to raise_error(
+        Mutant::Isolation::Error,
+        'foo'
+      )
+    end
+
```
dkubb: This extra blank line can be removed.

spec/unit/mutant/isolation/fork_spec.rb
mbj: yes, I think this is an artifact from earlier attempts.

lib/mutant/isolation.rb
dkubb: I guess I was just worried about interleaving of writes. I believe there may be some known issues with around this.

mbj: How could the write be interleaved? We only have one writer and one reader. OT: I also think we should bypass the buffers here, because "why rebuffering" something that is already in a "string buffer ready to be handed over to the kernel".

But from terms for "write atomitcy in this case" `writer.write(buffer); writer.close` has the same smeantics than `writer.syswrite(buffer); writer.close`.

lib/mutant/isolation.rb
```diff
@@ -33,30 +37,41 @@ module Fork
...
+      # @param [IO] writer
+      #
+      # @return [undefined]
+      def child(reader, writer)
+        reader.close
 
+        devnull.call do |null|
+          stderr.reopen(null)
+          stdout.reopen(null)
+          writer.write(marshal.dump(yield))
```
dkubb: FYI `Marshal#dump` takes an optional second `IO` object which it can dump to directly.

mbj: Interesting. But sadly there is no documentation for:

* How does it handle partial writes
* Does it use rubies buffering or not

So I'll change this one to an explicit syswrite instead. Also I suspect they write small amounts of data per object traversed (why would you just ofter such an interface if not) for our use case a single write will be more effective.

lib/mutant/isolation.rb
dkubb: > How could the write be interleaved? We only have one writer and one reader.

Ahh I didn't realize this was the case. I was operating under the assumption of there being multiple writers.

> But from terms for "write atomitcy in this case" writer.write(buffer); writer.close has the same smeantics than writer.syswrite(buffer); writer.close.

I guess if there's only one writer, that is the case. However, there aren't any strong guarantees like this. Ruby does all sorts of internal buffering when writing, as does the file system. Here's one example of this: https://github.com/steveklabnik/mono_logger/issues/1 (although I doubt this specifically applies to us since it's not talking about pipes, but it's something to keep in mind)

lib/mutant/isolation.rb
dkubb: > So I'll change this one to an explicit syswrite instead.

I think this is the safest approach.

lib/mutant/isolation.rb
```diff
@@ -33,30 +37,41 @@ module Fork
...
-                writer.write(Marshal.dump(block.call))
-                writer.close
-              end
+            pid = process.fork do
+              child(reader, writer, &block)
             end
 
             writer.close
-            Marshal.load(reader.read)
+            marshal.load(reader.read)
```
dkubb: I think you can pass in an `IO` object into `Mashall#load`, so you could write this as: `marshal.load(reader)`.

mbj: I think this is better, yes. Especially as the stream is "EOF terminated".

lib/mutant/isolation.rb
```diff
@@ -33,30 +37,41 @@ module Fork
       #   if block terminates abnormal
       #
       # rubocop:disable MethodLength
-      def self.call(&block)
-        IO.pipe(binmode: true) do |reader, writer|
-          writer.binmode
+      def call(&block)
+        io.pipe(binmode: true) do |reader, writer|
           begin
```
dkubb: What if you pull this inner loop into a method named `parent` and gave it the same signature as `#child`? eg:

```ruby
def call(&block)
  io.pipe(binmode: true) do |reader, writer|
    parent(reader, writer, &block)
  end
rescue => exception
  raise Error, exception
end

private

def parent(reader, writer, &block)
  pid = process.fork do
    child(reader, writer, &block)
  end
  writer.close
  marshal.load(reader)
ensure
  Process.waitpid(pid) if pid
end

def child(reader, writer, &block)
  reader.close
  writer.binmode

  devnull.call do |null|
    stderr.reopen(null)
    stdout.reopen(null)
    writer.syswrite(marshal.dump(yield))
  end
  writer.close  # XXX: can this be done outside the devnull.call block?
end
```

spec/unit/mutant/isolation/fork_spec.rb
```diff
@@ -0,0 +1,260 @@
...
+      )
+    end
+
+    subject { object.call(&isolated_block) }
+
+    class MessageExpectation
+      include Anima.new(:receiver, :selector, :arguments, :return, :yields, :exception)
+
+      # rubocop:disable ParameterLists
+      def self.new(receiver:, selector:, arguments: [], return: nil, yields: nil, exception: nil)
```
dkubb: Can this be done inside `#initialize`?

lib/mutant/isolation.rb
mbj: I like this one, but actually why not go a step more and wrap the "redundant passes of `reader, writer`" into a dedicated object?

spec/unit/mutant/isolation/fork_spec.rb
mbj: Yeah, but my policy is to do it on the "scope that needs the least power". `.new` does not see a specific instance.

mbj: Actually it sees "all instances", let me repharse as: It cannot access the API that writes explicitly to the instance state.

lib/mutant/isolation.rb
dkubb: I dunno if I would go that far, since I can't see such an object being anything more than a container.

mbj: Passing a "data clump" is always a sign of a missing abstraction. Where in this case it could be a sign of "ruby not providing tuples".

For now I'll take the explicit approach.

mbj: We effectively pass 3 arguments (reader, writer, block) to me this feels like redundancy. I'll try out a dedicated object to reduce this.

dkubb: > We effectively pass 3 arguments (reader, writer, block)

This is a good point. I was only thinking of it as 2.

mbj: The object looks horrible, mostly because we have to pass down the IO globals we captured as state. Once I finished wraping these up into `Mutant::IO` we can revisit this for now we'll go with the data clump.

lib/mutant/isolation.rb
```diff
@@ -1,62 +1,83 @@
...
+      # @return [undefined]
+      def child(reader, writer)
+        reader.close
+        writer.binmode
+
+        devnull.call do |null|
+          stderr.reopen(null)
+          stdout.reopen(null)
+          writer.syswrite(marshal.dump(yield))
+          writer.close
```
dkubb: Can this `#close` be done outside of the block, or does it have to be done here?

lib/mutant/isolation.rb
```diff
@@ -1,62 +1,84 @@
...
+      #
+      # @return [undefined]
+      def child(reader, writer)
+        reader.close
+        writer.binmode
+
+        devnull.call do |null|
+          stderr.reopen(null)
+          stdout.reopen(null)
+          writer.syswrite(marshal.dump(yield))
```
dkubb: I would recommend changing this so that it makes it more clear the specific action that is causing `stderr` and `stdout` to be be redirected:

```ruby
def child(reader, writer)
  output = devnull.call do |null|
    stderr.reopen(null)
    stdout.reopen(null)
    yield
  end

  reader.close
  writer.binmode.syswrite(marshal.dump(output))
  writer.close
end
```

dkubb:
@mbj I am fine with the PR now. Will merge when green.

lib/mutant/isolation/fork.rb
```diff
@@ -0,0 +1,67 @@
...
+      def child(reader, writer)
+        result = devnull.call do |null|
+          stderr.reopen(null)
+          stdout.reopen(null)
+          yield
+        end
+
+        reader.close
+        writer.binmode
+        writer.syswrite(marshal.dump(result))
```
dkubb: You *could* do `writer.binmode.syswrite(...)` if you wanted. I noticed that rubocop flagged this method due to too many expressions. I don't know what the limit is, but maybe this will get you under the threshold.

mbj:
@dkubb Ready for the next pass, I expect green and also changed a bit since your last acked state. Mostly to trade the lvar assignement for a block pass.

lib/mutant/isolation/fork.rb
```diff
@@ -0,0 +1,72 @@
...
+      #
+      # @return [undefined]
+      def child(reader, writer, &block)
+        reader.close
+        writer.binmode
+        writer.syswrite(marshal.dump(result(&block)))
+        writer.close
+      end
+
+      # The block result computed under silencing1
```
dkubb: There's a trailing `1` in the docs here.

mbj:
@dkubb Ready for the next pass. I fixed the "reaction" part of the message expectations to be explicit now, reduces lots of `nils` and "sniffing for nils" etc. Also it now enables the call side to have direct control about the return value.

spec/unit/mutant/isolation/fork_spec.rb
mbj: Another way to explain this is: `group_by + keys` is like `map + uniq`.

dkubb: I think `#map`, `#uniq` (or `#to_set`) might actually end up being more efficient.

mbj: > I think #map, #uniq (or #to_set) might actually end up being more efficient.

I primarily intend to not produce garbadge to throw it away. But as I throw the values of the hash away I should not think about this.

Too bad there is no `uniq_with` or something.

dkubb: > Too bad there is no uniq_with or something.

What about:

```ruby
hooks = expectations.reduce(Set.new) do |expectation, set|
  set << [expectation.receiver, expectation.selector]
end
```

> I primarily intend to not produce garbadge to throw it away.

This is probably always going to be the case to some degree. You have to iterate over the list once, at which time you may encounter expectations with a similar receiver and selector. At least in the above case you only store the first encounter, and the data structure it builds is not further reduced -- so there's less GC pressure.

lib/mutant/isolation/fork.rb
```diff
@@ -0,0 +1,72 @@
...
+      # Call block in isolation
+      #
+      # @return [Object]
+      #   returns block execution result
+      #
+      # @raise [Error]
+      #   if block terminates abnormal
+      def call(&block)
+        io.pipe(binmode: true) do |reader, writer|
+          parent(reader, writer, &block)
```
dkubb: I suppose we *could* define something like `*pair`  and yield it, and then pass that into `#parent` (?). There isn't a reason to name the variables at this point.

mbj: This would not enforce arity.

dkubb: > This would not enforce arity.

The `#parent` method enforces arity though.

This is what I was proposing:

```ruby
def call(&block)
  io.pipe(binmode: true) do |*pair|
    parent(*pair, &block)
  end
rescue => exception
  raise Error, exception
end
```

mbj: I meant more that the "yield" interface would not have arity enforced.

mbj: ^^ I mean IO.pipe could do any number of yields and we'd not notice until `parent` call.

But actually this is "default" in ruby. Fuck it.

spec/unit/mutant/isolation/fork_spec.rb
```diff
@@ -0,0 +1,325 @@
...
+        verifier.assert_done
+      end
+    end
+
+    let(:prefork_expectations) do
+      [
+        {
+          receiver:      io,
+          selector:      :pipe,
+          arguments:     [binmode: true],
```
dkubb: The alignment looks a bit "off" here.

mbj:
I'll merge this one as last flags where addressed.

PR #537, opened 2016-02-14
Extend and reorganize the readme

mikeweaver:


README.md
```diff
@@ -219,35 +203,67 @@ Mutation output is grouped by selection groups. Each group contains three sectio
     end
    -----------------------
    ```
+   
+Concurrency
+-----------
 
-Presentations
--------------
+By default, mutant will test mutations in parallel by running up to one process for each core on your system. You can control the number of processes created using the --jobs argument.
```
mbj: Can we quote the jobs flag like `--jobs` this will result in better readability.

README.md
```diff
@@ -219,35 +203,67 @@ Mutation output is grouped by selection groups. Each group contains three sectio
...
-There are some presentations about mutant in the wild:
+If the code under test relies on a database, you may experience problems when running mutant because of conflicting data in the database. For example, if you have a test like this:
+```
+m = MyModel.create!(...)
+expect(MyModel.first.name).to eq(m.name)
+```
+It might fail if some other test wrote a record to the MyModel table at the same time as this test was executed. (It would find the MyModel record created by the other test.) Most of these issues can be fixed by writing more specific tests. Here is a concurrent safe version of the same test:
+```
+m = MyModel.create!(...)
+expect(MyModel.find_by_id(m.id).name).to eq(m.name)
```
mbj: Can we use the `eql` matcher here, as this mirrors the preference for `#eql` ;)

README.md
```diff
@@ -219,35 +203,67 @@ Mutation output is grouped by selection groups. Each group contains three sectio
...
+```
+You may also try wrapping your test runs in transactions, rather than fixing your tests.
```
mbj: I'd remove the

> rather than fixing your tests

As I think you should do "both", fixing your tests to not be sequence dependent, and isolating them via system transactions.

README.md
```diff
@@ -219,35 +203,67 @@ Mutation output is grouped by selection groups. Each group contains three sectio
...
+Only Mutating Changed Code
+--------------------------
 
-[blockscore]: https://blog.blockscore.com/how-to-write-better-code-using-mutation-testing/
-[sitepoint]: http://www.sitepoint.com/mutation-testing-mutant/
-[arkency1]: http://blog.arkency.com/2015/06/how-good-are-your-ruby-tests-testing-your-tests-with-mutant/
-[arkency2]: http://blog.arkency.com/2015/05/mutation-testing-and-continuous-integration/
-[arkency3]: http://blog.arkency.com/2015/04/why-i-want-to-introduce-mutation-testing-to-the-rails-event-store-gem/  
-[solnic]: http://solnic.eu/2013/01/23/mutation-testing-with-mutant.html
+Running mutant for the first time on a existing codebase can be a rather disheartining experience due to the large number of alive mutations found! Mutant has a setting that can help. Using the --since argument, mutant will only mutate code that has been modified. This allows you to introduce mutant into an existing code base without drowning in errors. Example usage that will mutate all code changed between master and the current branch:
```
mbj: Can we use `--since` here  for readability?

mbj:
@mikeweaver I like this change and I left minor flags that should be easy to address. As we only do fast forward merges here, do you mind rebasing your branch on top of master.

mikeweaver:
Comments addressed and rebased on top of master. Should be ready to merge.

mbj:
Closing this one in favor of #539, containing the same change, but rebased. Authorship intact.

PR #538, opened 2016-02-15
Bubble exceptions in isolation up to parent process

mikeweaver:
If an exception is raised during the setup or tear down for the mutant it will cause the block to return no data. This causes Marshal.load to fail with a vague error message: "marshal data too short".

Instead, the block should catch the exception and return a stringified version. Then the isolation can raise an exception with the string.

### DIFF
--- a/lib/mutant/isolation.rb
+++ b/lib/mutant/isolation.rb
@@ -47,11 +47,18 @@ module Mutant
             end
 
             writer.close
-            Marshal.load(reader.read)
+            result = Marshal.load(reader.read)
+            if result.is_a?(Mutant::Result::Test)
+              result
+            else
+              raise Error, result.to_s
+            end
           ensure
             Process.waitpid(pid) if pid
           end
         end
+      rescue Error => exception
+        raise exception
       rescue => exception
         raise Error, exception
       end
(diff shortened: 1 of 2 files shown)


mikeweaver:
@mbj Let me know if this is what you were thinking. If so, then I will fix the tests.

mbj:
@mikeweaver I want to go a different implementation path, and as I told you in channel (which you might missed, sorry for that) I'm currently refactoring the `Mutant::Isolation::Fork` class to be better mutation coverable, before we add a behavior I want this to be done.

So I'm sorry to close one of your PRs again, as I want a different sequence of changes.

lib/mutant/isolation.rb
```diff
@@ -47,11 +47,18 @@ def self.call(&block)
             end
 
             writer.close
-            Marshal.load(reader.read)
+            result = Marshal.load(reader.read)
+            if result.is_a?(Mutant::Result::Test)
```
mbj: I think that this is a very weak test. Better have a dedicated "subprocess result object" that wraps A) a valid test result, B) an exception (with backtrace!)

See Haskells `Either` type or sum types in general.

mbj: Also we have still to account for the fact that the subproject might NOT return such an object, as it crashed while producing it. We are on ruby a low-guarantee language.

PR #540, opened 2016-02-18
update README limitations section

connorjacobsen:
@mbj Please make sure the updates I made are correct and clear. Also please let me know if you think a more explicit example / rewording would be better.

Small tweaks:
- Changed indentation level on code blocks in limitations section.
  - Rendered fine on GitHub, but looked funky on other Markdown renderers (Atom), now fine on both.
- Atom trimmed some trailing whitespace.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,48 +62,49 @@ Limitations
 
 Mutant cannot emit mutations for...
 
-* methods defined within a closure.  For example, methods defined using `module_eval`, `class_eval`,
-  `define_method`, or `define_singleton_method`:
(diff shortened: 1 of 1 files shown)


mbj:
@connorjacobsen The readme is not substantially de confusing the reader on which scenarios we can run into. 

The message: We cannot mutate what is *generated*, but we can mutate the *generator* does not make it. With, or without your PR.

I'll close this PR for now. If you could think about a better way to pharse this (maybe even just cite the above?) lets reopen it.

PR #543, opened 2016-03-06
Fix coverage for Mutant::Meta namespace

mbj:
* Remove unparser invariant checking, does not yield enough value
  anymore as unparser is more stable these days


mbj:
@dkubb Assuming all green, this is ready for review.

lib/mutant/meta.rb
```diff
@@ -15,15 +16,20 @@ class Example
       # @return [undefined]
       def self.add(&block)
         file = caller.first.split(':in', 2).first
-        ALL << DSL.run(file, block)
+        ALL << DSL.call(file, block)
       end
 
-      Pathname.glob(Pathname.new(__FILE__).parent.parent.parent.join('meta', '**/*.rb'))
+      Pathname.glob(Pathname.new(__dir__).parent.parent.join('meta', '*.rb'))
```
backus: If you're adjusting this line could you use the shorthand `Pathname(__dir__)`?

lib/mutant/meta/example/dsl.rb
```diff
@@ -9,18 +8,20 @@ class DSL
         # Run DSL on block
         #
         # @return [Example]
-        def self.run(file, block)
+        def self.call(file, block)
```
backus: I'm curious, why did you rename this to `call`?

lib/mutant/meta/example/verification.rb
```diff
@@ -0,0 +1,86 @@
...
+    class Example
+      # Example verification
+      class Verification
+        include Adamantium::Flat, Concord.new(:example, :mutations)
+
+        # Test if mutation was verified successfully
+        #
+        # @return [Boolean]
+        def success?
+          missing.empty? && unexpected.empty? && no_diffs.empty?
```
backus: I noticed you seem to have completely dropped the `unparser` method and `unparser.empty?`. Was that performing a check that is no longer necessary?

spec/unit/mutant/meta/example/verification_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+          no_diff: []
+        REPORT
+      end
+    end
+
+    context 'when there is extra generated node' do
+      let(:generated_nodes) { [s(:false), s(:nil)] }
+
+      specify do
+        should eql(strip_indent(<<-'REPORT'))
```
backus: Do you ever run your specs with a ruby version lower than 2.3? If not then I imagine this could use `<<~`

lib/mutant/meta.rb
mbj: I do not like using methods from Kernel, that infect every object without need.

lib/mutant/meta/example/dsl.rb
mbj: `.call` is the better convention. As Rubies own blocks / procs / lambdas follow these.

lib/mutant/meta/example/verification.rb
mbj: Yes, this is outlined in the commit message.

spec/unit/mutant/meta/example/verification_spec.rb
mbj: Yes 2.2 is still supported for development. I intend specs should pass under all versions mutant supports, as mutant supports 2.2 I cannot use 2.3 specific syntax.

lib/mutant/meta/example/verification.rb
```diff
@@ -0,0 +1,86 @@
...
+        end
+
+        # Error report
+        #
+        # @return [String]
+        def error_report
+          fail 'no error report on successful validation' if success?
+
+          YAML.dump(
+            'file' =>            example.file,
```
dkubb: I would suggest aligning the `=>` operators.

lib/mutant/meta/example/verification.rb
```diff
@@ -0,0 +1,86 @@
...
+        # @return [Boolean]
+        def success?
+          missing.empty? && unexpected.empty? && no_diffs.empty?
+        end
+
+        # Error report
+        #
+        # @return [String]
+        def error_report
+          fail 'no error report on successful validation' if success?
```
dkubb: You're ok with this failing with a `StandardError` exception? Not something more specific?

lib/mutant/meta/example/verification.rb
```diff
@@ -0,0 +1,86 @@
...
+
+        # Mutation report
+        #
+        # @param [Array<Parser::AST::Node>] nodes
+        #
+        # @return [Array<Hash>]
+        def format_mutations(nodes)
+          nodes.map do |node|
+            {
+              'node' =>   node.inspect,
```
dkubb: Instance of https://github.com/mbj/mutant/pull/543/files#r55152676

lib/mutant/meta/example/verification.rb
```diff
@@ -0,0 +1,86 @@
...
+          end
+        end
+
+        # No diff mutation report
+        #
+        # @return [Array, nil]
+        def no_diff_report
+          no_diffs.map do |mutation|
+            {
+              'node' =>   mutation.node.inspect,
```
dkubb: Instance of https://github.com/mbj/mutant/pull/543/files#r55152676

lib/mutant/meta/example/dsl.rb
```diff
@@ -9,18 +8,20 @@ class DSL
...
           instance = new(file)
           instance.instance_eval(&block)
```
dkubb: I think `#instance_eval` returns `self`. You could probably write this as:

```ruby
def self.call(file, block)
  new(file).instance_eval(&block).example
end
```

lib/mutant/meta/example/dsl.rb
dkubb: Why not use `&block` here to make it explicit that it's a Proc object.

dkubb:
@mbj I did a quick first pass over this PR.

lib/mutant/meta/example/verification.rb
mbj: Its a `RuntimeError`, and I do not want anyone to consider rescuing these errors, so I do not want to give any specific hint to do so via a dedicated exception class. This is only used from specs, and they *should* bubble to a failed example.

lib/mutant/meta/example/dsl.rb
mbj: It does not, it returns the last expression evaluated in block.

lib/mutant/meta/example/dsl.rb
mbj: Because the call side has to capture the block into an object already, as it needs to be evaluated somewhere else. It does not make sense for me to pass it as a block to capture it to pass it as a block again.

Also a real argument counts against the arity checks, where a block does not.

mbj:

lib/mutant/meta/example/dsl.rb
dkubb: I think the docs for `BasicObject#instance_eval` are incorrect then: http://ruby-doc.org/core-2.2.0/BasicObject.html#method-i-instance_eval

PR #545, opened 2016-03-14
Fix coverage of Mutant::Mutator namespace

mbj:


mbj:
@dkubb Ready for review, I assume it'll be all green.

lib/mutant/mutator/node/literal.rb
```diff
@@ -4,6 +4,17 @@ class Node
...
       class Literal < self
         include AbstractType
+
+      private
+
+        # Emit values
+        #
+        # @return [undefined]
+        def emit_values
+          values.each do |value|
```
dkubb: What about: `values.each(&method(:emit_type))` ?

lib/mutant/mutator/node/rescue.rb
```diff
@@ -28,8 +28,7 @@ def dispatch
         # @return [undefined]
         def mutate_rescue_bodies
           children_indices(RESCUE_INDICES).each do |index|
-            rescue_body = children.at(index)
-            next unless rescue_body
+            rescue_body = children.fetch(index)
```
dkubb: Is the `rescue_body` synthetic variable even needed?

dkubb:
@mbj I finished a first pass over the code. Also it looks like the build failed.

mbj:
> Also it looks like the build failed.

Circle runs into spec timeouts, I've to bump them.

lib/mutant/mutator/node/rescue.rb
mbj: Refactoring leftover, good catch.

lib/mutant/mutator/node/literal.rb
mbj: good catch.

mbj:
@dkubb Ready for the next pass. I did not push the timeouts for circle yet, in the hope this build gets a less busy base node.

lib/mutant/mutator/node.rb
```diff
@@ -56,10 +56,10 @@ def children
       # @return [undefined]
       #
       # rubocop:disable RedundantBlockCall - its not redundant here
-      def mutate_child(index, mutator = Mutator, &block)
+      def mutate_child(index, &block)
         block ||= TAUTOLOGY
-        child = children.at(index)
-        mutator.each(child, self) do |mutation|
+        child = children.fetch(index)
```
dkubb: Any reason for the `child` synthetic variable?

mbj: Same refactoring left over as you spotted elsewhere. I was focusing on a different area when I passed this code.

I do think this can easily be a mutation. If the synthetic lvar can be inlined with its only use-side, the mutation is alive.

If its important from side effectful sequencing, mutation will be killed.

lib/mutant/mutator/node/arguments.rb
```diff
@@ -59,7 +55,7 @@ def emit_mlhs_expansion
           mlhs_childs_with_index.each do |child, index|
             dup_children = children.dup
             dup_children.delete_at(index)
-            dup_children.insert(index, *child.children)
+            dup_children.insert(index, *child)
```
dkubb: Why the splat? Does the `child` respond to `#to_a` or `#to_ary` (can't remember which expands in the presence of the splat arg)? Or can child be `nil`? (doubtful, but figured I should ask)

lib/mutant/mutator/node/begin.rb
```diff
@@ -13,23 +13,11 @@ class Begin < self
         #
         # @return [undefined]
         def dispatch
-          Util::Array.each(children, self, &method(:emit_child_subset))
           children.each_with_index do |child, index|
             mutate_child(index)
             emit(child) unless children.eql?([child])
```
dkubb: What about: `emit(child) unless children.one?` ?

lib/mutant/mutator/node/arguments.rb
mbj: It does. It implements `#to_ary`. delegating to `#children`. I do not like it but given its supported API mutant forces the least powerful primitive in the current environment.

lib/mutant/mutator/registry.rb
```diff
@@ -21,22 +22,30 @@ def initialize
...
       # @return [Class]
       #
       # @raise [ArgumentError]
       #   raises argument error when mutator class cannot be found
-      def lookup(node)
-        type = node.type
-        @registry.fetch(type) do
+      def lookup(type)
+        contents.fetch(type) do
           fail RegistryError, "No mutator to handle: #{type.inspect}"
```
dkubb: Why use `type.inspect` in the error here, but not above? I would suggest choosing one approach in error messages.

lib/mutant/mutator/node/begin.rb
mbj: There is a reason I do not recall right now. If the specs pass I take your idea as a mutation.

meta/regex.rb
```diff
@@ -17,3 +17,17 @@
...
+  source 'true if /foo/'
+
+  singleton_mutations
+  mutation 'false if /foo/'
+  mutation 'true if //'
+  mutation 'nil if /foo/'
+  mutation 'true if true'
+  mutation 'true if false'
+  mutation 'true if nil'
+  mutation 'true if /a\A/'
```
dkubb: This is an odd regexp to generate. I don't think it can ever match anything. Was that intentional?

lib/mutant/mutator/node/begin.rb
mbj: @dkubb The reason is not existend anymore as unparser was fixed for that weird edge case using `#one?`.

meta/regex.rb
mbj: Thats the whole idea, have a regexp executed that NEVER matches something ;)

dkubb: Ahh good. I figured that was the case but I wanted to be sure.

mbj: If you know a more canonical one for "does not match anything" let me know.

tjchambers: I would have used `/\z\A/`. The end before the beginning. Not sure whether that works, but a comment would be helpful for the next person.

mbj:
```
irb(main):001:0> /\z\A/.match('')
=> #<MatchData "">
```

mbj: Lets go with `/nomatch\A/` as it kinda has "self documenting" properties.

lib/mutant/mutator/node/begin.rb
```diff
@@ -13,23 +13,11 @@ class Begin < self
         #
         # @return [undefined]
         def dispatch
-          Util::Array.each(children, self, &method(:emit_child_subset))
           children.each_with_index do |child, index|
             mutate_child(index)
-            emit(child) unless children.eql?([child])
+            emit(child) unless children.one?
           end
         end
```
dkubb: Tiny point, but do you want there to be a blank line before the class/module `end`?

lib/mutant/mutator/node/literal.rb
```diff
@@ -4,6 +4,15 @@ class Node
...
         include AbstractType
+
+      private
+
+        # Emit values
+        #
+        # @return [undefined]
+        def emit_values
+          values.each(&method(:emit_type))
+        end
```
dkubb: Instance of https://github.com/mbj/mutant/pull/545/files#r56919414

lib/mutant/mutator/node/literal/regex.rb
```diff
@@ -8,7 +8,7 @@ class Regex < self
           handle(:regexp)
 
           # No input can ever be matched with this
-          NULL_REGEXP_SOURCE = 'a\A'.freeze
+          NULL_REGEXP_SOURCE = 'nomatch\A'.freeze
```
dkubb: I like this better, makes it more clear what your intention is.

lib/mutant/mutator/util/array.rb
```diff
@@ -42,18 +42,6 @@ def dispatch
           end
 
         end # Element
```
dkubb: Instance of https://github.com/mbj/mutant/pull/545/files#r56919414

lib/mutant/mutator/node/begin.rb
mbj: I do not see this as required. But we should sync up on that one once. This code base is inconsistend in itself on that empty line. Lets avoid a cycle here for a global fix + enforcement to reduce that entropy in the future.

dkubb:
@mbj this looks ready to merge, except the build is still failing.

mbj:
@dkubb Ready for the next pass, assuming all green.

lib/mutant/mutator/node/begin.rb
```diff
@@ -13,23 +13,10 @@ class Begin < self
         #
         # @return [undefined]
         def dispatch
-          Util::Array.each(children, self, &method(:emit_child_subset))
-          children.each_with_index do |child, index|
-            mutate_child(index)
-            emit(child) unless children.eql?([child])
+          mutate_single_child do |child|
```
mbj: @dkubb Proactive comment: `mutate_single_child` yields 2 args, so `&method(:emit)` is not possible here.

mbj:
> @mbj this looks ready to merge, except the build is still failing.

I'm merging this as the review is green.

PR #547, opened 2016-03-19
Add support for explicit type in meta examples

mbj:
* This is needed to relate examples to node types for nodes that cannot
  stand alone. Example are :where nodes that do not have syntax outside
  a :case


mbj:
@dkubb Ready for review. The upcoming change for `Mutant::Mutator*` namespace coverage needs this change. But I figured to extract it into its own branch.

backus:
So what is the intended use case here? Is your intention to update the meta specs to read like this?

```ruby
Mutant::Meta::Example.add(:true) do
  source 'true'

  mutation 'nil'
  mutation 'false'
end
```

and then change the behavior such that only mutations to the `s(:true)` node in this case are checked?

mbj:
> So what is the intended use case here? Is your intention to update the meta specs to read like this?

Its verbosely described in the commit message.

backus:
> Its verbosely described in the commit message.

Ah alright. I read that actually but didn't entirely follow. So if I want to specify a mutation for a `s(:where)` node I do `Mutant::Meta::Example.add(:where)` and then what? I'm looking at the specs but I'm having a hard time confidently discerning the change in behavior.

spec/unit/mutant/meta/example/verification_spec.rb
```diff
@@ -3,9 +3,10 @@
 
   let(:example) do
     Mutant::Meta::Example.new(
-      'foo.rb',
-      s(:true),
-      expected_nodes
+      file:      'foo.rb',
+      node:      s(:true),
+      node_type: true,
```
backus: Should this be `true` or `:true`?

mbj:
> I'm looking at the specs but I'm having a hard time confidently discerning the change in behavior.

I do not use the new API yet, its an extraction from a followup change.

spec/unit/mutant/meta/example/verification_spec.rb
mbj: `:true` is correct, good catch.

backus:
> I do not use the new API yet, its an extraction from a followup change.

Alright I had a hunch but I wanted to make sure since I don't know the DSL codebase well

backus:
How would you feel about removing `--ignore-subject Mutant::Meta*` from the Rakefile? It looks like it wont fail the build and it seems to make sense for a PR like.

mbj:
> How would you feel about removing --ignore-subject Mutant::Meta* from the Rakefile? It looks like it wont fail the build and it seems to make sense for a PR like.

This should have been done alrady with the previous PR that covered the `Mutant::Meta*`, #543.

lib/mutant/meta.rb
```diff
@@ -14,9 +14,9 @@ class Example
       # Add example
       #
       # @return [undefined]
-      def self.add(&block)
+      def self.add(type = nil, &block)
```
dkubb: Should this be `type = Undefined` to distinguish between the case where an explicit `nil` was passed in vs. the case where no argument was passed in?

lib/mutant/meta/example/dsl.rb
```diff
@@ -19,9 +19,10 @@ def self.call(file, block)
         # Initialize object
         #
         # @return [undefined]
-        def initialize(file)
-          @file     = file
-          @source   = nil
+        def initialize(file, type)
+          @file, @type = file, type
```
dkubb: For consistency I would probably just have each assignment on their own line, rather than using two different ivar assignment approaches in one method.

lib/mutant/meta/example.rb
```diff
@@ -1,7 +1,7 @@
 module Mutant
   module Meta
     class Example
-      include Adamantium, Concord::Public.new(:file, :node, :mutations)
+      include Adamantium, Anima.new(:file, :node, :node_type, :expected)
```
dkubb: Possibly OT, but for Anima assert that all the attributes are required? I was just wondering what would happen if extra keys were provided. Would it be possible (or useful) for Anima to create an `#initialize` method that uses the new hash-based args, eg:

```ruby
def initialize(file:, node:, node_type:, expected:)
  @file      = file
  @node      = node
  @node_type = node_type
  @expected  = expected
end
```

lib/mutant/meta.rb
mbj: These do not intend to have a semantic difference. But actually: Why not change the DSL so the type has to be passed always, this means touching all meta examples, but thats the better way to do. I'll update the PR.

lib/mutant/meta/example.rb
mbj: > Possibly OT, but for Anima assert that all the attributes are required? I

Yes.

> I was just wondering what would happen if extra keys were provided

Exception from `Anima` namespace.

> #initialize method that uses the new hash-based args, eg:

The semantics already exist. From the call side this is indistinguishable.

mbj:
@dkubb ready for the next pass.

mbj:
@dkubb Ready for the next pass, should be green this build.

meta/and.rb
```diff
@@ -1,4 +1,4 @@
-Mutant::Meta::Example.add do
+Mutant::Meta::Example.add :and do
```
backus: I'm surprised that you opted to write these all without parenthesis as opposed to `Mutant::Meta::Example.add(:and) do`

mbj: I do think that only for DSLs explicit parens are not required.

backus: Yeah I agree with that style guideline. In this case I still think it would read better.

mbj: I'd consider to go "explicit parens always" in future. Just to kill a rule.

backus: I would support that. Could also restrict the rule to just ruby DSL like `attr_reader`

mbj: Lets go full hardcore and parens, always. Easier to think. Lets discuss this in channel and think about shipping a base rubocop config in devtools?

lib/mutant/meta/example.rb
dkubb: > The semantics already exist. From the call side this is indistinguishable.

Yeah, I was just saying if Anima is doing it in-code, maybe it would be worth considering pushing it down to have ruby itself handle it. I'm not sure if there's a way to do this without "string class_eval" though -- it's kind of sad there's no nice interface to assemble some kinds of ruby methods without resorting to string class_eval.

mbj: > Yeah, I was just saying if Anima is doing it in-code, maybe it would be worth considering pushing it down to have ruby itself handle it. I'm not sure if there's a way to do this without "string class_eval" though -- it's kind of sad there's no nice interface to assemble some kinds of ruby methods without resorting to string class_eval.

The `kwargs` are just syntax for ruby as in anima. There is AFAIK no ruby interface for checking hash contents, or in-lining AST nodes as behavior.

dkubb:
@mbj I am fine to merge this assuming you and @backus don't have anything outstanding you are discussing.

PR #548, opened 2016-03-19
Update project rubocop preferences

backus:
This PR contains many opinionated rubocop changes which I think represent this project's style preferences.

### DIFF
--- a/lib/mutant/color.rb
+++ b/lib/mutant/color.rb
@@ -24,7 +24,7 @@ module Mutant
         text
       end
 
-    private
+    private # rubocop:disable AccessModifierIndentation
 
       # Initialize null color
       #
--- a/config/flay.yml
+++ b/config/flay.yml
@@ -1,3 +1,3 @@
 ---
 threshold: 18
-total_score: 1172
+total_score: 1168
--- a/config/rubocop.yml
+++ b/config/rubocop.yml
@@ -11,6 +11,7 @@ BlockNesting:
 
 # Align with the style guide.
 CollectionMethods:
+  Enabled: true
   PreferredMethods:
     collect:  'map'
     inject:   'reduce'
--- a/config/rubocop.yml
+++ b/config/rubocop.yml
@@ -129,7 +129,7 @@ ParallelAssignment:
 
 # Allow additional specs
 ExtraSpacing:
-  Enabled: false
+  AllowForAlignment: true
 
 # Buggy
 FormatParameterMismatch:
--- a/config/rubocop.yml
+++ b/config/rubocop.yml
@@ -155,3 +155,7 @@ IndentArray:
 # My style is not supported
 MultilineMethodCallIndentation:
   Enabled: false
+
+# Align keys and values in a multiline hash
+Style/AlignHash:
+  EnforcedColonStyle: table
--- a/config/rubocop.yml
+++ b/config/rubocop.yml
@@ -17,14 +17,8 @@ CollectionMethods:
     find:     'detect'
     find_all: 'select'
 
-# Do not force public/protected/private keyword to be indented at the same
-# level as the def keyword. My personal preference is to outdent these keywords
-# because I think when scanning code it makes it easier to identify the
-# sections of code and visually separate them. When the keyword is at the same
-# level I think it sort of blends in with the def keywords and makes it harder
-# to scan the code and see where the sections are.
 AccessModifierIndentation:
-  Enabled: false
+  EnforcedStyle: outdent
 
 # Limit line length
 LineLength:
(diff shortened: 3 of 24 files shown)


lib/mutant/color.rb
```diff
@@ -24,7 +24,7 @@ def format(text)
         text
       end
 
-    private
+    private # rubocop:disable AccessModifierIndentation
```
backus: This line is disabled because the class definition begins at column 12 so rubocop tries to indent this `private` to column 12 as well

backus: @mbj For the sake of consistent indentation rules would you prefer either of the following?

```ruby
module Mutant
  class Color

    # ...

    NONE =
      Class.new(self) do

        # Format null color
        #
        # @param [String] text
        #
        # @return [String]
        #   the argument string
        def format(text)
          text
        end

      private

        # Initialize null color
        #
        # @return [undefined]
        def initialize
        end

      end.new

  end
end
```

or

```ruby
module Mutant
  class Color

    # ...

    NONE = Class.new(self) do

             # Format null color
             #
             # @param [String] text
             #
             # @return [String]
             #   the argument string
             def format(text)
               text
             end

           private

             # Initialize null color
             #
             # @return [undefined]
             def initialize
             end

           end.new

  end
end
```

I personally prefer the first case.

backus:
@mbj @dkubb The project rubocop config reads

```yml
# Prefer String#% over Kernel#sprintf
FormatString:
  Enabled: false
```

Rubocop can enforce using `String#%` so if that is the project preference I can make that change. With that said, the project uses the `Kernel#format` method in many places:

```sh
Offenses:

lib/mutant/matcher/method/singleton.rb:68:24: C: Favor String#% over format.
              env.warn(format('Can only match :defs on :self or :const got %s unable to match', receiver.type.inspect))
                       ^^^^^^
lib/mutant/reporter/cli/format.rb:71:13: C: Favor String#% over format.
            format(Printer::Config, env.config)
            ^^^^^^
lib/mutant/reporter/cli/format.rb:78:13: C: Favor String#% over format.
            format(Printer::StatusProgressive, status)
            ^^^^^^
lib/mutant/reporter/cli/format.rb:116:13: C: Favor String#% over format.
            format(Printer::Status, status)
            ^^^^^^
lib/mutant/reporter/cli/printer.rb:78:16: C: Favor String#% over format.
          puts(format(string, *arguments))
               ^^^^^^
lib/mutant/reporter/cli/printer.rb:85:39: C: Favor String#% over format.
          puts(colorize(status_color, format(string, *arguments)))
                                      ^^^^^^
spec/support/corpus.rb:94:14: C: Favor String#% over format.
        puts format(
             ^^^^^^
spec/support/corpus.rb:196:16: C: Favor String#% over format.
          puts format('Starting - %s', path)
               ^^^^^^
spec/support/corpus.rb:210:16: C: Favor String#% over format.
          puts format('Mutations - %4i - %s', count, path)
               ^^^^^^
```

I assume you guys want to be consistent. Which do you prefer?

lib/mutant/color.rb
backus: /cc @dkubb on this question too ^

dkubb: Given the choice I prefer the first case too.

backus: Do you prefer it over the way it is currently written (meaning you therefore prefer just leaving in the `rubocop:disable`)?

dkubb:
> With that said, the project uses the Kernel#format method in many places:

@backus Is it actually `Kernel#format`? I'd be surprised if @mbj used it directly like that.

lib/mutant/color.rb
dkubb: I think I prefer the original code actually. I would keep the existing formatting if it were up to me.

backus:
> @backus Is it actually Kernel#format? I'd be surprised if @mbj used it directly like that.

@dkubb There is a class or two which defines its own `#format` but I believe this is a normal `Kernel#format` for example: https://github.com/mbj/mutant/blob/master/lib/mutant/matcher/method/singleton.rb#L68

lib/mutant/color.rb
backus: Yeah I think I agree. I think the cop is useful overall though so I hope that doesn't translate to removing the cop too

mbj:
> @backus Is it actually Kernel#format? I'd be surprised if @mbj used it directly like that.

Very old code. Before I begun to dislike the Kernel infection. Also rubocop default from years back preferred the kernel method.

mbj:
@backus Needs to be rebased. If you want to fix `Kernel#format` use in this cycle: Go for it.

backus:
@mbj rebased and updated `FormatString` cop

PR #549, opened 2016-03-19
Fix misspelled spec file

backus:
I noticed that the bootstrap spec's filename was misspelled ("boostrap"). I've fixed that. I also took the time to check for specs where the specified constant does not match the file path.

Note: I believe it is also good practice to reflect the described method in the filename if you do something like `RSpec.describe Foo, '#bar'` but I did not make the dozen changes this would entail. I'm happy to do so if you agree


backus:
@dkubb @mbj this should be ready for review

mbj:
@backus Thx for that. OT: In `xspec` the file name should be asserted.

PR #551, opened 2016-03-21
Fix visiblity of NamedChildren methods

mbj:
Another extraction of #545

### DIFF
--- a/spec/unit/mutant/ast/named_children_spec.rb
+++ b/spec/unit/mutant/ast/named_children_spec.rb
@@ -8,6 +8,14 @@ RSpec.describe Mutant::AST::NamedChildren do
       end
     end
 
+    def publish
+      klass.class_eval do
+        public :foo, :bar
+
+        public :remaining_children, :remaining_children_indices, :remaining_children_with_index
+      end
+    end
+
     let(:instance) { klass.new(node) }
 
     let(:node_foo) { s(:foo) }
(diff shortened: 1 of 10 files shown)


mbj:

Context: We include the `NamedHelpers` also for the `Mutant::Mutator::Node*` namespace, and here the LSP compatible interface is only `#call`, where the `NamedHelpers` used to create public methods by default.

The `AST::Meta` namespace is not inheriting from a shared subclass, so the "arbitrarily different" public interfaces are correct (and used).

I plan to remove having to re-specify the structure of the AST in both namespaces, but thats another PR.

spec/unit/mutant/ast/named_children_spec.rb
```diff
@@ -17,26 +25,42 @@
...
     describe 'generated methods' do
+      specify 'are private by default' do
+        %i[
+          foo
+          bar
+          remaining_children
+          remaining_children_indices
+          remaining_children_with_index
+        ].each do |name|
+          expect(klass.private_instance_methods.include?(name)).to be(true)
```
dkubb: What about: `expect(klass.private_instance_methods).to include(name)`

mbj: I do not like the truthyness check it internally delegates to. I do want a `#equal?`. On the return.

PR #557, opened 2016-03-30
Explain RSpec config options

wied03:
Will still work with RSpec runner/config code runs

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,17 @@ To mutation test Rails models with rspec comment out ```require 'rspec/autorun'`
 RAILS_ENV=test bundle exec mutant -r ./config/environment --use rspec User
 ```
 
+Passing in RSpec Options
+------------------------
+
+**NOTE: Experimental**
+
+You can control some aspects of RSpec using the `SPEC_OPTS` environment variable as usual. If you want mutant to only pay attention to specs in a certain directory, you can run
+
+```sh
+SPEC_OPTS="--pattern spec/subdir_only/**/*_spec.rb" mutant --use rspec SomeClass
+```
+
 Limitations
 -----------
 


wied03:
Was going to create an issue and then realized you could do this. Thought it might make it explicit in the docs in case it's not obvious to others.

mbj:
@wied03 I like this. I'd never use it by myself as my workflow is different, but for some people it'll be useful. Before I merge it I should spend a bit more time thinking about this documentation addition as it would "cement" a public interface.

wied03:
In my case, I have a bunch of Opal scripts (and specs) that won't run under MRI, so that's what prompted this.

mbj:
> In my case, I have a bunch of Opal scripts (and specs) that won't run under MRI, so that's what prompted this.

You can exclude these already via `mutant: false` rspec metadata, is that an option in your case?

wied03:
The problem is not with those JS specs running when they shouldn't, more that they `require` files that don't exist with MRI

mbj:
I see, here it makes sense mutant not even trying to load these specs. I think we should document your workaround, but label it as experimental. As we always wanted to add an rspec option pass through.

wied03:
Sorry to piggy back with a noob question, but do the subjects you provide to mutant pass through to RSpec as filters? From what you're saying, it looks like they do, but wasn't sure

mbj:
I meant I wanted to add an explicit CLI option pass through to the mutant CLI. `--rspec-pass="$rspec-options-here"`.

mbj:
@wied03 Can you add a note that this is an experiemental API? Than I'd be ready to merge it.

wied03:
@mbj - on your 1st comment, I get what you meant, I was just taking advantage of this to understand how mutant works in an unrelated way.

on the 2nd comment - sure

mbj:
> @mbj - on your 1st comment, I get what you meant, I was just taking advantage of this to understand how mutant works in an unrelated way.

Maybe I do not get your question. Can you formulate it in a different way?

wied03:
When I pass in a bunch of subjects on the mutant command line, is that only telling mutant what to mutate or is it also telling RSpec which specs (by the describe subject) to run?

mbj:
> When I pass in a bunch of subjects on the mutant command line, is that only telling mutant what to mutate or is it also telling RSpec which specs (by the describe subject) to run?

Only what mutant should mutate, specs are selected after loading *all* rspec examples based on rspec metadata.

wied03:
@mbj - Should the doc also say then that you can use RSpec's `--example` switch to control that? I haven't thought about this much at yet as to whether you want to keep the scope of which tests run and what you mutate in sync.

Also I can squash those commits if you need them to be.

mbj:
> @mbj - Should the doc also say then that you can use RSpec's --example switch to control that? I haven't thought about this much at yet as to whether you want to keep the scope of which tests run and what you mutate in sync.

Mutant needs to deeply hook into rspec to run subsets of examples after selection very efficient. At a much finer granularity and with less overhead than rspec API itself allows. You have to carefully test each rspec flags as some do not play well with the way mutant hooks rspec.

You normally want to have your rspec examples written in a way manual selection is not needed. All my projects are done this way and mutant typically only supports what I need.

wied03:
Right, but does anyone run Mutant on all of their subjects in their application in the same run or do they break it down into chunks? If they break it down into chunks but the whole spec suite is running, doesn't that make the coverage look worse than it is? Tests might run that don't hit the subjects all, so they won't fail. This might be a complete noob question.

mbj:
@wied03 Mutant selects per subject a subset of your tests to run, to target the kills against relevant specs.

So not your full suite is run per mutation, a well choosen subset.

The definition of "well choosen" depends on if you align your rspec metadata to the needs of mutant.

mbj:
@wied03 This is the code: https://github.com/mbj/mutant/blob/master/lib/mutant/selector/expression.rb#L7-L21

mbj:
@dkubb, @backus Could @wied03 Ask for what I describe as "real" vs "implicit" coverage?

mbj:
@wied03 And yes, please squash the commits ;)

mbj:
@wied03 thx.

PR #560, opened 2016-04-03
Add mutation from `foo[n..-1]` → `foo.drop(n)`

backus:


backus:
@mbj @dkubb this is ready for review

PR #561, opened 2016-04-04
Add multiple assignment mutations

backus:
This PR adds several new mutations for parallel assignment

### DIFF
--- a/meta/lvasgn.rb
+++ b/meta/lvasgn.rb
@@ -26,8 +26,8 @@ Mutant::Meta::Example.add :lvasgn do
   # TODO: fix invalid AST
   #   These ASTs are not valid and should NOT be emitted
   #   Mutations of lvarasgn need to be special cased to avoid this.
-  mutation s(:begin, s(:lvasgn, :a__mutant__, s(:nil)), s(:lvar, :a))
-  mutation s(:begin, s(:nil), s(:lvar, :a))
-  mutation s(:begin, s(:self), s(:lvar, :a))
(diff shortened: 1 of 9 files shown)


backus:
@dkubb @mbj ready for review

meta/lvasgn.rb
```diff
@@ -6,3 +6,28 @@
...
+
+Mutant::Meta::Example.add :lvasgn do
+  source 'a = nil; a'
+
+  mutation 'a = nil; nil'
+  mutation 'a = nil; self'
+  mutation 'a = nil'
+  # TODO: fix invalid AST
+  #   These ASTs are not valid and should NOT be emitted
+  #   Mutations of lvarasgn need to be special cased to avoid this.
```
mbj: I think we should not add more "invalid AST emitted" to the code base. Lets fix the lvasgns first?

backus: Is the issue here that we shouldn't replace an assignment with a noop in general? Or is it a question of not eliminating the assignment when we can statically check if the lvar is used below?

mbj: > Is the issue here that we shouldn't replace an assignment with a noop in general? Or is it a question of not eliminating the assignment when we can statically check if the lvar is used below?

I was woundering if we should not add to the "mutation operator can cause invalid AST but correct source" AKA "unintend source" problem, and instead fix it before.

mbj:
Closing this PR, as all contributions will require a CLA from now on.

Issue #291, opened 2015-01-22
Add a mutation for kwargs


dkubb:
When using mutant on some code that uses kwargs, it came up with only one alive mutation:

```diff
-def initialize(name, includes:, **options)
+def initialize(name__mutant__, includes:, **options)
   super
 end
```

Now obviously I can kill this one by renaming `name` to `_name`, but I noticed there were no mutations for `includes` to `includes__mutant` nor for `**options` to `**options__mutant__`. I think both of these should be added to make sure these options are being handled.

PR #562, opened 2016-04-05
Add keword arument mutations

backus:
Adds mutations for keword arguments (closing #291):

| Source                   | Mutation                  |
| :---------------------   | :---------------------    |
| `def foo(bar:); end`     | `def foo(_bar:); end`     |
| `def foo(bar: baz); end` | `def foo(_bar: baz); end` |
| `def foo(bar: baz); end` | `def foo(bar:); end`      |

### DIFF
--- a/config/flay.yml
+++ b/config/flay.yml
@@ -1,3 +1,3 @@
 ---
 threshold: 16
-total_score: 1160
+total_score: 1166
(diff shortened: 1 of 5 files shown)


backus:
@dkubb @mbj this is ready for review

PR #564, opened 2016-04-15
Add mutation from `/regex/i` → `/regex/`

backus:
Related to #18


lib/mutant/mutator/node/regopt.rb
```diff
@@ -0,0 +1,31 @@
+module Mutant
+  class Mutator
+    class Node
+
+      # Regular expression options mutation
+      class Regopt < self
+
+        MUTATED_FLAGS = IceNine.deep_freeze([:i])
```
mbj: Lets use `%i[]` here? AFAIK mutant does not support rubies where this syntax does not exist.

dkubb:
@backus I see no problem with this. @mbj do you mind if I merge, or do you want to merge?

dkubb:
@backus would it make sense for the other flags to be mutated away?

mbj:
AFAIK yes, removing: `m`, `x` are sure to be semantic reducing operators (The resulting regexp accepts a smaller language). `o` needs to be more carefuly evaluated.

backus:
> @backus would it make sense for the other flags to be mutated away?

I mutated every option in my first iteration on this PR and I changed to what was merged because those cases were more complex. Honestly, even though it has only been 5 days, I don't totally remember those reasons although I'm sure they would standout to me again if I went to implement them.

Issue #19, opened 2013-01-25
Mutate regexp body


mbj:
The rbx AST exposes the regexp as ```#source``` attribute that containts the source as string. This string could be parsed by a regexp library and the AST could be mutated by an additional mutator tree ```Mutant::Mutator::Regexp```. This will outline many missing testcases.

backus:
Closed by #565! :smile: 🎈

PR #565, opened 2016-04-18
Mutate regexp body

backus:
* Adds a bijective mapper to and from the `Regexp::Parser` domain
* Supports unparsing of all Regexp mutations
* Adds PoC mutation `/^/` → `/\A/`


mbj:
Related to #19

lib/mutant/ast/types.rb
```diff
@@ -38,8 +38,28 @@ module Types
...
+        keep
+        quantifier
+        free_space
+        property
+        escape
+      ].map { |entry| "regexp_#{entry}" })
+
       # Nodes that are NOT generated by parser but used by mutant / unparser.
-      EXTRA = symbolset.(%w[empty])
+      EXTRA = symbolset.(%w[empty]) + REGEXP
```
dkubb: I'm not sure if `EXTRA` will be frozen after appending `REGEXP`.

dkubb:
@backus this is awesome! Feel free to ping me for reviews about this anytime; this is one of the features I've wanted most in mutant and I'd love to help our however I can.

I would say something like this should be merged early once you've got one or two basic mutations working reliably and @mbj has approved it. Your work provides an outline, and we can gradually fill in various other mutations over time.

lib/mutant/mutator/node/regexp.rb
```diff
@@ -0,0 +1,104 @@
...
+            Node.new(Regexp::Parser.parse(source, 'ruby/2.3'))
+          end
+
+          class Node < DelegateClass(Regexp::Expression::Base)
+            def to_source
+              to_s
+            end
+
+            def with(expressions)
+              new_node =
```
dkubb: I realize you didn't ask for review, but I figured I'd mention this anyway. This code is a bit inconsistent with code in Node#dispatch. In that code you explicitly `dup`, then mutate the object then use the object elsewhere. I don't mind using `tap` like this, but consistency is probably more important, so I would write this as:

```ruby
def with(expression)
  new_node = dup
  new_node.expressions = expressions
  self.class.new(new_node)
end
```

dkubb:
@backus does [regexp_parser](https://github.com/ammar/regexp_parser) support parsing modifiers? Or is that something more in the realm of what [ast](https://github.com/whitequark/ast) supports? Not saying they need to be handled in this PR, but it's something we may want to think about mutating in the future; afaik they aren't handled yet.

mbj:
@dkubb the modifiers are part of the AST format.

lib/mutant/mutator/node/regexp.rb
```diff
@@ -0,0 +1,104 @@
+module Mutant
+  class Mutator
+    class Node
+
+      # Generic mutator
+      module RegularExpression
+        class Parser
+          def self.parse(source)
+            Node.new(Regexp::Parser.parse(source, 'ruby/2.3'))
```
mbj: The regexp version should likely be reflected from `RUBY_VERSION`.

lib/mutant/mutator/node/regexp.rb
```diff
@@ -0,0 +1,104 @@
...
+          end
+
+          class Node < DelegateClass(Regexp::Expression::Base)
```
mbj: Instead to use an interface with mutable API, what about mapping the `Regexp::Expression::Base` to a `Parser::AST` node on custom `type` so we can use the same machinery, and reduce the need to even fight mutable state.

meta/regular_expression.rb
```diff
@@ -0,0 +1,14 @@
+Mutant::Meta::Example.add :'regexp_expression' do
+  source '/(foo|bar)/'
```
mbj: This is IMO a more complex example than needed for the first iteration, as it contains an OR and a grouping, lets reduce it to `/foo|bar/` to focus on only one thing?

backus:
> @backus does regexp_parser support parsing modifiers? Or is that something more in the realm of what ast supports? Not saying they need to be handled in this PR, but it's something we may want to think about mutating in the future; afaik they aren't handled yet.

@dkubb see #564 :smile

lib/mutant/ast/types.rb
backus: Oh good point

lib/mutant/mutator/node/regexp.rb
backus: Yeah I like this more

lib/mutant/mutator/node/regexp.rb
backus: If you are ok with it I think this might might be worth punting. Right now, if you only swap out the `#expressions` attribute on a `Regexp::Expression` instance, then you get unparsing for free. This might be much more complex if everything is mapped to a `Parser::AST`. WDYT?

mbj: I think writing a bijective mapper from `Regexp::Expression` to `Parser::AST` is not that complex and will be worth it so we can keep the mutable API contained. It'll not leave the mutation engine for regexps.

backus:
I mentioned this as a disclaimer in the slack room but I'll repeat here: https://github.com/mbj/mutant/pull/565/commits/bcfc441494008a8e461b3da8b214f148c5d14b52 is very much WIP so feel free to take a look but don't worry about reviewing almost all of the code is likely to change.

meta/regular_expression.rb
backus: Just as a note, I'll probably do a simpler mutation like `^` -> `\A`. The representation for alternations is actually pretty complex compared to simple nodes like the beginning of string anchor

backus:
@dkubb @mbj this is still a work in progress but I'm now done with the first draft of the bijective regexp mapper (boy was that a lot of work). I would really love some feedback on the mapper code. This does not emit mutations right now because I've been focusing on mapping. Specifically, I'd love to hear what you think can be improved in terms of design / architecture.

@ammar I'd also appreciate your input if you wouldn't mind taking a look. I'm mapping everything from the `Regexp::Expression` domain and back so I'm using your gem very heavily here.

lib/mutant/ast/regexp/transformer/direct.rb
```diff
@@ -0,0 +1,97 @@
...
+          register :regexp_script_hiragana_property
+          register :regexp_script_katakana_property
+          register :regexp_letter_any_property
+          register :regexp_digit_type
+          register :regexp_space_type
+          register :regexp_word_type
+          register :regexp_nondigit_type
+          register :regexp_nonspace_type
+          register :regexp_nonword_type
+          register :regexp_dot_meta
```
mbj: @backus What about a `%[regexp_a regexp_b ...].each(&method(:register))` to avoid the duplication? Should also fix the flog score.

lib/mutant/ast/regexp/transformer/quantifier.rb
```diff
@@ -0,0 +1,95 @@
...
+          register :regexp_greedy_one_or_more
+          register :regexp_greedy_zero_or_one
+          register :regexp_possessive_zero_or_one
+          register :regexp_reluctant_zero_or_more
+          register :regexp_reluctant_one_or_more
+          register :regexp_possessive_zero_or_more
+          register :regexp_possessive_one_or_more
+          register :regexp_greedy_interval
+          register :regexp_reluctant_interval
+          register :regexp_possessive_interval
```
mbj: Instance of: https://github.com/mbj/mutant/pull/565/files#r61689896

lib/mutant/ast/regexp/transformer/recursive.rb
```diff
@@ -0,0 +1,49 @@
...
+      class Transformer
+        class Recursive < self
+          register :regexp_alternation_meta
+          register :regexp_nlookahead_assertion
+          register :regexp_passive_group
+          register :regexp_nlookbehind_assertion
+          register :regexp_lookbehind_assertion
+          register :regexp_lookahead_assertion
+          register :regexp_atomic_group
+          register :regexp_capture_group
```
mbj: Instance of: https://github.com/mbj/mutant/pull/565/files#r61689896

lib/mutant/ast/regexp/transformer/text.rb
```diff
@@ -0,0 +1,51 @@
...
+        class Text < self
+          register :regexp_literal_literal
+          register :regexp_comment_group
+          register :regexp_hex_escape
+          register :regexp_named_group
+          register :regexp_number_backref
+          register :regexp_name_call_backref
+          register :regexp_literal_escape
+          register :regexp_whitespace_free_space
+          register :regexp_codepoint_list_escape
```
mbj: Instance of: https://github.com/mbj/mutant/pull/565/files#r61689896

mbj:
@backus I do not have time for a full pass but did a small one.

lib/mutant/ast/regexp/transformer.rb
```diff
@@ -0,0 +1,82 @@
+module Mutant
+  module AST
+    module Regexp
+      class Transformer
+        include AbstractType
+
+        REGISTRY = {} # rubocop:disable MutableConstant
```
dkubb: Instead of a constant, what about using `@registry = {}` here? Then there's no question it's mutable.

Also, I assume that after some initial population of the registry, it's contents don't change again? If that's so, then I might suggest adding a `.finalize` class method that deep freezes the registry so it can't be changed at runtime.

lib/mutant/ast/regexp/transformer.rb
```diff
@@ -0,0 +1,82 @@
...
+
+        class ExpressionToAST
+          PREFIX = :regexp
+
+          include Concord.new(:expression), Procto.call, AST::Sexp, AbstractType, Adamantium
+
+        private
+
+          def children
+            return [] if expression.terminal?
```
dkubb: You can use `EMPTY_ARRAY` instead of `[]`.

lib/mutant/ast/regexp/transformer/alternative.rb
```diff
@@ -0,0 +1,27 @@
...
+          class ExpressionToAST < Transformer::ExpressionToAST
+            def call
+              quantify(ast(*children))
+            end
+          end
+
+          class ASTToExpression < Transformer::ASTToExpression
+            def call
+              root = ::Regexp::Expression::Alternative.new(0, 0, 0)
+
```
dkubb: I might suggest removing the blank lines in methods to be consistent with the formatting in https://github.com/mbj/mutant/pull/565/commits/b3b6c22e8868baee9691c75127e3baba833ce8c7#diff-c01564451f86f0ea4fa9bab318f0a5f1R16

lib/mutant/ast/regexp/transformer/quantifier.rb
```diff
@@ -0,0 +1,97 @@
...
+                suffix
+              end
+            end
+
+            def type
+              QUANTIFIER_MAP.fetch(node.type).first
+            end
+
+            def suffix
+              QUANTIFIER_MAP.fetch(node.type).fetch(1)
```
dkubb: What about defining 3 constants so that you aren't using the "magic number" `1` here? Then you can use Hash#fetch in `#type` and `#mode` similarly to `#suffix`.

lib/mutant/ast/regexp/transformer/quantifier.rb
```diff
@@ -0,0 +1,97 @@
...
+            end
+
+            def mode
+              QUANTIFIER_MAP.fetch(node.type).last
+            end
+
+            def interval_text
+              return "{#{min}}" if min.equal?(max)
```
dkubb: What about something like:

```ruby
def interval_text
  interval = [min, max].map { |num| num if num.positive? }.uniq
  "{#{interval.join(',')}}"
end
```

lib/mutant/ast/regexp/transformer/quantifier.rb
```diff
@@ -0,0 +1,97 @@
...
+              regexp_reluctant_one_or_more:   [:one_or_more,  '+?', :reluctant],
+              regexp_possessive_zero_or_more: [:zero_or_more, '*+', :possessive],
+              regexp_possessive_one_or_more:  [:one_or_more,  '++', :possessive],
+              regexp_greedy_interval:         [:interval,     '',   :greedy],
+              regexp_reluctant_interval:      [:interval,     '?',  :reluctant],
+              regexp_possessive_interval:     [:interval,     '+',  :possessive]
+            )
+
+            def call
+              expression = Regexp.to_expression(subject)
```
dkubb: Instance of https://github.com/mbj/mutant/pull/565/commits/b3b6c22e8868baee9691c75127e3baba833ce8c7#r61763338

lib/mutant/ast/regexp/transformer/recursive.rb
```diff
@@ -0,0 +1,51 @@
...
+              regexp_nlookbehind_assertion: [::Regexp::Token.new(:assertion, :nlookbehind, '(?<!'), ::Regexp::Expression::Assertion::NegativeLookbehind],
+              regexp_lookbehind_assertion:  [::Regexp::Token.new(:assertion, :lookbehind,  '(?<='), ::Regexp::Expression::Assertion::Lookbehind],
+              regexp_lookahead_assertion:   [::Regexp::Token.new(:assertion, :lookahead,   '(?='),  ::Regexp::Expression::Assertion::Lookahead],
+              regexp_atomic_group:          [::Regexp::Token.new(:group,     :atomic,      '(?>'),  ::Regexp::Expression::Group::Atomic],
+              regexp_capture_group:         [::Regexp::Token.new(:group,     :capture,     '('),    ::Regexp::Expression::Group::Capture]
+            )
+            # rubocop:enable LineLength
+
+            def call
+              expression = expression_class.new(expression_token)
```
dkubb: Instance of https://github.com/mbj/mutant/pull/565/commits/b3b6c22e8868baee9691c75127e3baba833ce8c7#r61765189

lib/mutant/ast/regexp/transformer/root.rb
```diff
@@ -0,0 +1,27 @@
...
+          class ExpressionToAST < Transformer::ExpressionToAST
+            def call
+              quantify(ast(*children))
+            end
+          end
+
+          class ASTToExpression < Transformer::ASTToExpression
+            def call
+              root = ::Regexp::Expression::Root.new
+
```
dkubb: Instance of https://github.com/mbj/mutant/pull/565/commits/b3b6c22e8868baee9691c75127e3baba833ce8c7#r61765189

lib/mutant/ast/regexp/transformer/recursive.rb
```diff
@@ -0,0 +1,51 @@
...
+
```
dkubb: Instance of https://github.com/mbj/mutant/pull/565/commits/b3b6c22e8868baee9691c75127e3baba833ce8c7#r61765189

lib/mutant/ast/regexp/transformer/quantifier.rb
```diff
@@ -0,0 +1,97 @@
...
+
```
dkubb: Instance of https://github.com/mbj/mutant/pull/565/commits/b3b6c22e8868baee9691c75127e3baba833ce8c7#r61765189

lib/mutant/mutator/node/regexp.rb
```diff
@@ -0,0 +1,40 @@
...
+
+          def dispatch
+          end
+        end
+
+        class RootExpression < Node
+          handle(:regexp_root_expression)
+
+          def dispatch
+            children.each_index do |index|
```
dkubb: You could write: `children.each_index(&method(:mutate_child))`

spec/integrations.yml
```diff
@@ -88,6 +88,8 @@
     - optional/capi/string_spec.rb
     '#<RegexpError: invalid multibyte escape: /\xAA/>':
     - language/regexp/escapes_spec.rb
+    "#<Regexp::Scanner::PrematureEndError: Premature end of pattern at #{str}>":
+    - language/regexp/interpolation_spec.rb
```
dkubb: I'd suggest indenting this underneath the string, assuming it is still valid YAML syntax.

dkubb:
@back

dkubb:
@backus I did a pass over the second commit in this PR.

lib/mutant/ast/regexp/transformer/quantifier.rb
dkubb: Also note that `Numeric#positive?` is only available in ruby 2.3.0+. I don't know offhand what versions of ruby mutant targets officially, but this might be something to keep in mind.

mbj: Mutant currently targets 2.2+.

lib/mutant/ast/regexp/transformer/direct.rb
```diff
@@ -0,0 +1,99 @@
...
+              quantify(ast)
+            end
+          end
+
+          class ASTToExpression < Transformer::ASTToExpression
+            include LookupTable
+
+            # rubocop:disable LineLength
+            TABLE = IceNine.deep_freeze(
+              regexp_one_or_more_escape:       [::Regexp::Token.new(:escape,   :one_or_more,      '\+'),           ::Regexp::Expression::EscapeSequence::Literal],
```
mbj: I dunno if its worth it, but we may be able to remove the repeated `::Regexp::Token.new` calls via placing the arguments in the value array and later mapping over the values to instantiate the token object.

This reduces literall call side duplication.

```ruby
[
  [:regexp_one_or_more_escape, [:escape, :one_or_more, '\+'], ::Regexp::Expression::EscapeSequence::Literal],
  ...
].map { |type, arguments, klass| [type, [::Regexp::Token.new(*arguments), klass] }.to_h
```

Will also help to reduce flay / flog score increases.

lib/mutant/ast/regexp/transformer/quantifier.rb
```diff
@@ -0,0 +1,97 @@
...
+            end
+          end
+
+          class ASTToExpression < Transformer::ASTToExpression
+            include NamedChildren
+
+            children :min, :max, :subject
+
+            QUANTIFIER_MAP = IceNine.deep_freeze(
+              regexp_greedy_zero_or_more:     [:zero_or_more, '*',  :greedy],
```
mbj: Instead of mapping to an array of 3 elements (effectively a tulpe) and having to remember the indexes each tuple member is represended with, what about using a dedicated value object?

```
class Quantifier
  include Concord::Public.new(:type, :suffix, :mode)
end
```ruby
QUANTIFIER_MAP = IceNine.deep_freeze({
  regexp_greedy_zero_or_more:     [:zero_or_more, '*',  :greedy]
}.map { |ast_type, arguments| [ast_type, Quantifier.new(*arguments)] }.to_h)
```

This will also remove or eliminate the "ugly" (due `Array#{first,last} non strictness) API usage in `#type`, `#mode`, `#suffix` implementations above.

lib/mutant/ast/regexp/transformer.rb
backus: Will do. I have some local code which extracts a generic `Mutant::TypeRegistry` which I'll insert here (shared behavior with `Mutant::Mutator::Registry`) and I'll probably implement the finalize there

backus: To be clear, I'm not talking about extracting `TypeRegistry` and having `Mutant::Mutator::Registry` inherit from it. It looks like this instead:

```ruby
module Mutant
  # Generic registry for mapping types
  class TypeRegistry
    include Concord.new(:contents)

    # Initialize object
    #
    # @return [undefined]
    def initialize
      super({})
    end

    # Raised when the type is an invalid type
    RegistryError = Class.new(TypeError)

    # Register entry class for AST node class
    #
    # @param [Symbol] type
    # @param [Class] entry
    #
    # @return [self]
    def add(type, entry)
      fail RegistryError, "Invalid type registration: #{type.inspect}" unless AST::Types::ALL.include?(type)
      fail RegistryError, "Duplicate type registration: #{type.inspect}" if contents.key?(type)
      contents[type] = entry
      self
    end

    # Lookup entry for node
    #
    # @param [Symbol] type
    #
    # @return [Class]
    #
    # @raise [RegistryError]
    #   raises error when entry is not found in registry
    def lookup(type)
      contents.fetch(type) do
        fail RegistryError, "No entry for: #{type.inspect}"
      end
    end

  end # Registry

end # Mutant
```

```ruby
module Mutant
  class Mutator
    # Registry for mutators
    class Registry
      include Concord.new(:type_registry)

      # Initialize object
      #
      # @return [undefined]
      def initialize
        super(TypeRegistry.new)
      end

      # Register mutator class for AST node class
      #
      # @param [Symbol] type
      # @param [Class:Mutator] mutator
      #
      # @return [self]
      def register(type, mutator)
        type_registry.add(type, mutator)
        self
      end

      # Call type_registry
      #
      # @return [Enumerable<Mutation>]
      def call(node, parent = nil)
        type_registry.lookup(node.type).call(node, parent)
      end

    end # Registry

    REGISTRY = Registry.new

  end # Mutator
end # Mutant
```

lib/mutant/ast/regexp/transformer/quantifier.rb
backus: Ah good catch re: `#positive?`

backus: @dkubb that is pretty clever (in a good way). I'll probably make that the implementation

lib/mutant/ast/regexp/transformer.rb
backus: @dkubb regarding using `@registry`: I'm not how well that will work due to strange inheritance behavior. I could use `@@registry` I guess

dkubb: @backus FWIW, I was going to suggest extracting the registry into it's own data structure. Often when you have a local object with lots of methods related to adding/removing/changing it, that's a sign that you have a concept that can be extracted into something more self contained.

dkubb: @backus although I am not entirely sure I understand what the separation between `Registry` and `TypeRegistry` is in the example above. Given that `Registry` has no other state besides a `TypeRegistry` instance I'm not sure why you don't collapse them down and have a single `Registry` object that you interact with.

backus: @dkubb that is essentially what I've done with this `TypeRegistry`. It still needs to be a constant or a updatable at some class level

backus: > @backus although I am not entirely sure I understand what the separation between `Registry` and `TypeRegistry` is in the example above. Given that `Registry` has no other state besides a `TypeRegistry` instance I'm not sure why you don't collapse them down and have a single `Registry` object that you interact with.

So the difference is the `TypeRegistry` is just responsible for mapping types to objects. It fails if you try to register something already registered. It fails if you try to register a key that is not a valid ast type. `Mutant::Mutator::Registry` builds on top and supplies the `call` method which does a `lookup` on a node type and then calls the mutator. This was the interface before which is why I preserved it

dkubb: > This was the interface before which is why I preserved it

Ahh I see. @mbj any reason for the separation? I could see there being a reason iff the `Registry` was going to be refactored to have different kinds of instances injected into it.

Assuming there's not a compelling reason I would recommend collapsing the implementation and exposing only the `Registry` public methods in the interface.

backus: > Assuming there's not a compelling reason I would recommend collapsing the implementation and exposing only the Registry public methods in the interface.

By this you mean reducing the registry down to just registration and then replacing instances of `REGISTRY.call` with `REGISTRY.lookup(...).call(...)`?

lib/mutant/ast/regexp/transformer/quantifier.rb
backus: I implemented @mbj's input which addressed this problem with a different approach

lib/mutant/ast/regexp/transformer.rb
dkubb: I was thinking more something like this:

```ruby
module Mutant
  class Mutator
    class Registry
      Error = Class.new(TypeError)

      def initialize
        @entries = Hash.new do |_entries, type|
          fail Error, "No entry for: #{type.inspect}"
        end
      end

      def register(type, mutator)
        fail Error, "Invalid type registration: #{type.inspect}" unless AST::Types::ALL.include?(type)
        fail Error, "Duplicate type registration: #{type.inspect}" if @entries.key?(type)
        @entries[type] = mutator
        self
      end

      # XXX: what is the `parent` argument for in the interface?
      def call(type, parent = nil)
        @entries[type]
      end

    end # Registry

    REGISTRY = Registry.new

  end # Mutator
end # Mutant
```

lib/mutant/ast/regexp/transformer/quantifier.rb
dkubb: FWIW I thought @mbj's approach was superior so glad you went that way.

lib/mutant/ast/regexp/transformer.rb
backus: >       # XXX: what is the `parent` argument for in the interface?

`parent` argument is for the mutator. The you call it like so:

 ```ruby
REGISTRY.call(s(:str, 'foo'), s(:regexp, s(:str, 'foo'), s(:regopt))
```

Comment (backus) [inline lib/mutant/ast/regexp/transformer.rb]: I think the `Registry` should just be reduced down and the `call` should be moved to `Mutator.call` which assumes a `REGISTRY` constant

Comment (mbj) [inline lib/mutant/ast/regexp/transformer.rb]: > I think the Registry should just be reduced down and the call should be moved to Mutator.call which assumes a REGISTRY constant

I had this in a local branch for cleaning up the mutators but never pushed it so far.

Comment (backus) [inline lib/mutant/ast/regexp/transformer.rb]: > I had this in a local branch for cleaning up the mutators but never pushed it so far.

Mind if I just make that change then in this PR?

File: lib/mutant/ast/regexp/transformer/recursive.rb [hunk 24]
```diff
@@ -0,0 +1,49 @@
...
+            def call
+              quantify(ast(*children))
+            end
+          end
+
+          class ASTToExpression < Transformer::ASTToExpression
+            include LookupTable
+
+            # rubocop:disable LineLength
+            TABLE = IceNine.deep_freeze(
```
mbj: Instance of: https://github.com/mbj/mutant/pull/565/files#r61767066

lib/mutant/ast/regexp/transformer/text.rb
```diff
@@ -0,0 +1,55 @@
...
+            def call
+              quantify(ast(expression.text))
+            end
+          end
+
+          class ASTToExpression < Transformer::ASTToExpression
+            include LookupTable
+
+            # rubocop:disable LineLength
+            TABLE = IceNine.deep_freeze(
```
mbj: Instance of: https://github.com/mbj/mutant/pull/565/files#r61767066

config/flog.yml
```diff
@@ -1,2 +1,2 @@
 ---
-threshold: 27.7
+threshold: 33.8
```
mbj: Thats a more than 25% increase, I think there is more duplication to fight to remove some AST complexity till we can merge this. There are unaddressed flags likely to contribute:

* https://github.com/mbj/mutant/pull/565/files#r61767066 (with some instances)

spec/unit/mutant/ast/regexp_spec.rb
```diff
@@ -0,0 +1,226 @@
...
+      expression = Mutant::AST::Regexp.to_expression(ast)
+      round_trip = expression.to_re
+
+      expect_exp_text = indent(YAML.dump(Regexp::Parser.parse(regex)))
+      actual_exp_text = indent(YAML.dump(Mutant::AST::Regexp.to_expression(ast)))
+
+      diff = Mutant::Diff.build(expect_exp_text, actual_exp_text).colorized_diff
+
+      fail_message = <<-MSG
+Failed performing round trip:
```
mbj: There is an strip indent helper you can use via `fail_message = strip_indent(<<-MSG)` to reduce the need to break the indentation flow.

lib/mutant/ast/regexp/transformer.rb
mbj: Maybe extract it to a dedicated one? This one is huge and should not get blown up more. Ideally even gets broken down in smaller ones.

backus: Sure that works. I guess my real question was "mind if I handle implementing that" so another PR works for me

config/flog.yml
backus: *nod* yeah I don't view this as mergeable yet (hopefully the state of the specs makes that obvious). Just approaching mergeable. Hopefully one more weekend of work on my end will get us to 99% ready

mbj: Flay / Flog scores are not to be interpreted absolute, only relative to the code change. Ideally both grow logarithmic to the amount of complexity.

spec/unit/mutant/ast/regexp_spec.rb
backus: Cool. This part of the spec was also sort of messy and verbose to aid my TDD flow since I had to add 70 new AST types. I plan to clean this file up and make the output a bit less verbose

config/flog.yml
backus: > Thats a more than 25% increase ... Ideally Flay / Flog scores grow logarithmic to the amount of complexity.

Yeah I agree. I will note that this PR *should* necessarily be a big jump since the number of ast types is almost doubling:

2.3.0 :005 > (Mutant::AST::Types::ALL - Mutant::AST::Types::REGEXP).size
 => 94
2.3.0 :006 > Mutant::AST::Types::ALL.size
 => 162

mbj: The number of inhabitants of a specific already existing context typically does not result in a big jump of the global scores.

backus: > The number of inhabitants of a specific already existing context 

I'm not sure I'm following what you mean by this could you elaborate?

mbj: We had N types before, and now have N*2 types. But the concept of registred type existed already. And the concept of registred type is only a small subset of mutants overall complexity presented in the both scores.

So a ~25% increase of a global score does not map quite right to the amount of complexity added.

backus: Yeah I definitely expect this score and flay to go down a fair amount before merging. I was more so using the number of node types as a proxy for increased complexity in general.

mbj: I'm more concerned about flog, flay seems to be fitting.

backus: Ah yeah. So the flog score only jumped because right now `Mutant::Mutator::Node::Literal::Regex#dispatch` is intentionally lazy and incomplete. I can probably fix that tonight. That method just hasn't seen any love since the initial PoC because I've been almost entirely focused on the bijective mapping

lib/mutant/mutator/node/regexp.rb
backus: > I think writing a bijective mapper from `Regexp::Expression` to `Parser::AST` is not that complex

lol'd when I reread this

backus: ^ Also to clarify on the above comment: I absolutely agree that it is worth it now that I have had to heavily use the mutable `Regexp::Expression` mutable API. I just found it to be way more work than I anticipated

mbj: Compared to the benefit its not. Having to work with mutable / class per type nodes is uglish like hell.

I've been there. Mutant initially was using the RBX AST thats heavily biased to one specific operation: Bytecode emitting. Not a generic AST useful for transformations.

lib/mutant/ast/regexp/transformer/direct.rb
backus: This change actually increases the flay score

lib/mutant/ast/regexp/transformer.rb
```diff
@@ -0,0 +1,76 @@
...
+          def subexpressions
+            node.children.map(&Regexp.public_method(:to_expression))
+          end
+        end
+
+        module LookupTable
+        private
+
+          def expression_token
+            self.class::TABLE.fetch(node.type).first
```
dkubb: Instead of `#first` here, and `#last` below, which are rather weak assertions about the shape of the code, what about doing something like `Quantifier` below and creating a simple class to represent the expression. You can call `#token` and `#regexp_class` on the object or something.

lib/mutant/ast/regexp/transformer/alternative.rb
```diff
@@ -0,0 +1,27 @@
...
+          class ExpressionToAST < Transformer::ExpressionToAST
+            def call
+              ast(*children)
+            end
+          end
+
+          class ASTToExpression < Transformer::ASTToExpression
+            def call
+              root = ALTERNATIVE.dup
```
dkubb: Since this is always acting on `root` and returning the object from this method what about using `Object#tap`? There's a few places in this change where this could be used.

Although I am not sure how @mbj feels about that in mutant code; I don't know if he uses it elsewhere.

lib/mutant/ast/regexp/transformer/direct.rb
mbj: Flay is fine to increase.

mbj: Flay grows logarithmic with the amount of code ideally. I care about flog, as the growth is much bigger compated to the code size.

lib/mutant/ast/regexp/transformer.rb
backus: Done. I extracted a `lookup_table` class method which handles this and also applies the `Token` extraction @mbj mentioned here: https://github.com/mbj/mutant/pull/565#discussion_r61767602

Still have to kill some mutations

lib/mutant/type_registry.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    # Initialize object
+    #
+    # @return [undefined]
+    def initialize
+      super({})
+    end
+
+    # Raised when the type is an invalid type
+    RegistryError = Class.new(TypeError)
```
dkubb: You could probably get by just naming this `Error`, since the name of the parent namespace already includes the term "Registry".

lib/mutant/ast/regexp/transformer/alternative.rb
mbj: I'd also prefer `Object#tab` here. Mostly because we remove the "redundant" re appearance of the `root` token.

```
ALTERNATIVE.dup.tap do |root|
  root.expressions = subexpressions
end
```

Is easier to digest (for me).

mbj: We do not have to "artificially" cause the return value to be correct, I think thats the main part.

mbj: OT: Can we freeze the duplicate?

spec/unit/mutant/ast/regexp_spec.rb
mbj: Once mutant dropped 2.2, we can use the new squiggly heredoc syntax.

lib/mutant/ast/regexp/transformer/alternative.rb
backus: There are some instances where `regexp_parser` needs a mutable object. In those cases I can add a comment saying the object is intentionally not frozen. k?

mbj: Yeah, also this information is the perfect evidence why we map the regexp from `regexp_parser` to our own immutable structure.

lib/mutant/ast/regexp/transformer/direct.rb
```diff
@@ -0,0 +1,99 @@
...
+            def call
+              quantify(ast)
+            end
+          end # ExpressionToAST
+
+          class ASTToExpression < Transformer::ASTToExpression
+            include LookupTable
+
+            # rubocop:disable LineLength
+            TABLE = Table.create(
```
dkubb: This isn't a blocker, but what about moving `TABLE` up to the `Direct` class, and then doing:

```ruby
TABLE.each { |(node_type)| register(node_type) }
```

This would remove the duplication between the table and the list of node types.

lib/mutant/ast/regexp/transformer/quantifier.rb
```diff
@@ -0,0 +1,90 @@
...
+          end # ExpressionToAST
+
+          class ASTToExpression < Transformer::ASTToExpression
+            include NamedChildren
+
+            children :min, :max, :subject
+
+            Quantifier = Class.new.include(Concord::Public.new(:type, :suffix, :mode))
+
+            QUANTIFIER_MAP = IceNine.deep_freeze({
```
dkubb: Instance of https://github.com/mbj/mutant/pull/565/files#r62434817

lib/mutant/ast/regexp/transformer/recursive.rb
```diff
@@ -0,0 +1,49 @@
...
+            def call
+              quantify(ast(*children))
+            end
+          end # ExpressionToAST
+
+          class ASTToExpression < Transformer::ASTToExpression
+            include LookupTable
+
+            # rubocop:disable LineLength
+            TABLE = Table.create(
```
dkubb: Instance of https://github.com/mbj/mutant/pull/565/files#r62434817

lib/mutant/ast/regexp/transformer/text.rb
```diff
@@ -0,0 +1,55 @@
...
+            def call
+              quantify(ast(expression.text))
+            end
+          end # ExpressionToAST
+
+          class ASTToExpression < Transformer::ASTToExpression
+            include LookupTable
+
+            # rubocop:disable LineLength
+            TABLE = Table.create(
```
dkubb: Instance of https://github.com/mbj/mutant/pull/565/files#r62434817

lib/mutant/ast/regexp/transformer/direct.rb
backus: In order to hoist up this constant I wouldn't be able to specify these methods 

```ruby
def expression_token
  self.class::TABLE.lookup(node.type).token
end

def expression_class
  self.class::TABLE.lookup(node.type).regexp_class
end
```

dkubb: True. I suppose doing: `TABLE = Direct::TABLE` would bring it into the right scope. It's ugly, but probably not as ugly as data duplication.

backus: What about doing the `register` below the `ASTToExpression` class? Then I could just do `ASTToExpression::TABLE.each`

backus: I resolved this by moving the registers below the `TABLE` usage

lib/mutant/ast/regexp/transformer/quantifier.rb
backus: Resolved

lib/mutant/ast/regexp/transformer/recursive.rb
backus: Resolved

lib/mutant/ast/regexp/transformer/text.rb
backus: Resolved

lib/mutant/ast/regexp/transformer/options_group.rb
```diff
@@ -0,0 +1,66 @@
...
+              )
+            end
+
+            # Flag text constructed from enabled options
+            #
+            # @return [String]
+            def text
+              flags = options.map { |key, value| key if value }.join
+
+              "(?#{flags}-:"
```
backus: Would be interested in replacing this with

```ruby
"(?#{options.keys.select(&options).join}-:"
```

but that isn't ruby 2.3 compatible

backus:
@dkubb @mbj I've made some final tweaks and I now think this is officially ready for review and hopefully mergeable this week :smile:

backus:
@ammar You mentioned you were away in https://github.com/ammar/regexp_parser/pull/19 so I'm re-pinging. Would you be interested in reviewing this PR? This adds `regexp_parser` to mutant and heavily uses `Regexp::Parser`. I'd love your feedback on my implementation

ammar:
@backus I don't have enough context on the mutant side of things.

As for regexp_parser, the mapping tables seem long and verbose. If you're only targeting a subset of the expressions, then that makes sense. If not, then I think that the the defined token constants can be used to generate the tables. Unless there are different semantics for the tokens in this context.

That's just a quick impression.

dkubb:
> I don't have enough context on the mutant side of things.

Mutant is a mutation testing system. It changes a single node in the AST and then runs the tests to see what fails. If there are no failures then that node is not fully tested. In effect we are testing the tests for completeness.

This specific PR adds the ability for mutant to mutate nodes in the regexp AST, not just the ruby one. As I'm sure you know, regexps can contain lots of assumptions and embedded behaviour in a very dense representation. At best regexs are only partially tested, but this PR will ensure tests cover all the behaviour and logic branches at least once.

> As for regexp_parser, the mapping tables seem long and verbose. If you're only targeting a subset of the expressions, then that makes sense. If not, then

@ammar My assumption, if it's not already true in this PR, is that we will be mutating every kind of regular expression.

> I think that the the defined token constants can be used to generate the tables. Unless there are different semantics for the tokens in this context.

Are you saying that you think *all* the meta-data that would be required to build this table can be inferred from the `regexp_parser` library?

backus:
I might have more work to do after all. @ammar's comments made me consider adding `regexp_parser` as one of the integrations for the corpus tests and that uncovered more edge cases that were not revealed by Rubyspec.

mbj:
@backus Do you expect there is something we can extract from this PR for faster merge?

backus:
@mbj I'd like to figure out how to ship the current code without having to handle mapping for conditional regular expressions. Ideally mutant would behave normally with them it just wouldn't be able to mutate them yet. If we could do that then it would be easier to merge soon

mbj:
> conditional regular expressions.

Whats a conditional regular expression? Should we subset what mutant supports instead, and in case there is no mapping simply do not emit mutations?

backus:
> Whats a conditional regular expression? 

In ruby you can encode a conditional branch in your your regular expression. Here is an annotated and runnable example:

```ruby
pattern = /
  \A(f)?  # optional capture group which requires first letter to be 'f'
  (?(1)   # conditional on whether capture group 1 was successful
    (oo)  # if capture group 1 was successful then next characters should be 'oo'
    |
    (bar) # otherwise the string should be 'bar'
  )
/x

pattern =~ 'foo' # => 0
pattern =~ 'bar' # => 0
pattern =~ 'fbar' # => nil
pattern =~ 'baz' # => nil
```

backus:
> Should we subset what mutant supports instead, and in case there is no mapping simply do not emit mutations?

Yes I think so. I just need to figure out what the bijective mapper does in this situation

backus:
@mbj for a better explanation of conditional regexps: http://blog.honeybadger.io/using-conditionals-inside-ruby-regular-expressions/#conditionals

mbj:
I do not like them, use a real parser for that stuff. I'd be fine to start without them. Just scan the AST for unsupported types and early exit when you see one? You could diff the constants in `regexp_parser` and the table contents in mutants mapper, if there is something mutant cannot convert yet, you do not emit mutations.

backus:
> I do not like them, use a real parser for that stuff. I'd be fine to start without them.

Agreed.

> Just scan the AST for unsupported types and early exit when you see one? You could diff the constants in regexp_parser and the table contents in mutants mapper, if there is something mutant cannot convert yet, you do not emit mutations.

So the steps for mutating a regular expression are:

1. use `Regexp::Parser` to parse regex source into `Regexp::Expression`s
2. convert `Regexp::Expression`s into `Parser::AST::Node`s
3. mutate nodes
4. convert nodes back to expressions
5. use expressions to unparse

So I would probably have to add a new step 2 which would be something like "Check to see if `Regexp::Expression` is supported and don't map to `Parser::AST::Node`s it contains expressions we don't know how to map.

Right now I think this PR supports the important subset of ruby regular expressions. If you agree with the added step I mentioned above then I'm cool with implementing this tonight so that we still have some hope of merging this week

backus:
@mbj @dkubb I filled out the remaining Regexp mapping that adding regexp_parser to the corpus exposed. I also added a simple filter for skipping conditional regular expressions which is the only thing the mapper does not currently support. The build should be passing now so this should be ready for review

lib/mutant/ast/regexp.rb
```diff
@@ -0,0 +1,41 @@
...
+      UNSUPPORTED_EXPRESSION_TYPE = :conditional
+      private_constant(*constants(false))
+
+      # Check if expression is supported by mapper
+      #
+      # @param expression [Regexp::Expression]
+      #
+      # @return [Boolean]
+      def self.supported?(expression)
+        return if expression.type.equal?(UNSUPPORTED_EXPRESSION_TYPE)
```
dkubb: Should this return `false` given that the documented return type is `Boolean`? It may mean adding a new spec, but for public interfaces I think this is worth the effort to have a clean, unambiguous interface.

lib/mutant/ast/regexp/transformer/alternative.rb
```diff
@@ -0,0 +1,37 @@
...
+        class Alternative < self
+          register :regexp_sequence_expression
+
+          # Mapper from `Regexp::Expression` to `Parser::AST::Node`
+          ExpressionToAST = Class.new(Recursive::ExpressionToAST)
+
+          # Mapper from `Parser::AST::Node` to `Regexp::Expression`
+          class ASTToExpression < Transformer::ASTToExpression
+            ALTERNATIVE = IceNine.deep_freeze(
+              ::Regexp::Expression::Alternative.new(0, 0, 0)
```
dkubb: What do these `0` magic numbers actually represent? Maybe define 3 constants for them so it's more clear what their purpose is.

lib/mutant/ast/regexp.rb
backus: This method does always return `false` when the method isn't being called recursively. If the method is being called with an expression from anywhere else then it is likely being called with a `Regexp::Expression::Root` which will never be `UNSUPPORTED_EXPRESSION_TYPE` and will never be `terminal?`

lib/mutant/ast/regexp/transformer/quantifier.rb
```diff
@@ -0,0 +1,112 @@
...
+      class Transformer
+        # Transformer for regexp quantifiers
+        class Quantifier < self
+          # Mapper from `Regexp::Expression` to `Parser::AST::Node`
+          class ExpressionToAST < Transformer::ExpressionToAST
+            # Transform quantifier into node
+            #
+            # @return [Parser::AST::Node]
+            def call
+              ast(expression.min, expression.max)
```
dkubb: Are these being called on a `Range`? If so, note that `#max` is implemented using `Enumberable#max` which could be expensive given a large range. Assuming the Range is inclusive you can just use `#begin` and `#end`. If it's an exclusive Range, assuming it's an `Integer`, you could use `range.end - 1` for the last element.

backus: No these are attributes belonging to `Regexp::Expression::Quantifier`

backus: https://github.com/ammar/regexp_parser/blob/master/lib/regexp_parser/expression/quantifier.rb

lib/mutant/ast/regexp.rb
dkubb: What about changing it to something like:

```ruby
def self.supported?(expression)
  return true if expression.terminal?
  expression.all? do |subexp|
    !subexp.type.equal?(UNSUPPORTED_EXPRESSION_TYPE) &&
    public_send(__method__, subexp)
  end
end
```

This way it's more clear what is happening, and the method is [total](https://en.wikipedia.org/wiki/Partial_function#Total_function).

lib/mutant/ast/regexp/transformer/alternative.rb
backus: Three numbers that are usually set as a result of parsing a regular expression that don't seem to actually matter for unparsing. I can extract constants or alternatively I could just leave a comment

https://github.com/ammar/regexp_parser/blob/master/lib/regexp_parser/expression/sequence.rb#L10

dkubb: >  I can extract constants or alternatively I could just leave a comment

I am fine with a comment.

lib/mutant/ast/regexp.rb
backus: I might prefer specifying the recursive call site in a test as well. I'll have to play with this rewrite but I think it behaves a bit differently. Writing `return false` for both guards seems more clear to me though

backus:
@dkubb I addressed your two remaining flags so this should be ready for another pass /cc @mbj

dkubb:
@backus I have no further flags.

@mbj do you mind reviewing it and making sure you're ok to maintain this code going forward? If you are then I think this is ready to merge.

andyw8:
@dkubb @backus you may already know, but this PR discussion appears in Chapter 2 of [Working in Public](https://www.amazon.com/Working-Public-Making-Maintenance-Software/dp/0578675862).

Screen Shot 2020-09-24 at 9 34 08 PM

mbj:
@andyw8 I was not aware. Thanks for letting us know. Too bad the book is now outdated as mutant is not opensoruce anymore, Also during the license transition I had to remove this regexp mutation testing feature.

mbj:
On topic: It appears I can bring the regexp mutations back, there was a misunderstanding on my side.

dgollahon:
@mbj If you would like help reintroducing those I would be happy to help.

mbj:
@dgollahon I think a revert of my original removal would likely apply without major edits. Feel free to try it.

All my mutant time (well most of it) is currently going into unparser 2.7. I decided to stop mangling and go for full AST equivalence. This makes the emitted code more beautiful (and correct in many cases). In theory I should be done soon.

PR #566, opened 2016-04-24
Add "expected_errors" to corpus integrations

backus:
This change drops the concept of `excludes` from the corpus integration configuration. Instead with `permitted_errors` you specify pairs of paths and expected errors.

---

I extracted this PR from #565 which I think requires a better system for specifying what exceptions we are allowed within the corpus tests. I replaced the `exclude` configuration with `permitted_errors` because I think this is a more specific and strict guarantee for mutant's corpus tests.

---

The second commit in this PR replaces the concept of `permitted_errors` with `expected_errors`. With this change we also assert that we did not expect an error for every corpus file which did encounter an error during mutation generation.

### DIFF
--- a/spec/integrations.yml
+++ b/spec/integrations.yml
@@ -5,36 +5,100 @@
   mutation_coverage: false
   mutation_generation: true
   expect_coverage: 0 # not run
-  exclude:
-    # Binary encoded source subjected to limitations see README of unparser
-    - core/array/pack/{b,h,u}_spec.rb
-    - language/regexp/escapes_spec.rb
-    - core/array/pack/shared/float.rb
-    - core/array/pack/shared/integer.rb
-    - core/array/pack/{c,m,w}_spec.rb
-    - core/regexp/shared/quote.rb
-    - core/encoding/compatible_spec.rb
-    - core/io/readpartial_spec.rb
-    - core/env/element_reference_spec.rb
-    - core/dir/pwd_spec.rb
-    - core/string/casecmp_spec.rb
-    - core/string/unpack/{b,c,h,m,u,w}_spec.rb
-    - core/string/unpack/b_spec.rb
-    - core/string/unpack/shared/float.rb
-    - core/string/unpack/shared/integer.rb
-    - core/symbol/casecmp_spec.rb
-    - optional/capi/integer_spec.rb
+  expected_errors:
+    "#<Parser::SyntaxError: invalid multibyte escape: /\xAA/>":
+      - language/regexp/escapes_spec.rb
+    "#<Parser::SyntaxError: literal contains escape sequences incompatible with UTF-8>":
+      - core/array/fixtures/encoded_strings.rb
+      - core/array/pack/shared/string.rb
+      - core/array/pack/shared/unicode.rb
+      - core/encoding/converter/convert_spec.rb
+      - core/encoding/converter/last_error_spec.rb
+      - core/encoding/converter/primitive_convert_spec.rb
+      - core/encoding/converter/primitive_errinfo_spec.rb
+      - core/encoding/converter/putback_spec.rb
+      - core/encoding/fixtures/classes.rb
+      - core/encoding/invalid_byte_sequence_error/error_bytes_spec.rb
+      - core/encoding/invalid_byte_sequence_error/incomplete_input_spec.rb
+      - core/encoding/invalid_byte_sequence_error/readagain_bytes_spec.rb
+      - core/encoding/replicate_spec.rb
+      - core/file/expand_path_spec.rb
+      - core/io/gets_spec.rb
+      - core/io/read_spec.rb
+      - core/io/shared/gets_ascii.rb
+      - core/io/write_spec.rb
+      - core/marshal/dump_spec.rb
+      - core/marshal/fixtures/marshal_data.rb
+      - core/marshal/shared/load.rb
+      - core/random/bytes_spec.rb
+      - core/regexp/match_spec.rb
+      - core/regexp/shared/new_ascii.rb
+      - core/string/byteslice_spec.rb
+      - core/string/codepoints_spec.rb
+      - core/string/comparison_spec.rb
+      - core/string/count_spec.rb
+      - core/string/delete_spec.rb
+      - core/string/element_set_spec.rb
+      - core/string/encode_spec.rb
+      - core/string/encoding_spec.rb
+      - core/string/fixtures/iso-8859-9-encoding.rb
+      - core/string/getbyte_spec.rb
+      - core/string/gsub_spec.rb
+      - core/string/new_spec.rb
+      - core/string/scrub_spec.rb
+      - core/string/shared/chars.rb
+      - core/string/shared/codepoints.rb
+      - core/string/shared/each_codepoint_without_block.rb
+      - core/string/shared/encode.rb
+      - core/string/shared/eql.rb
+      - core/string/shared/succ.rb
+      - core/string/slice_spec.rb
+      - core/string/squeeze_spec.rb
+      - core/string/unicode_normalize_spec.rb
+      - core/string/valid_encoding_spec.rb
+      - core/symbol/casecmp_spec.rb
+      - core/time/_dump_spec.rb
+      - core/time/_load_spec.rb
+      - language/regexp/encoding_spec.rb
+      - language/string_spec.rb
+      - library/digest/md5/shared/constants.rb
+      - library/digest/md5/shared/sample.rb
+      - library/digest/sha1/shared/constants.rb
+      - library/digest/sha256/shared/constants.rb
+      - library/digest/sha384/shared/constants.rb
+      - library/digest/sha512/shared/constants.rb
+      - library/openssl/shared/constants.rb
+      - library/socket/basicsocket/recv_spec.rb
+      - library/socket/socket/gethostbyname_spec.rb
+      - library/stringscanner/getch_spec.rb
+      - library/stringscanner/shared/get_byte.rb
+      - library/zlib/deflate/deflate_spec.rb
+      - library/zlib/deflate/set_dictionary_spec.rb
+      - library/zlib/gzipreader/each_byte_spec.rb
+      - library/zlib/gzipreader/eof_spec.rb
+      - library/zlib/gzipreader/getc_spec.rb
+      - library/zlib/gzipreader/pos_spec.rb
+      - library/zlib/gzipreader/read_spec.rb
+      - library/zlib/gzipreader/rewind_spec.rb
+      - library/zlib/inflate/append_spec.rb
+      - library/zlib/inflate/inflate_spec.rb
+      - library/zlib/inflate/set_dictionary_spec.rb
+      - library/zlib/zstream/flush_next_out_spec.rb
+      - optional/capi/encoding_spec.rb
+      - optional/capi/string_spec.rb
+    '#<RegexpError: invalid multibyte escape: /\xAA/>':
+      - language/regexp/escapes_spec.rb
 - name: auom
   namespace: AUOM
   repo_uri: 'https://github.com/mbj/auom.git'
   mutation_coverage: true
   mutation_generation: true
-  exclude: []
+  expected_errors: {}
   expect_coverage: 1
 - name: axiom
   namespace: Axiom
   repo_uri: 'https://github.com/dkubb/axiom.git'
   mutation_coverage: false
   mutation_generation: true
-  exclude: []
+  expected_errors: {}
   expect_coverage: 1
--- a/spec/support/corpus.rb
+++ b/spec/support/corpus.rb
@@ -29,9 +29,11 @@ module MutantSpec
       START_MESSAGE               = 'Starting - %s'.freeze
       FINISH_MESSAGE              = 'Mutations - %4i - %s'.freeze
 
+      DEFAULT_MUTATION_COUNT = 0
+
       include Adamantium, Anima.new(
-        :exclude,
         :expect_coverage,
+        :expected_errors,
         :mutation_coverage,
         :mutation_generation,
         :name,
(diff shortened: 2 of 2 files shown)


backus:
@mbj this is ready for review

mbj:
@backus Conceptually +1. This is what I imagined when I was elaborating on "whitelisting errors". Will do a code pass later the day.

spec/support/corpus.rb
```diff
@@ -124,6 +116,52 @@ def checkout
...
+      end
+
+      # Yield and conditionally suppress errors according to whitelist
+      #
+      # @param path [Pathname] path responsible for exception
+      # @param default_return [Object] return value if exception is permitted
+      #
+      # @return value returned from yield if no exception raised
+      # @return `default_return` otherwise
+      def filter_exceptions(path, default_return)
```
mbj: This whitelists exceptions. But I wounder if it should instead "expect" exceptions, so we notice when the expected exception does not get raised anymore so we can reduce our whitelist.

The prior implementation was affected by "fixed without our notice" also, we should look into improving that.

spec/support/corpus.rb
```diff
@@ -80,21 +82,11 @@ def verify_mutation_generation
...
-              nil # Make rubocop happy
-            end
-
-          if node
-            count += Mutant::Mutator::REGISTRY.call(node).length
-          end
 
-          count
+        total = Parallel.map(effective_ruby_paths, options) do |path|
+          filter_exceptions(path, DEFAULT_MUTATION_COUNT) { count_mutations(path) }
```
mbj: As this is the only call side for `filter_exceptions` passing in the 2nd argument from a constant is redundant.

mbj: I notice your intention is to make the `filter_exceptions` method generic, but as long we do not have a 2nd call side, lets simply go for `count_mutations(path)` here that contains the semantics for exception filtering expectation (maybe through another extracted method, but this one should know the defaults by itself).

spec/support/corpus.rb
```diff
@@ -29,9 +29,11 @@ class Project
       START_MESSAGE               = 'Starting - %s'.freeze
       FINISH_MESSAGE              = 'Mutations - %4i - %s'.freeze
 
+      DEFAULT_MUTATION_COUNT = 0
```
mbj: I came to the conclusion we should freeze integer literals also, to prevent accidental modification of their singleton class.

backus: What would that even look like?

mbj: Some function somebody passes an anonymous value in to add some debuging or tracing mecnahism to *all* objects passing by.

You could even violate basic math:

> What would that even look like?

Actually recent ruby forbids opening the singleton class for `Fixums` nice.

spec/support/corpus.rb
backus: I actually planned on implementing this but didn't because I wasn't sure if there might be different exceptions under different ruby versions since that determines which version of `Parser` is used.

mbj: Mutant should only ever work on the last parser version. So in case we see problems from multiple versions lets just tighten the version constraint before silencing too much corpus.

backus: That is what I assumed but I didn't know definitively. That is helpful.

I expect to push "expected errors" in the next hour or so

mbj: Mutant is in a hard domain (mostly because it works on ruby a known to be unsharp langauge), keeping the tool sharp for my use cases and the maintenance overhead low is the main priority.
I do not run any tool that requires legacy parser versions.

backus:
@mbj I've updated the PR in response to your feedback. New changes are in the second commit. I also extended the PR description to describe briefly what changed

spec/support/corpus.rb
```diff
@@ -124,6 +116,40 @@ def checkout
...
+      #
+      # @param path [Pathname] path responsible for exception
+      #
+      # @return [Fixnum] mutations generated
+      def count_mutations_and_check_errors(path)
+        relative_path = path.relative_path_from(repo_path)
+
+        count = count_mutations(path)
+
+        expected_errors.assert_success(relative_path)
```
dkubb: It looks like `count_mutations` is a pure function, so does it matter if this assertion is done before or after it? I generally like to put assertions as early as possible in the chain, in order to safe unnecessary work.

Also, it'll allow you to return `count_mutations(path)` instead of creating a synthetic variable to return.

spec/support/corpus.rb
```diff
@@ -226,6 +243,53 @@ def system(arguments)
         end
       end
 
+      # Mapping of files which we expect to cause errors during mutation generation
+      class ErrorWhitelist
+        UnnecessaryExpectation = Class.new(StandardError)
```
dkubb: Small point, but I would generally align similar operators when they appear on successive lines.

spec/support/corpus.rb
```diff
@@ -226,6 +243,53 @@ def system(arguments)
...
+        UNNECESSARY_EXPECTATION = 'Expected to encounter %s while mutating "%s"'.freeze
```
dkubb: Instead of defining this as a constant and having it loosely paired with the `UnnecessaryExpectation` constant, what about defining an actual `UnnecessaryExpectation` class where this is the default message? Then when you raise the exception below you could use something like:

```ruby
fail UnnecessaryExpectation.new(map, path)
```

spec/support/corpus.rb
```diff
@@ -242,13 +306,24 @@ def system(arguments)
...
                 s(:key_symbolize, :mutation_generation,
                   s(:guard, s(:or, s(:primitive, TrueClass), s(:primitive, FalseClass)))),
-                s(:key_symbolize, :exclude,             s(:map, s(:guard, s(:primitive, String))))
+                s(:key_symbolize, :expected_errors,
+                  s(:block,
+                    s(:guard, s(:primitive, Hash)),
+                    s(:custom,
+                      [
+                        ->(hash) { hash.map { |key, value| [Pathname.new(key), value] }.to_h  },
+                        ->(hash) { hash.map { |key, value| [key.to_s,          value] }.to_h  }
```
dkubb: There's an extra space before the closing `}` here and on the line above.

spec/support/corpus.rb
```diff
@@ -80,21 +82,11 @@ def verify_mutation_generation
...
-            end
-
-          if node
-            count += Mutant::Mutator::REGISTRY.call(node).length
-          end
 
-          count
+        total = Parallel.map(effective_ruby_paths, options) do |path|
+          count_mutations_and_check_errors(path)
         end.inject(0, :+)
```
dkubb: Should this `0` be replaced with `DEFAULT_MUTATION_COUNT`?

spec/support/corpus.rb
backus: > does it matter if this assertion is done before or after it

Yes. `assert_success` is only run if an exception isn't raised.

dkubb: > assert_success is only run if an exception isn't raised.

Ahh ok, it wasn't clear to me that `#count_mutations` could raise an exception. Maybe it should be documented using `@raise` ?

backus: Ah good point I missed that

spec/support/corpus.rb
backus: Yup

spec/support/corpus.rb
backus: Nice catch. I want to PR something to rubocop eventually that flags `ExtraSpacing` for multiple lines which are aligned

spec/support/corpus.rb
backus: @dkubb I added `@raise` to [line 144](https://github.com/mbj/mutant/pull/566/files#diff-a6b9f047fe088f086dce98b7ec90f402R144). It is pretty generic because the method generically raises an error from either `parser` or `regexp_parser` but it conveys this

backus:
@dkubb @mbj I've updated this PR in response to your feedback and it should be ready for review again

backus:
I've changed up the format for how expected errors are specified. For some reason the release of parser 2.3.1.0 led to way more UTF-8 related exceptions in the corpus spec.

spec/support/corpus.rb
```diff
@@ -226,6 +245,60 @@ def system(arguments)
         end
       end
 
+      # Mapping of files which we expect to cause errors during mutation generation
+      class ErrorWhitelist
+        class UnnecessaryExpectation < StandardError
+          MESSAGE = 'Expected to encounter %s while mutating "%s"'.freeze
+
+          def initialize(error, path)
```
dkubb: If you wanted to you could use `*error_info` here, and then do `super(MESSAGE % error_info)`. You're not really using any of the values on their own, and really just proxying them to `#%`.

spec/support/corpus.rb
```diff
@@ -226,6 +245,60 @@ def system(arguments)
...
+        # @raise provided exception if we are not expecting this error
+        #
+        # This method is reraising exceptions but rubocop can't tell
+        # rubocop:disable Style/SignalException
+        #
+        # @return [undefined]
+        def assert_error(path, exception)
+          original_error = exception.cause || exception
+
+          raise exception unless map.fetch(original_error.inspect, []).include?(path)
```
dkubb: You could probably use `EMPTY_ARRAY` in place of `[]`.

spec/integrations.yml
```diff
@@ -5,36 +5,100 @@
...
-    - core/io/readpartial_spec.rb
-    - core/env/element_reference_spec.rb
-    - core/dir/pwd_spec.rb
-    - core/string/casecmp_spec.rb
-    - core/string/unpack/{b,c,h,m,u,w}_spec.rb
-    - core/string/unpack/b_spec.rb
-    - core/string/unpack/shared/float.rb
-    - core/string/unpack/shared/integer.rb
+    "#<Parser::SyntaxError: literal contains escape sequences incompatible with UTF-8>":
+    - core/array/fixtures/encoded_strings.rb
```
dkubb: So these paths are related to the error string above? It wasn't clear to me that they were related. Before looking closely I thought this was just a list of expected errors.

Would it be syntactically valid to indent each list underneath the error they relate to? I think that might show the relationship better.

dkubb:
@backus I left a couple of comments in the review. Overall it's looking really good, most of the things I flagged are just tiny points.

spec/support/corpus.rb
backus: Hm I would have to use `Mutant::EMPTY_ARRAY` since this is `MutantSpec`. I'm not sure if that is worth it

spec/integrations.yml
backus: I can indent them another two spaces. Is that what you're talking about here?

backus:
@dkubb I made most of the changes you mentioned. I left a comment in the one case I didn't make a change. Ready for another pass

spec/support/corpus.rb
dkubb: Oh heh, I was just looking at this code as if it were part of `Mutant::`. Yeah, I agree with leaving it as-is for now. If `MutantSpec::EMPTY_ARRAY` is ever created then I'd change it, but for now it's cool.

dkubb:
> I've changed up the format for how expected errors are specified. For some reason the release of parser 2.3.1.0 led to way more UTF-8 related exceptions in the corpus spec.

Could this be related to this change: https://github.com/whitequark/parser/commit/95401a20e8f4532e32f6361da3918ac8e4bd18c7

spec/support/corpus.rb
```diff
@@ -80,21 +82,11 @@ def verify_mutation_generation
           start:        method(:start),
           in_processes: parallel_processes
         }
+
         total = Parallel.map(effective_ruby_paths, options) do |path|
```
dkubb: Could this be written as:

```ruby
total = Parallel.map(effective_ruby_paths, options, &method(:count_mutations_and_check_errors))
  .inject(DEFAULT_MUTATION_COUNT, :+)
```

spec/support/corpus.rb
```diff
@@ -124,6 +116,42 @@ def checkout
 
     private
 
+      MissingExpectedException = Class.new(StandardError)
```
dkubb: Does defining this here actually make it private? I don't think it has any effect. I would tend to define constants near the top of the namespace, just to give a consistent layout, and to not provide a false signal that it is private.

Also I did a quick search and I don't see this constant used anywhere in this diff. Is it used elsewhere?

spec/support/corpus.rb
```diff
@@ -226,6 +245,60 @@ def system(arguments)
...
+        # Assert that we expect to not encounter an error for the specified path
+        #
+        # @param path [Pathname]
+        #
+        # @raise [UnnecessaryExpectation] if we are expecting an exception for this path
+        #
+        # @return [undefined]
+        def assert_success(path)
+          map.each do |error, paths|
+            next unless paths.include?(path)
```
dkubb: I tend to only use guard clauses if the code I am guarding is more than one line of code. In this case you could probably write this as a one-liner:

```ruby
map.each do |error, paths|
  fail UnnecessaryExpectation.new(error, path) if paths.include?(path)
end
```

spec/support/corpus.rb
backus: Woops I think this might be a holdover actually from the initial implementation. Now I just re-raise

spec/support/corpus.rb
backus: Yeah I could rewrite that way. 

Regarding reduce: I'm just following the rubocop preferences for mutant. If @mbj prefers reduce I can update the config accordingly and make the change

dkubb:
@backus this is looking really good. The only things I could find were really tiny things that probably have no significant bearing on functionality; I only flagged them because we're on the final polishing step.

backus:
> @backus this is looking really good. The only things I could find were really tiny things that probably have no significant bearing on functionality; I only flagged them because we're on the final polishing step.

Cool. I can make some final tweaks now

spec/support/corpus.rb
backus: Yeah I like that

backus:
> Could this be related to this change: whitequark/parser@95401a2

Almost definitely. Good catch

PR #596, opened 2016-05-25
Fix regexp_parser syntax for ruby 2.3.1

backus:
I'm interested in adding to regexp_parser for a more robust solution. See https://github.com/ammar/regexp_parser/issues/23

### DIFF
--- a/lib/mutant/ast/regexp.rb
+++ b/lib/mutant/ast/regexp.rb
@@ -3,7 +3,20 @@ module Mutant
     # Regexp source mapper
     module Regexp
       UNSUPPORTED_EXPRESSION_TYPE = :conditional
+
       private_constant(*constants(false))
+
+      # Parse regex string into expression
+      #
+      # @param regexp [String]
+      #
+      # @return [Regexp::Expression]
+      def self.parse(regexp)
+        ::Regexp::Parser.parse(
+          regexp,
+          "ruby/#{RUBY_VERSION.split('.').first(2).join('.')}"
+        )
+      end
 
       # Check if expression is supported by mapper
       #
--- a/Gemfile
+++ b/Gemfile
@@ -3,3 +3,5 @@ source 'https://rubygems.org'
 gemspec name: 'mutant'
 
 eval_gemfile File.expand_path('../Gemfile.shared', __FILE__)
+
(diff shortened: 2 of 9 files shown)


backus:
@mbj this should be ready for review.

@mvz would you mind checking to confirm this fixes your issue?

mvz:
@backus Yes, this fixes my issue. Thanks!

lib/mutant/ast/regexp.rb
```diff
@@ -3,8 +3,18 @@ module AST
...
+
       private_constant(*constants(false))
 
+      # Parse regex string into expression
+      #
+      # @param regexp [String]
+      #
+      # @return [Regexp::Expression]
+      def self.parse(regexp)
+        ::Regexp::Parser.parse(regexp, "ruby/#{RUBY_VERSION.to_f}")
```
mbj: `RUBY_VERSION.to_f`?

backus: Truncates '2.3.1' to 2.3

mbj: This is not a good idea. You rely on transitive property `String -> Float -> OtherString`. Lets do something more explicit, splitting the string and joining together the first 2 components.

Using type coercions that only by "accident" produce a value mutation (effectively you mutate a string) is error prone.

backus:

I've locked the `regexp_parser` version to 0.3.3. The 0.3.4 release added the syntax file for Ruby 2.3.1 but also introduced a regression which means we would have to wait to merge until 0.3.5 (see ammar/regexp_parser#24). This isn't relevant to this PR though since this is a *generic* fix which will always try to use `"ruby/MAJOR.MINOR"` with `Regexp::Parser` which should help us avoid this problem in the future.

backus:
Hm ok it looks like the corpus is now failing anyways even with the lock. I'll investigate later but we may just have to wait for 0.3.5

mbj:
@backus With that part of my reason for #599 becomes invalid. Still we have to deal with the megabytes of warnings generated running a regexp parser enabled mutant against the corpus.

Is there any chance to get these warnings fixed upstream?

backus:
Closing in favor of #618

PR #599, opened 2016-05-29
Revert "Mutate regexp body"

mbj:
Reverts mbj/mutant#565


mbj:
I wounder if we should revert the regexp parser integration temporarily till the following problems are solved:

* Regexp escapes that fail the corpus
* Warnings that make the build output not efficiently readable.

mbj:
@dkubb, @backus ^^ Given we have as much warnings from regexp_parser and CircleCI has to truncate the output (so debugging against CI output is hard), I consider master to be broken.

And I really *dislike* master is not releasable as is.

I do not want to put pressure on anyone espeially @backus, but `regexp_parser` is not as stable as we thought and I think we should continue a bit in a branch before moving on.

backus:
> Regexp escapes that fail the corpus

So regarding regexp escapes failing the corpus that should be fixed in the 0.3.5 release which I anticipate @ammar will probably release in a day or two (he already merged the fix ammar/regexp_parser@26e7c14).

> Warnings that make the build output not efficiently readable.

I reduced the worst of the warnings in a PR that landed in 0.3.4. The build on #596 was just especially bad due to me locking it back at 0.3.3 because I thought it would let us work around the regression. There are still warnings but it no longer floods the output (running `bundle exec rspec spec/unit` produces 68 warnings).

> And I really *dislike* master is not releasable as is.

Yeah I'm sorry about that. This surprised me.

---

So regarding doing a revert I am of two minds here.

1. I understand not wanting master to fail and block other PRs. I am fine with reverting to avoid this. Personally though I am also nervous about returning to an indefinite feeling about not knowing if the regexp PR will ever be merged again.

2. My personal preference would be to see if @ammar releases 0.3.5 before Tuesday and, if he does, then we fix master by upgrading and if there are still issues we can revert. I understand how waiting would be unpleasant though.

Therefore, if you do want to revert, my request would be that we communicate clearly before merging the revert about the conditions where my regexp work gets merged back into master. I think that should be a passing build using 0.3.5 with added tests for the regressions that caused this issue. I do not think that eliminating all warnings though should be a blocker. I would be fine with saying "lets resolve these warnings before we merge new regexp related features" though.

mbj:
> Yeah I'm sorry about that. This surprised me.

I did not check the CI build with great detail, and actually that warnings should have failed the build the first place. So its not at all your fault.

> conditions where my regexp work gets merged back into master. 

No regressions. Added warnings are a regression, failed corpus is a regression.

mbj:
So I consider to merge this with the following roadmap:

1. Merge this PR
2. Make new warnings fail the build
3. Merge the regexp PR with warnings and corpus fixed.

mbj:
There is one long standing warning generated by ragels ruby emitter. I'd accept this one as it only occurs once, but not in the mutant main loop.

mbj:
> My personal preference would be to see if @ammar releases 0.3.5 before Tuesday and, if he does, then we fix master by upgrading and if there are still issues we can revert. I understand how waiting would be unpleasant though.

I expanded my thoughts on this in the channel, but let me sum it up here:

* The current master is unreleasable. And this is *my* fault. I should have checked the consequences more carefully
* Adding anything (maybe even not related to regexps at all, for example documentation fixes etc) while the master is unreleasable is stupid and will not happen.
* Bringing master back to a releasable state is top priority
* The revert takes least amount of time to get master back into releasable state
* PRing the regexp integration once the known problems are fixed is easy.

Hence I decide for the revert.

mbj:
> Personally though I am also nervous about returning to an indefinite feeling about not knowing if the regexp PR will ever be merged again.

Why wouldn't they, I just stated in channel I'll take the work to get regexp mutations fixed in case you do not have the time for it.

mbj:
Related: https://github.com/ammar/regexp_parser/issues/25

PR #605, opened 2016-06-14
Array access mutations

lotyrin:
In particular, to avoid sampling (introducing nondeterminism) when first/last will do.

`sample` -> `first`
`sample` -> `last`
`first` -> `last`

Edit, now also mutates:

`pop` -> `last`
`shift` -> `first`

To point out unnecessary array modifications, as well as mutating

`last` -> `first`


lib/mutant/mutator/node/send.rb
```diff
@@ -18,6 +18,8 @@ class Send < self
           reverse_each:  %i[each],
           reverse_merge: %i[merge],
           map:           %i[each],
+          sample:        %i[first last],
+          first:         %i[last],
```
mbj: Lets also add `last` -> `first`

mbj:
@lotyrin Thx for that, I added a small suggestion that would fit very well into your PR.

tjchambers:
This is getting nasty. I can see lots of code that will now not pass... Thanks(?) for adding this!

tjchambers:
In that spirit what about `shift` to `first`? Then someone would have to prove the need to have the rest of the array altered. Thoughts? `pop` to `last`?

mbj:
@tjchambers If you have code that can transformed from `Array#first` to `Array#last` without your tests noticing, your array under test can have only one or zero elements. Which does not enumerate all the cases you need to cover. You at least need an array of minimum size 2 for having `Array#first` differ to `Array#last`.

tjchambers:
@mbj You hit precisely why that worries me re: mutations.

mbj:
> In that spirit what about shift to first? Then someone would have to prove the need to have the rest of the array altered. Thoughts? pop to last?

Both are good. `shift` is access + modification, transform to `first` is just access. So a semantic reduction, a valid operator.

mbj:
> @mbj You hit precisely why that worries me re: mutations.

Ohh, I thought you where telling above the new mutations would cause invalid alives.

tjchambers:
@mbj No - they would be _valid_ just more complicated to kill.

mbj:
> @mbj No - they would be valid just more complicated to kill.

I do not see extra complication.

And even if there where extra complication: The benefit of forcing arrays of length > 1 would outweight it.

If your code only works under test on edge case array length (0, 1), something is significantly wrong.

tjchambers:
@mbj - sometimes - perhaps incorrectly - I use `.first` to deconstruct an array.

mbj:
And maybe array was the wrong datastructure in the first place.

tjchambers:
@mbj You are inferring it is an array, but sometimes it is a n AR relation.

backus:
How might you kill `sample` → `last` reliably when dealing with a small array @mbj? For example, consider this snippet from your own project: https://github.com/mbj/axiom-fuzzer/blob/master/lib/axiom/fuzzer/operation.rb#L37-L42

mbj:
@backus inject the RNG

mbj:
@backus So essentially avoid "global state controlled" RNG in `Array#shuffle` or a message expectation to "prove" to the mutation testing engine "I really wanted that".

backus:
@mbj could you provide a code sample? It seemed like the only way to make `sample` deterministic was to use its RNG argument.

mbj:
> @mbj could you provide a code sample? It seemed like the only way to make sample deterministic was to use its RNG argument.


Worst case escape hatch, not recommended:
```
expect(some_array).to receive_messages(sample: a_result)
```

mbj:
That `axiom-fuzzer` code would needed to be refactored to inject the RNG ideally, so you can reproduce specific fuzzing sequences. Nice backpressure for a more decent design.

backus:
> > @mbj could you provide a code sample? It seemed like the only way to make sample deterministic was to use its RNG argument.
> 
> 
> Worst case escape hatch, not recommended:
> ```
> expect(some_array).to receive_messages(sample: a_result)
> ```

Alright but are you then saying that this mutation would only be killable (when I do want randomness) with a message expectation? That seems awkward. Is there a tool or set of tools in the ruby standard library that I can turn to that lets me produce a `#sample` replacement?

mbj:
> Alright but are you then saying that this mutation would only be killable (when I do want randomness) with a message expectation? That seems awkward. Is there a tool or set of tools in the ruby standard library that I can turn to that lets me produce a #sample replacement?

When you do want randomness, specify you want it. Randomness has no place to cause an effect in a unit test executation anyway. It should be mocked out always, and ideally injected.

If you want randomness in a test, go for property based testing (that also mixes well with mutation testing, but thats another topic).

Canonically go with:

`array.fetch(rng.rand(array.length))` where you inject the RNG. is the ideal choice. You can than inject a test RNG with predictable results.

mbj:
And `Array#sample` can even natively take an RNG, see: http://docs.ruby-lang.org/en/2.0.0/Array.html#method-i-sample

mbj:
So the best way to deal with it is `array.sample(random: rng_controlled_by_test)`

backus:
> > Alright but are you then saying that this mutation would only be killable (when I do want randomness) with a message expectation? That seems awkward. Is there a tool or set of tools in the ruby standard library that I can turn to that lets me produce a #sample replacement?
>
> When you do want randomness, specify you want it. Randomness has no place to cause an effect in a unit test executation anyway. It should be mocked out always, and ideally injected.
>
> If you want randomness in a test, go for property based testing (that also mixes well with mutation testing, but thats another topic).
>
> Canonically go with:
>
> `array.fetch(rng.rand(array.length))` where you inject the RNG. is the ideal choice. You can than inject a test RNG with predictable results.

Alright. I think that is generally fine if the combined effect of this PR is 1) discouraging unneeded indeterminism and 2) encouraging dependency injection when indeterminism is necessary.

With that said, I do think that there should be some FAQ or section in the README that covers particular mutations where the only way to kill that mutation is to write your code in a significantly different way. I'm not saying the list should be exhaustive or that it is your duty to maintain it (I am not saying that to be clear) but it might be good to centralize example mutation kills like this. A good way to start would be to have contributors include a short explanation (perhaps in a `doc/` directory [like reek does](https://github.com/troessner/reek/tree/master/docs)).

Having a resource like this would be a big value add for mutant I think.

mbj:
> Having a resource like this would be a big value add for mutant I think.

Lets add it to the meta. I always wanted to add "Kill hints" there.

backus:
> Lets add it to the meta. I always wanted to add "Kill hints" there.

Interesting. How thorough are you imagining those hints would be? Ideally, for example, this sample mutation would mention:

1. Do you need randomness or if you can actually just use `first` or `last`? Kill this mutation by changing the source
2. Do you want randomness that you can't prove in your tests due to `#sample()` relying on a global RNG? Kill this mutation by injecting an RNG into the subject under test
3. Do you need to depend on the global RNG or can you not refactor for some other reason? Kill this mutation using a message expectation

backus:
Not saying it might be formatted like I formatted it above, just pointing out that, ideally, mutant would provide some explanation of its opinions like I noted above

mbj:
First we'd need a way to map the meta to the effective mutations, than we can attach anything we want.

mbj:
I've to drop out of the discussion now, heads down in something else, bbiab.

lib/mutant/mutator/node/send.rb
lotyrin: I considered it, but in the case where one actually doesn't care which item is selected out of a list, it seemed like having a stable state would be nice. Based on the conversation I incited, I think you guys have thought about this significantly more than I have though. I'd definitely add last -> first and shift to first, pop to last, if it were to help the PR along.

mbj:
@lotyrin thx.

PR #606, opened 2016-06-18
Fix incorrect, inconsistent, and missing documentation

backus:


mbj:
@backus Thx, as usual.

PR #613, opened 2016-07-03
Add support for rspec 3.5

backus:


backus:
@mbj this should be ready for review assuming that the builds pass. I assume that you will want to drop support for 3.3 if mutant is going to support 3.5. If you would like to keep support for 3.3 then I can just drop the second commit

mbj:
@backus No its fine to drop 3.3.

PR #617, opened 2016-07-24
Fix RuboCop and Reek violations

backus:
Introduced by new devtools release


backus:
@mbj ready assuming build passes

PR #618, opened 2016-07-24
Reintroduce Regexp mutator

backus:
* See #565 for initial code review discussion
* See #599 for the revert discussion


backus:
Includes fix from #596

backus:
@mbj CI is nice and warning free :)

lib/mutant/mutator/node/literal/regex.rb
```diff
@@ -31,6 +32,45 @@ def dispatch
             emit_type(s(:str, NULL_REGEXP_SOURCE), options)
           end
 
+          # Mutate regexp body
+          #
+          # @note will only mutate parts of regexp body if the
+          # body is composed of only strings. Regular expressions
+          # with interpolation are skipped
```
mbj: I think mutant could recurse into the interpolation with ease, and these would also be very interesting, as more likely to reside inside subjects that actually get mutated.

And I'm fine to do this in the next iteration.

backus: Yeah I agree it should be an easy addition

backus:
@mbj build is green now

mbj:
@backus I'm ready to merge after rebase.

backus:
@mbj rebased

Issue #392, opened 2015-08-04
Infinite test run w/ a group of sequences


dgollahon:
In a gem i released called identified I have a file where I create an array with a very specific sequence of numbers using the line below:

```ruby
[*(1..9).step(2), *(10..98).step(2), *(2..8).step(2), *(11..99).step(2)]
```

Mutant hangs and never shows any useful output when the method is covered by the mutant selector. It's currently just ignored in my mutant.yml. The context of the source can be found here: https://github.com/dgollahon/identified/blob/master/lib/identified/ssn/sequential_group_number.rb

@mbj, this is what i showed you on the hangout w/ @backus.

Would be interested to know why it fails.

mbj:
Thx for the report. I'll look into this the next OSS cycle.

mbj:
I did not forgot the issue, but I did not had a pure OSS cycle for month. Just a heads up I still plan to look into this reporoduction.

PR #622, opened 2016-07-25
Fix range mutations

dgollahon:
regarding #392 in the mutation-testing slack:

> So right now I think we should do the following:
* Remove mutations to infinite on ranges
* Lift visiblity for infinite runtime mutations in the readme

> Since I just recalled another fact: applying the N+1 mutation infinite times gives you infinite.

> @dgollahon: Feel free to submit your first PR to remove the range to infinite mutations, citing this discussion.

I don't think we discussed the `NaN` mutations, but they seemed also seemed worth removing to me while I was looking at the range mutator.

I also thought it'd be cool to get rid of emitting `nil`s in ranges, but I didn't know how to do that cleanly.

Let me know if you'd prefer the commits be squashed into one--part of why I kept them separate was because I didn't know if you'd disagree about removing `NaN` though I can't see what they could be used for.

### DIFF
--- a/lib/mutant/mutator/node/literal/range.rb
+++ b/lib/mutant/mutator/node/literal/range.rb
@@ -11,7 +11,7 @@ module Mutant
             erange: :irange
           }.freeze
 
-          children :start, :_end
+          children :lower_bound, :upper_bound
 
           handle(*MAP.keys)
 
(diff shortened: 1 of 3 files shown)


dgollahon:
@mbj: ready for review :)

lib/mutant/mutator/node/literal/range.rb
```diff
@@ -11,7 +11,7 @@ class Range < self
             erange: :irange
           }.freeze
 
-          children :start, :_end
+          children :lower_bound, :upper_bound
```
mbj: I think you can remove these named children and in `#dispatch` simply go for `#emit_children_mutations`.

mbj:
@dgollahon I left one small flag about a possible DSL usage reduction.

PR #623, opened 2016-07-25
Add slack invite system

dgollahon:
Commit Summary:
- Adds a link to the `slackin` invite system.
- Adds badge showing number of active users in slack.
- Adds channel listings for each `mutation-testing` channel with links.

From slack:
> @mbj: I suggest that all tools who have a channel here get PRs adding the information about the channel in the README.

I just went ahead and added *all* the channels to the README, but I don't know if that's what you wanted. Happy to update any of the wording / channel listings / etc. to your liking.

As a note, when you click on the slack badge it goes to the `mutation-testing` directly. I could, instead, make it point to the `slackin` invite server which might increase discoverability but it seemed like clicking "slack" should send you to a slack room. Let me know your preference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@ mutant
 [![Inline docs](http://inch-ci.org/github/mbj/mutant.png)](http://inch-ci.org/github/mbj/mutant)
 [![Gem Version](https://img.shields.io/gem/v/mutant.svg)](https://rubygems.org/gems/mutant)
 [![Flattr](http://api.flattr.com/button/flattr-badge-large.png)](https://flattr.com/thing/1823010/mbjmutant-on-GitHub)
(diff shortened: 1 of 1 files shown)


dgollahon:
@mbj: ready for review

mbj:
@dgollahon I'm fine with the diff but noticed that the image for the slack badge does not render, is this because this PR is not merged, or something that still should be resolved?

dgollahon:
Hmm, where were you looking? Can you check https://github.com/dgollahon/mutant/blob/3e5883660174cedfc6424e6b98da54b9108e7e5d/README.md ? It shows up for me right now. It's possible that when heroku goes to sleep the badge might not resolve, but I would expect it to wake up and render it eventually.

mbj:
> It shows up for me right now. It's possible that when heroku goes to sleep the badge might not resolve, but I would expect it to wake up and render it eventually.

makes sense, mutants readme may have enough traffic to keep it alive. will merge once rebased.

PR #633, opened 2016-08-14
Add mutation from `/foo|bar/` → `/foo/`, `/bar/`

backus:


backus:
@mbj this is rebased and ready for review

backus:
This still depends on #636 being merged

lib/mutant/mutator/node/regexp/alternation_meta.rb
```diff
@@ -0,0 +1,19 @@
+module Mutant
+  class Mutator
+    class Node
+      module Regexp
+        # Mutator for pipe in `/foo|bar/` regexp
+        class AlternationMeta < Node
+          handle(:regexp_alternation_meta)
+
```
mbj: `#dispatch` should normally be private.

backus: k

PR #634, opened 2016-08-14
Add regexp support for /\h/

backus:


backus:
@mbj ready for review

PR #635, opened 2016-08-14
Extract regexp meta duplication to `regexp_mutations`

backus:


backus:
@mbj ready for review

PR #636, opened 2016-08-14
Remove Node::Regexp::Generic

backus:
Extracted from #633. For regexp mutations to be useful the node types which do not have dedicated mutators need to at least invoke the mutators for their children in order to be helpful. This change makes a separate generic mutator for regular expressions unnecessary

### DIFF
--- a/lib/mutant/mutator/node/generic.rb
+++ b/lib/mutant/mutator/node/generic.rb
@@ -5,16 +5,45 @@ module Mutant
       # Generic mutator
       class Generic < self
 
+        unsupported_nodes = %i[
+          ensure
+          redo
+          retry
+          arg_expr
+          blockarg
+          kwrestarg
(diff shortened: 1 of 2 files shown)


backus:
It would be nice to add something like this in a followup PR:

```ruby
module Mutant
  # Registry for mapping AST types to classes
  class Registry

    ...

    # Return unhandled AST types
    #
    # @return [Array<Symbol>]
    def unhandled
      AST::Types::ALL - contents.keys
    end

  end
end
```

and then either configure the mutator to default to `Node::Generic` for unhandled AST types OR simply require 'node/generic' last and have it specify `handle(*REGISTRY.unhandled)`

backus:
@mbj ready for review

backus:
@mbj this is rebased and ready for review

PR #640, opened 2016-09-16
Remove unused let definitions in unit specs

backus:


backus:
@mbj this is ready for review

PR #641, opened 2016-09-16
Fix indeterminism from include_context condition

backus:
It looks like this guard clause was causing issues that led to
the build sometimes failing with

    super: no superclass method `expression'

this change fixes this issue

Extracted from #640


backus:
Closing this in favor of #642

PR #642, opened 2016-09-25
Upgrade to latest devtools

backus:
Devtools 0.1.10 upgraded to RSpec 3.5 which broke this behavior https://github.com/mbj/mutant/blob/770f34eee55b92bf6066b13c917da50215c16dd5/spec/unit/mutant/ast/regexp_spec.rb#L73.

The second commit upgrades to the latest devtools version which upgrades reek and requires working around a few new smells. One of the smells is `InstanceVariableAssumption` which flags code that uses instance variables that were not initialized in the the `#initialize` method. I've fixed these smells by introducing private attribute readers. The other smell is a `ManualDispatch` smell which I've added a reek ignore for.

Finally, I've updated the spec warnings file to match a parser update

Extracted from #640

### DIFF
--- a/spec/unit/mutant/ast/regexp_spec.rb
+++ b/spec/unit/mutant/ast/regexp_spec.rb
@@ -70,9 +70,9 @@ module RegexpSpec
 
         include_context 'regexp transformation'
 
-        return if regexp.encoding.name.eql?('ASCII-8BIT')
-
-        include_context 'regexp round trip'
+        unless regexp.encoding.name.eql?('ASCII-8BIT')
+          include_context 'regexp round trip'
+        end
       end
     end
   end
--- a/lib/mutant/env/bootstrap.rb
+++ b/lib/mutant/env/bootstrap.rb
@@ -56,16 +56,21 @@ module Mutant
         Env.new(
           actor_env:        Actor::Env.new(Thread),
           config:           config,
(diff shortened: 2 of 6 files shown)


backus:
@mbj ready for review

mbj:
@backus As always thank you.

spec/unit/mutant/ast/regexp_spec.rb
```diff
@@ -70,9 +70,9 @@ def self.expect_mapping(regexp, type, &block)
 
         include_context 'regexp transformation'
 
-        return if regexp.encoding.name.eql?('ASCII-8BIT')
-
-        include_context 'regexp round trip'
+        unless regexp.encoding.name.eql?('ASCII-8BIT')
```
dkubb: I wonder if it would be cleaner to use the constant here, eg:

```ruby
unless regexp.encoding.equal?(Encoding::ASCII_8BIT)
```

mbj: @dkubb AFAIK only mutants spec suite has such places, regardless a constant lookup is more primitive than a lookup on a presentation (the string) of that constant. Very good catch.

backus: Good idea I can make that change

PR #643, opened 2016-09-25
Remove unused methods on test doubles

backus:


backus:
@mbj this is ready for review

spec/unit/mutant/subject_spec.rb
```diff
@@ -26,8 +26,7 @@ def foo
   let(:context) do
     double(
       'Context',
-      source_path: 'source_path',
-      source_line: 'source_line'
+      source_path: 'source_path'
```
dkubb: Not a blocker, but this could probably be moved up to a one-liner.

Issue #572, opened 2016-05-23
Add mutation `/$/` → `/\z/`


backus:

mbj:
@backus I bet you meant `/\z` as the target.

backus:
Fixed

PR #646, opened 2016-10-09
Add mutation for $ -> \z

dgollahon:
Addresses #572


lib/mutant/mutator/node/generic.rb
```diff
@@ -37,17 +37,18 @@ class Generic < self
         ]
 
         unsupported_regexp_nodes = AST::Types::REGEXP.to_a - %i[
-          regexp_root_expression
+          regexp_alternation_meta
           regexp_bol_anchor
-          regexp_word_type
-          regexp_nonword_type
           regexp_digit_type
+          regexp_eol_anchor
```
dgollahon: This one was added, but I decided to sort the rest of the list because I was not sure where to insert it.

dgollahon:
@mbj: Can you review?

meta/regexp/regexp_eol_anchor.rb
```diff
@@ -0,0 +1,8 @@
+Mutant::Meta::Example.add :regexp_eol_anchor do
+  source '/$/'
+
+  singleton_mutations
+  regexp_mutations
+
+  mutation '/\\z/'
```
backus: Can this just be a single `\`?

dgollahon: Yeah :+1:

backus:
@dgollahon this is certainly not a requirement, but what I usually like to do when I'm adding new mutations which require an entirely new set of examples in the meta, is add an initial commit (before adding any source code) with the previous mutations for that source. I then follow that commit with my implementation. See the two commits on #633 as an example

dkubb:
LGTM. Will merge on green CI.

PR #649, opened 2016-10-09
Sort regexp nodes lexicographically

dgollahon:
- To reduce diff noise in following PRs and have a well-defined ordering
  of nodes.


dgollahon:
@dkubb

Issue #574, opened 2016-05-23
Add mutation `/(foo)/` → `/(?:foo)/`


backus:

mbj:
This was addressed via #651

PR #651, opened 2016-10-10
Add capture group -> non capture group mutation

dgollahon:
- Mutates regexps like `(foo)` to `(?:foo)`.

Addresses #574


mbj:
@dgollahon Thx!

PR #652, opened 2016-10-10
Add capture group removal mutation

dgollahon:
Addresses #576.

Depends on #651. Diff: https://github.com/dgollahon/mutant/compare/feature/capture_group_to_noncapture_group_mutation...dgollahon:feature/capture_group_to_contents_mutation

### DIFF
--- a/lib/mutant/mutator/node/regexp/capture_group.rb
+++ b/lib/mutant/mutator/node/regexp/capture_group.rb
@@ -17,6 +17,7 @@ module Mutant
             return unless group
 
             emit(s(:regexp_passive_group, group))
+            # emit(group)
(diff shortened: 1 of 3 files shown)


dgollahon:
I need to investigate the failing test case. @backus added a reproducing unit test in the slack channel that I will work with later on:

```ruby
Mutant::Meta::Example.add :regexp_capture_group do
  source '/^(one|two){2,3}([^d\]efm-qz\,\-]*)(ghi)+$/'
end
```

dgollahon:
This is actually somewhat involved and I don't plan to follow through on it here so I'll close this to reduce noise.

PR #653, opened 2016-10-10
Add mutations from `/a*/` -> `/a+/`

dgollahon:
- Mutates greedy zero-or-more regexps to greedy one-or-more regexps.


lib/mutant/mutator/node/regexp/greedy_zero_or_more.rb
```diff
@@ -0,0 +1,21 @@
...
+        class GreedyZeroOrMore < Node
+          handle(:regexp_greedy_zero_or_more)
+
+          # Emit mutations
+          #
+          # Replace `/a*/` with `/a+/`
+          #
+          # @return [undefined]
+          def dispatch
+            emit(s(:regexp_greedy_one_or_more, *children))
```
mbj: This should recurse into the children.

OT, not blocking: NotAnd we should have a generic forcing function / spec to recurse into children by default.

dgollahon: Ah, you're right. Sorry for the oversight.

> we should have a generic forcing function / spec to recurse into children by default.

Interesting, I wonder what that would look like...

mbj: there are several ideas in the backlog of the slack channel I discussed with @backus

lib/mutant/mutator/node/regexp/greedy_zero_or_more.rb
```diff
@@ -0,0 +1,24 @@
...
+          children :min, :max, :subject
+
+          # Emit mutations
+          #
+          # Replace `/a*/` with `/a+/`
+          #
+          # @return [undefined]
+          def dispatch
+            emit(s(:regexp_greedy_one_or_more, *children))
+            emit_subject_mutations
```
mbj: Should we recurse into *all* children? Not only the `subject`?

dgollahon: Apparently `min` and `max` are `Integer`s and `emit_min_mutations` doesn't work:

```
     Failure/Error: self::REGISTRY.lookup(node.type).call(node, parent)

     NoMethodError:
       undefined method `type' for 0:Fixnum
```

I'm happy to add mutations for these at some point but I assume I have to do something custom to get this to work.

mbj: k. I assumed a similar AST design than we have in ruby. This should be left as is for now.

Curious why we have both `min` and `max` in the first place, when the node defines matching numbers already by type.

mbj: There is an explicit repetition syntax: `a{1,2}` so I assume that this attributes are leaked from that node.

PR #655, opened 2016-10-10
Add mutation from `foo[bar] || baz` to `foo.fetch(bar) { baz }`

backus:


backus:
@mbj this is ready for review

lib/mutant/mutator/node/binary.rb
```diff
@@ -60,6 +61,21 @@ def emit_left_negation
           emit(s(node.type, n_not(left), right))
         end
 
+        # Emit mutation from `a[b] || c` to `a.fetch(b) { c }`
+        #
+        # @return [undefined]
+        def emit_fetch_mutation
+          send_node = AST::Meta::Send.new(left)
```
mbj: Instantiating `AST::Meta::Send` is guarded to only happen for `left` nodes of type `:send`? It should.

backus: Mutant kept removing it and I'm not sure I can think of another node that might have `:[]` as the second positional argument in the node (making it the `#selector` attribute under the meta)

mbj: This is a "type check by AST structure", I've to think about counter examples, but from a first glance I think you are right.

backus: @mbj I ended up adding a test case that used `csend` which justifies including `n_send?`

mbj: We should not create `send_node` that assumes `left` is `:send` before the node type check.

backus: k

backus: @mbj fixed

backus:
@mbj ready for another pass

mbj:
Closing this PR as all contributions will require a CLA from now on.

PR #656, opened 2016-10-10
Add mutation from `proc { }` to `lambda { }`

backus:


backus:
@mbj this is ready for review

backus:
@mbj this is ready for review

lib/mutant/ast/meta/send.rb
```diff
@@ -21,6 +21,17 @@ class Send
 
         public :arguments
 
+        # Test if node is defining a proc
+        #
+        # @return [Boolean]
+        def proc?
+          selector.equal?(:proc) ||
```
mbj: Should this branch assert the receiver is implicit self?

backus: > Should this branch assert the receiver is implicit self?

Addressed

backus:
This is rebased now

lib/mutant/ast/meta/send.rb
```diff
@@ -21,6 +21,17 @@ class Send
 
         public :arguments
 
+        # Test if node is defining a proc
+        #
+        # @return [Boolean]
+        def proc?
+          !receiver && selector.equal?(:proc) ||
```
mbj: I think you should split this up into 2 private methods: `naked_proc?` and `proc_new?` or something.

lib/mutant/ast/meta/send.rb
```diff
@@ -59,6 +66,25 @@ def receiver_possible_top_level_const?
...
+        end
+
+        # Test if node is `Proc.new { ... }`
+        #
+        # @return [Boolean]
+        def proc_new?
+          receiver                &&
+            selector.equal?(:new) &&
+            n_const?(receiver)    &&
+            Const.new(receiver).name.equal?(:Proc)
```
mbj: I'd love if we could at some point simply inject a value probe to evaluate the constant in the environment to simply do `#equal?(::Proc)`.

PR #658, opened 2016-10-11
Add mutation from `[]` to `Set.new`

backus:


backus:
@mbj this is ready for review

tjchambers:
As I have proven on real code this is a silent killer because of how AR does not consider an `Array` and `Set` to be similar.

lib/mutant/mutator/node/literal/array.rb
```diff
@@ -32,6 +33,15 @@ def mutate_body
             end
           end
 
+          # Emit `Set.new`
+          #
+          # @return [undefined]
+          def emit_empty_set
+            return unless children.empty?
```
backus: WDYM move it into a guard? Like this?

```ruby
emit(s(:send, s(:const, nil, :Set), :new)) unless children.empty?
```

mbj: Actually `if`, not unless.

```
return unless foo
bar
```

Is equivalent to:

```
bar if foo
```

backus: Er, yeah. Wasn't thinking. Alrighty

tjchambers:
I look at this from this perspective, which I assume you argue with: 

When #mutant mutates code on one side it is transforming the code into something that highlights missing specs. In another side it is challenging the implementation in a way that "suggest" (my word) equivalency. 

In doing so mutant is taking some responsibility in recommending a potential alternative to the code in question. In many cases it is stronger code form being recommended. But what I believe - and I could be mistaken - is it does not not mutate into something that has different behavior, behavior that removes functionality. 

The fact that mutations are performed without any caveats, or educational references, is IMHO a fundamental flaw. When a mutation is highlighted in the log there is nothing akin to rubocop or reek that provides a reference to that specific mutations behavior and the underlying reasoning. 

I realize this is OSS. For it to be adopted and for the masses of developers to be educated as to the value, it deserves to explicitly educate the community - like myself - rather that dismiss that as "their fault".

mbj:
> When #mutant mutates code on one side it is transforming the code into something that highlights missing specs. 

Correct.

> In another side it is challenging the implementation in a way that "suggest" (my word) equivalency.

Incorrect. Not equivalency: Reduced semantics.

>  But what I believe - and I could be mistaken - is it does not not mutate into something that has different behavior, behavior that removes functionality.

Instance of above: It intentionally mutates to something less powerful and you have to proof to the tool you want the more powerful version, or you change to the less powerful version to fix the "mutant flag" (alive mutation).

> I realize this is OSS. For it to be adopted and for the masses of developers to be educated as to the value, it deserves to explicitly educate the community - like myself - rather that dismiss that as "their fault".

Yeah. And I've made my policy clear:

* I do not care about other people, I only care about myself in OSS and my clients.
* If my code is usable "as-is" without compromises for other people: Great, cool! This is the primary reason I put such tools into the OSS. So other people can use the tools also, but if they do not like them "as is" they either have to argument with me to find a flaw in my reasoning, or fork. I cannot make everyone happy nor I try. Nor I care about having "many users". I care about having a value producing tool for me, my clients first: Than for everybody else (if at all, as my OSS time is likely exhausted by the "for me" part already).

mbj:
To put it on record here:

* Number of users is an irrelevant metric for me.
* More users is actually counter productive.
* This is NOT a commercial product where more users would be important.
* More users do not make maintenance easier.
* I'm totally happy to discuss mutations that may be wrong, but I do not see the point into selectively disabling "good" (from the perspective of the axioms that served us well) mutations just to make it easier for entry level users.

lib/mutant/mutator/node/literal/array.rb
```diff
@@ -16,8 +16,9 @@ def dispatch
             emit_singletons
             emit_type
             mutate_body
+            emit_empty_set
             return unless children.one?
-            emit(children.first)
+            emit(Mutant::Util.one(children))
```
mbj: I think you want to combine the early return above with a guard for this emit.

mbj:
@backus With the 2 flags addressed and rebased I'm ready to merge this.

kbrock:
Please reconsider.

This mutation will complain when a user uses a basic array.
That seems a bit much.

Maybe we want a `--pedantic` mode for stuff like this?

mbj:
> Maybe we want a --pedantic mode for stuff like this?

No options, never. Maintenance overhead. I'll not do this on my limited time. Nor I'd maintain it.

I'll run this mutation against mutant itself before adding it, and some of my own major commerical code bases.

backus:
Heh @kbrock I'll probably also run this on a few codebases of mine before I finish this mutation. Looking back at it now I'm actually not sure if it is a good idea or not. I'll need to look at the effect first.

kbrock:
I appreciate that this is for you, and you are being nice by letting others (like me) use it.

But to agree with @tjchambers - I consider read mutant's output as a statement like: I suggest you use this other code instead, unless you can prove to me otherwise (with better specs). It forms a style guide.

But when mutant suggests I change from a very common coding practice to one that is not so common, it can feel like an academic troll. Especially when I don't know how to write a test to prove that I want a basic array over a Set. Or if I do know how to write the test, it is an arbitrary `responds_to?` or `kind_of?` test

mbj:
> But when mutant suggests I change from a very common coding practice to one that is not so common, it can feel like an academic troll. Especially when I don't know how to write a test to prove that I want a basic array over a Set. Or if I do know how to write the test, it is an arbitrary responds_to? or kind_of? test

One that shows you'd accept duplicates.

kbrock:
>One that shows you'd accept duplicates.

We're talking about discouraging using a simple array?

Yes, I appreciate that the duplicate aspect is more limiting. But it is also returning values in an arbitrary order. I thought it gave ActiveRecord a hard time as well (can't remember the specifics though)

Heck, ruby even changed hash to have an order.
Yes, order should not be needed for a hash, but it sure makes life easier for many developers.

With this change, every time I want to use an array, I need to consider the differences between an array and a set, and find a way to justify using a very basic datatype.

Seems a mutation like this just pushes developers into the weeds.

Sorry, this is not getting us anywhere

Thanks for the great tool

lib/mutant/mutator/node/literal/array.rb
kbrock: thanks - great mini thread. will take this back to my team

mbj:
> We're talking about discouraging using a simple array?

In case duplicates are not valid for the datastructure you are modeling, yes.

> Yes, I appreciate that the duplicate aspect is more limiting. But it is also returning values in an arbitrary order. I thought it gave ActiveRecord a hard time as well (can't remember the specifics though)

AR does produce arrays in its API, mutant would not change this API. All your call sides to AR would still work as expected.

> With this change, every time I want to use an array, I need to consider the differences between an array and a set, and find a way to justify using a very basic datatype.

Everytime you use a literal empty array you'd ask your self if a set would be the better choice over an array a more powerful data structure that allows duplicates and meaningful order. Yes.

> Seems a mutation like this just pushes developers into the weeds.

If it does not push myself into the weeds, I'm fine with it. Sorry to disappoint you I do not create this tool to make anyone but myself happy.

But you may have noticed: This PR was not yet merged, I still have to test it on my own codebases to prove my above reasoning is actually valid against the commercial needs of these projects which drive all the decisions on my OSS acitvities.

lib/mutant/mutator/node/literal/array.rb
mbj: Feel free to invite your team into the discussion. OT: did you ment to reply to this part of the thread?

kbrock:
@backus thanks for all your work here. You've introduced a lot of momentum back to the project. Very cool stuff.

backus:
> @backus thanks for all your work here. You've introduced a lot of momentum back to the project. Very cool stuff.

You're welcome :smile:. I'm glad its helpful. I've been able to grow a lot as a developer over the past few years thanks to tools like mutant so I personally also really enjoy adding back to the project. You should checkout [`rubocop-rspec`](http://github.com/backus/rubocop-rspec) too if you're writing RSpec. I've been trying to bring that gem back from the dead over the past 6 months or so.

dgollahon:
I wonder if a set should be preferred by default. The way this works you have to prove you need to be able to hold duplicates (so the set is the default option). But a set is a more complex structure and silently ignores duplicates--and that's fine if that's what you need, but using a set when you don't need one can also be a source of bugs just as much. I don't know that "prove you need duplicates" is any better than if we mutated in the reverse direction "prove you need this structure to swallow duplicates".

Seems to me like it would make the most sense to have both or neither. Personally I don't want both because I think it would be unlikely to save me from any trouble (I could be wrong there) and it would be somewhat difficult to demonstrate why you need one over the other in some code paths (so more testing overhead).

backus:
@dgollahon I agree. You articulated better what I was roughly thinking earlier when @kbrock woke this PR up. A set certainly doesn't seem more primitive to me.

backus:
I'm not 100% convinced but I'm closing this PR for now. I want to investigate a bit further how I feel about the cases that this mutation pulls up in my own codebase, but after thinking a bit more (and talking to @dgollahon) I think this mutation is wrong. It seems like this mutation is trying to be more strict but I think it is just forcing a burden on the user. As @dgollahon mentioned to me privately, if there was a data structure in the ruby standard library that raised an error when you added a duplicate entry then I'd be all for this. A set does not though it has silently different semantics that I don't think are arguably any better than an array.

tjchambers:
@backus sorry to see this closed.

Perhaps this is irrelevant, but a mutation that mutated USAGE of an array in such a way as to create duplicates may be an interesting mutation, but not necessarily befitting of the *stronger* notion. To be more clear I mean `[x]` -> `[x, x]` to indicate the software supports or proves otherwise this notion of duplicate elements.

kbrock:
@backus Did you get a chance to run this on other projects? Did it give you some insight or do you feel it just generated noise?

>I'm not 100% convinced but I'm closing this PR for now

This is at least the second time the `[] -> Set` mutation has come up.
And that is probably for a reason.
There is a concept in here that wants to rear its head. I just can't find it.

PR #659, opened 2016-10-11
Add mutation from `!!foo` to `foo`

backus:


backus:
@mbj this is ready for review

lib/mutant/mutator/node/send.rb
```diff
@@ -126,6 +127,18 @@ def emit_receiver_selector_mutations
...
+        # Emit mutation from `!!foo` to `foo`
+        #
+        # @return [undefined]
+        def emit_double_negation_mutation
+          return unless selector.equal?(:!) && n_send?(receiver)
+
+          negated = AST::Meta::Send.new(meta.receiver)
+          return unless negated.selector.equal?(:!)
+
+          emit(negated.receiver)
```
mbj: Should this implicit return the explicitly guarded to remove the eager return?

backus: Yup

mbj:
@backus We forgot the changelog. Feel free to PR and merge that one yourself.

backus:
@mbj alright. I should probably do a pass over changes since last release in general. I think many things are not in the changelog

mbj:
@backus Unfortunate but correct.

PR #660, opened 2016-10-11
Add compound assignment mutations

dgollahon:
Mutate compound assignment expressions like `a += b` to `a + b` and `a = b`. This idea occurred to me this afternoon. In the case of `a += b` => `a + b`, since the expression has the same result, the assignment aspect could easily go untested in many cases. I'm less sure of the usefulness of `a = b`, but I can think of a few contrived cases where it would be valid. It seems reasonable to me to try breaking the two operators into two mutations of each part though.

### DIFF
--- a/config/flay.yml
+++ b/config/flay.yml
@@ -1,3 +1,3 @@
 ---
 threshold: 16
-total_score: 1300
+total_score: 1296
--- a/lib/mutant/mutator/node/op_asgn.rb
+++ b/lib/mutant/mutator/node/op_asgn.rb
@@ -20,6 +20,31 @@ module Mutant
             !n_self?(node)
           end
           emit_right_mutations
+          emit_compound_assignment_mutations
+        end
+
+        # Mutate compound assignments like `+=` to `+` and `=`
+        def emit_compound_assignment_mutations
+          case left.type
+          when :lvasgn then emit_lvar_mutation
+          when :ivasgn then emit_ivar_mutation
+          when :send   then emit_send_mutation
+          end
+        end
+
+        def emit_lvar_mutation
+          emit(s(:send, s(:send, nil, *left), operation, right))
+          emit(s(:lvasgn, *left, right))
+        end
+
+        def emit_ivar_mutation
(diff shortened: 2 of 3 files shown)


dgollahon:
@mbj: Could you review?

dkubb:
@dgollahon you can think of `a += b` as being equivalent to `a = a + b`. From there you could apply all the normal mutations, eg: `a = a`, `a = b`, `a + b`, `a`, and so on.

I don't know if there's precedence in mutant for this "expand then mutate", but it would also be useful for other kinds of compound operations like `||=`, `>=`, `<=` and so on.

mbj:
Note that `a += b` is not the semantic equivalent to: `a = a + b` because in the latter: `a` is evaluated twice.

Same for all other op assign operations. I do not have the time currnetly to review this in debt but we should keep it in mind.

dkubb:
> Note that a += b is not the semantic equivalent to: a = a + b because in the latter: a is evaluated twice.

Well, that specific case it would be an lvar. If it were a method then it'd be written as `self.a` then `#a` (to retrieve the current value) and then `#a=` would be executed. You're correct that a expansion probably won't work if there are side effects; if they are pure functions then they are equivalent -- although I guess in ruby we have no way of knowing if they are pure or not.

mbj:
> Well, that specific case it would be an lvar. 

True, I do not reject the mutation based on the "could be evaluated twice" scenario, I do only add this information to the context of the discussion.

Before we consider to merge this I do want to do some corpus testing.

dkubb:
I would be ok if an actual "expand + mutate" mechanism didn't happen, just that we are thinking about what possible mutations can be applied to the expanded forms; since many of the are valid transformations for the compound form.

mbj:
> I would be ok if an actual "expand + mutate" mechanism didn't happen, just that we are thinking about what possible mutations can be applied to the expanded forms; since many of the are valid transformations for the compound form.

yeah, agreed.

dgollahon:
Willing to rebase/update this if desired, but since it wasn't originally approved and has gone stale, i'll close it.

PR #662, opened 2016-10-13
[WIP] Fail fast on neutral failures

backus:
This PR introduces an incremental improvement in failing fast on neutral failures. This is my first pass at implementing neutral failure fast failing. I still need to add functionality for displaying a special error message in cases where execution was halted due to a neutral failure. Otherwise this code is ready for review


spec/unit/mutant/result/subject_spec.rb
```diff
@@ -106,4 +106,30 @@
...
   end
+
+  describe '#neutral_failure?' do
+    subject { object.neutral_failure? }
+
+    context 'without failing neutral mutations' do
+      let(:mutation_results) do
+        [
+          instance_double(Mutant::Result::Mutation, success?: false, neutral_failure?: false),
+          instance_double(Mutant::Result::Mutation, success?: true, neutral_failure?: false)
```
dkubb: I'd probably vertically align the `neutral_failure?` key here for easier readability.

lib/mutant/runner/sink.rb
```diff
@@ -49,6 +49,20 @@ def result(mutation_result)
 
     private
 
+      # Test if an evil failure has been reported
+      #
+      # @return [Boolean]
+      def evil_failure?
+        !status.subject_results.all?(&:success?)
```
mbj: I'd be okay to have a `FailurePredicate` mixin that defines `#failure?` as the negation of `#success?`. But for consistency we should only ever concretely implement `#success?`.

mbj:
@backus I'm fine with the overall approach but flags.

I do not think the "we terminated because of neutral violation" message has to be added in this itartion.

Terminating on neutral is the correct semantics. Hence this is a free-standing improvement.

To implement the meassaging correct, I'd consider to define a "termination reason" type:

```
data TerminationReason = Done | NeutralViolation Mutation
```

Which than gets expanded by the CLI (or other reporters) to dedicated messaging.

lib/mutant/runner/sink.rb
backus: Added the failure predicate. I think it turned out really nice.

spec/unit/mutant/result/subject_spec.rb
backus: Updated

mbj:
Closing this PR as all contributions will require a CLA from now on.

Also this approach will not fit into the currnent runner. I'll do an implementation of these semantics on my own.

PR #663, opened 2016-10-14
Remove unnecessary exclude from integrations.yml

backus:


backus:
Upstream rubyspec changes led to this build failure. @mbj ready for review.

PR #667, opened 2016-10-16
Fix rspec warning about ordered stubs

backus:
Previously the spec output was spammed with

```
WARNING: `allow(...).to receive(..).ordered` is not supported and willhave no effect, use `and_return(*ordered_values)` instead.. Called from /home/dev/mutant/spec/unit/mutant/reporter/cli/tput_spec.rb:6:in `expect_command'.
```


backus:
@mbj this is ready for review

PR #668, opened 2016-10-16
Fix spec warning assertions when running specs under mutant

backus:
Certain mutations emit warnings which can cause an otherwise passing
mutant run to fail. For example, the `remove_method` mutation will
emit a warning when applied to certain methods like `#initialize`,
`#object_id`, and `#__send__`.

See https://git.io/vPiFS


backus:
Going to close this in favor of removing the `remove_method` mutation

Issue #108, opened 2013-09-08
Add Mutation: def my_method; end -> remove_method :my_method


dkubb:
The purpose of this mutation is to remove methods that are duplicating functionality in their parent class. It will expose cases where methods were pushed up to the parent, but not removed from the child classes.

PR #674, opened 2016-10-21
Remove `remove_method` mutation

backus:
See also #108


backus:
@mbj as discussed earlier

mbj:
Yes this is better done as a boot time mutation. It violates the "user should be able to edit the mutation in and observe the same effect" invariant.

backus:
Rebased

PR #676, opened 2016-10-21

mjago:
As the author of the ruby spec utf-8 changes it may save you trouble if I make these PR's ?


backus:
That would be fantastic. I would really appreciate that

backus:
Thanks @mjago

PR #678, opened 2016-10-22

mjago:


mbj:
@mjago thx, as usual.

PR #681, opened 2016-10-24

mjago:


mbj:
@mjago Thx. Soon the exclude list will be empty, I'd be happy to remove the entire mechanism!

PR #684, opened 2016-10-26

mjago:


mbj:
@mjago Could it be you use this excludes as guide which violations to fix in rubyspec ;) ?

mjago:
@mbj actually It's the parser gem - but I believe you use that too?

mbj:
> @mbj actually It's the parser gem - but I believe you use that too?

yes, I was assuming you are the person behind the changes to rubyspec (or parser)?

PR #445, opened 2015-09-22
Add test minitest integration

mbj:
[fix #92]

References: #330.

PR #687, opened 2016-10-29

mjago:
@backus This is the last of the incompatible utf-8 escape sequence errors so I removed the entire section from integrations.yml :smiley: 
How are you detecting the regexp parse error - I'm not getting that?

### DIFF
--- a/spec/integrations.yml
+++ b/spec/integrations.yml
@@ -7,25 +7,6 @@
   expected_errors:
     "#<Parser::SyntaxError: invalid multibyte escape: /\xAA/>":
       - language/regexp/escapes_spec.rb
-    "#<Parser::SyntaxError: literal contains escape sequences incompatible with UTF-8>":
-      - core/string/shared/each_codepoint_without_block.rb
-      - core/string/shared/encode.rb
-      - core/string/shared/eql.rb
-      - core/string/shared/succ.rb
-      - core/string/slice_spec.rb
-      - core/string/squeeze_spec.rb
-      - core/string/unicode_normalize_spec.rb
-      - core/string/valid_encoding_spec.rb
-      - core/symbol/casecmp_spec.rb
-      - core/time/_dump_spec.rb
-      - core/time/_load_spec.rb
-      - language/regexp/encoding_spec.rb
-      - language/string_spec.rb
-      - library/openssl/shared/constants.rb
-      - library/socket/socket/gethostbyname_spec.rb
-      - library/zlib/inflate/set_dictionary_spec.rb
-      - optional/capi/encoding_spec.rb
-      - optional/capi/string_spec.rb
     '#<RegexpError: invalid multibyte escape: /\xAA/>':
       - language/regexp/escapes_spec.rb
     "#<Regexp::Scanner::PrematureEndError: Premature end of pattern at #{str}>":


backus:
Awesome! Thanks again for doing this upstream maintenance and helping sync up our error whitelist. I'll look into the regexp parse error soon. Right now though I need to look into a small bug that your upstream changes helped us find (thank you for that)!

backus:
@mjago the remaining errors are encountered while mutating regular expressions that we parse using [`Regexp::Parser`](https://github.com/ammar/regexp_parser).

backus:
Thanks @mjago! I had to add another change to the commit to fix a bug your work helped uncover. I've preserved the commit attribution to you. Thanks again for the help!

mbj:
@backus Technically you could have added the bugfix to mutants master before merging this PR. But as the gross of the commit is attributed correctly to @mjago its probably fine that way.

backus:
@mbj no since we always pull in the upstream rubyspec (and @mjago's changes are already in upstream master) there was no way to get a passing build without having one commit that had both the updated expected errors AND the exposed bugfix

mjago:
Glad I could help - hope to come back and check out Mutant more - looks like a very interesting project. I wonder if it can be made to work with mspec (derivative of rspec) ?

backus:
@mjago probably although I don't know exactly the differences between rspec and mspec. Checkout https://github.com/mbj/mutant/pull/445 for an example of what it looks like to integrate a different test runner integration

mbj:
> I wonder if it can be made to work with mspec (derivative of rspec) ?

The API needed is minimal. Maybe lets meet int the mutant slack channel to discuss details?

dkubb:
> since we always pull in the upstream rubyspec

@mbj @backus Should the version of rubyspec be "fixed" in a specific config file? This way you can always know what version passes against mutant and not have random failures in the future? It would mean bumping the version explicitly, but I think that's ok.

I'd probably add explicit sha hashes for all 4 of the projects that are pulled in from `spec/integrations.yml`.

mjago:
> Maybe lets meet int the mutant slack channel to discuss details?

@mbj Sounds good

backus:
@dkubb, @mbj suggested that before but I figured that the main source of entropy (rubyspec) would be resolved soon and I otherwise like the periodic bug fixes and lack of need to do "integration update" PRs. Still happy to do it (I don't mind much either way) but I think a lot of the pain will be gone now.

dkubb:
In general I like for my dependencies to be fixed because I don't like it when a previously passing build suddenly starts failing, even though nothing in the build source has changed. Anything, that if it changed, could effect the build should be accounted for imho.

I'll let @mbj weigh in though, since this is his project and I am just an interested bystander. ;)

mbj:
> @mbj @backus Should the version of rubyspec be "fixed" in a specific config file?

If we had no @mjago I'd go for this. But as "most" of the rubyspec pain was resoved by his work, I think we should continue to pull in rubyspec to introduce the right kind of pressure here to fix our stuff.

In the history of this project rubypsec was a very good source of bugs in the mutation testing engine, and before the work on fixing the encoding problems by @mjago was started: This continous stream of variance was no big problem.

I'd vote for keeping it as is. In the hope rubyspec now is an even more quality source of: "potentially unexpected ruby AST constructs".

PR #688, opened 2016-10-29
Add support for regexp_bol_escape

backus:
This PR takes the changes in #687, fixes a bug that the upstream rubyspec changes uncovered, and adds one new exclusion to the `integrations.yml` which is visible now that utf-8 issues aren't prohibiting us from parsing the offending file


mbj:
@backus Must this depend on #687 (that fails right now)?

backus:
> @backus Must this depend on #687 (that fails right now)?

Yeah. I'll have to fuse that commit with some other changes

backus:
@mbj this is ready for review

PR #691, opened 2016-12-02
Allow empty match expressions

mbj:


mbj:

PR #693, opened 2016-12-11
Checkin Gemfile.lock for deterministic builds

mbj:
The referenced gems (especially rubocop) innovate on point releases. Deterministic builds are important, nothing that is green on a PR that is merged later to master should be red on merged to master.


backus:
@mbj what about just adding harder constraints in the Gemfile?

mbj:
> @mbj what about just adding harder constraints in the Gemfile?

less easy to maintain.

PR #711, opened 2017-06-01
Define glob pattern for each repo

jekuta:
This PR should fix the failing build.

This file was added not long ago to the rubyspec project https://github.com/ruby/spec/blob/master/command_line/fixtures/bad_syntax.rb

In the ruby corpus specs mutant iterates on all the rubyspec files. When [`Parser::CurrentRuby.parse(path.read)`](https://github.com/mbj/mutant/blob/master/spec/support/corpus.rb#L149) hits that file a `Parser::SyntaxError` is raised.

Since the offending file was created as a fixture, mutant probably shouldn't run it? 

This PR makes the glob pattern configurable on repository level for the integrations. I'm don't know if this a direct you want to take with these specs, but I created it more to show a one way to fix this issue.

### DIFF
--- a/spec/integrations.yml
+++ b/spec/integrations.yml
@@ -3,6 +3,7 @@
   namespace: Rubyspec
   repo_uri: 'https://github.com/ruby/rubyspec.git'
   repo_ref: 'origin/master'
+  ruby_glob_pattern: '**/*_spec.rb'
   mutation_coverage: false
   mutation_generation: true
   expected_errors:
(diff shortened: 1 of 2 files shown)


mbj:
@jekuta I like this idea, more than what I planned to do. As it should be more stable over time.

Can you confirm that the fixture you (indirectly) exclude is intentionally invalid ruby? If so I'd be willing to merge this as is.

jekuta:
@mbj Yeah, it was added together with a spec that tests the output when a file with invalid syntax is given https://github.com/ruby/spec/commit/e6665aea4fdc521332cddd879fa54e36948e7245

mbj:
You got it. The `**/*_spec.rb` pattern should cover what mutant intents to use as corpus.

jekuta:
Great, thanks for fast reply!

mbj:
@jekuta Heh, thx for looking at the established patterns in this repo and creating a PR that fits it the first iteration. You re-used all the machinery, concord, morpher, ... perfect contribution.

PR #719, opened 2017-07-03
Add Ruby 2.4 support for mutant

barthez:
Mutant gem was producing deprecation warnings while running on Ruby 2.4 regarding Fixnum usages. This PR fixes those warnings. It also does:

 - Update RuboCop and fix offenses
 - Relax parser gem version requirement (it will allow to run and test unparser gem with parser 2.4)
 - Update other dependencies: unparser, devtools, diff-lcs
 - Run regexp_parser integrations against v0.4.3 - fix incompabilities
 - Make CI use Ruby 2.4
 - Fix tests with getting warnings.yml whitelist up to date

/cc @mbj


spec/support/warning.rb
```diff
@@ -57,7 +57,7 @@ def add(warning)
     end
 
     warnings  = Pathname.new(__dir__).join('warnings.yml').freeze
-    whitelist = IceNine.deep_freeze(YAML.load(warnings.read))
```
pyromaniac: Don't do this, just disable rubocop, it doesn't know what it wants.

circle.yml
```diff
@@ -1,7 +1,7 @@
 ---
 machine:
   ruby:
-    version: '2.3.3'
+    version: '2.4.1'
```
pyromaniac: Can't we run test over the both versions just in case?

spec/support/warning.rb
pyromaniac: It will cause problems with symlinks support and other yaml stuff, in this case we are loading yaml from the trusted source, so load is good enough

circle.yml
barthez: I'm not sure, we would need to migrate to Circle CI 2.0. I will check if it is possible.

mbj: I'd love to get a test matrix for 2.3 and 2.4 branches of ruby.

Otherwise I'd be fine to simply drop 2.3 support. People who need it can use older versions of mutant.

barthez:
We are not able to run mutant tests with regexp_parser on ruby 2.3.4 (only one provided with circleci docker images).

Waiting for https://github.com/ammar/regexp_parser/pull/40 to proceed.

barthez:
@mbj It is read your final review.

I've fixed ruby version to 2.3.3 on CI so tests will pass. It can be changed when https://github.com/ammar/regexp_parser/pull/40 will be merged and released.

I guess you need to update required checks for this repo if you want to merge it.

Do you want me to squash everything to single commit?

mbj:
@barthez Thx. I'll do a pass over this later this day.

mbj:
@barthez I'm fine with all of this but do want to investigate a bit into the last commit. It should not be required. Can you link me to the error that shows up without it?

Other than that: I'd like to get each build in a PR passing, not only relying on the last one to be green. This makes bisecting much easier.

There are the following options to archive this, in order of preference descending:

* Trigger a build for each commit
* Extract other PRs from this commit
* Squash them all together

Let me know if you want to do this, if not I'll handle it by myself.

barthez:
Let [this build](https://circleci.com/gh/mbj/mutant/3197) be example. Check out artifacts - those are `bundle install` output logs from rspec specs. `Bundler.with_clean_env` was removing too much of an environment. It was working on CircleCI 1.0 probably because of different setup - some caching or stuff.

I would rather squash everything together. As far as I remember it was failing even without even single change.

barthez:
@mbj Ok, I'm done here.

mbj:
> I would rather squash everything together. As far as I remember it was failing even without even single change.

I'm okay with that reasoning.

mbj:
Released as `v0.8.14`.

PR #728, opened 2017-10-07
Add mutation from class to module

ignat-z:
Mostly for util-like classes, something lile

```ruby
class FileNameSanitizer
  def self.call(file_name)
    file_name.gsub(/[^0-9A-Z]/i, '_')
  end
end
```

The main aim is not to change from class to module, but change way of using (initializers and so on)

### DIFF
--- a/lib/mutant/mutator/node/class.rb
+++ b/lib/mutant/mutator/node/class.rb
@@ -13,7 +13,15 @@ module Mutant
         #
         # @return [undefined]
         def dispatch
+          mutate_type
           emit_body_mutations if body
+        end
+
+        # Emit class -> type mutations
+        #
+        # @return [undefined]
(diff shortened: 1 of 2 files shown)


mbj:
@ignat-z This mutation is valid and falls within the "semantic reduction" class of mutations.

But be aware that in current mutant you'd very likely not see it being emitted, as mutant only mutates instance and singleton methods so far. An implementation for constant scope subjects requires so called "boot time mutations" to correctly insert the mutation, as retroactively (runtime mutation insertion) changing the constant scope is not guaranteeing the mutation gets inserted correctly. Think about an aliased constant, or capture in a block etc.

I'm fine to merge this, but please be aware about the limitation. Boot time mutations is something I always wanted to do but never found the time.

PR #735, opened 2018-01-23
Whitelist regexp_parser warning and bump regexp_parser to 0.4.9

zaidan:


mutant.gemspec
```diff
@@ -35,7 +35,7 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency('equalizer',     '~> 0.0.9')
   gem.add_runtime_dependency('anima',         '~> 0.3.0')
   gem.add_runtime_dependency('concord',       '~> 0.1.5')
-  gem.add_runtime_dependency('regexp_parser', '~> 0.4.4')
+  gem.add_runtime_dependency('regexp_parser', '~> 0.4.5', '< 0.4.6')
```
mbj: I suspect specifying `~> 0.4.6` just had the same effect?

zaidan: According to bundler documentation `~> 0.4.6` would be identical to `>= 0.4.6 and < 0.5`. `0.4.6` is not working. If they had use semantic versioning from the beginning, we could write `~> 0.4.5`. This is a stupid Ruby issue. I was wondering if it is possible to switch to 4 version number segments, like in Haskell.

mbj: yeah, I forgot. Do not do much ruby anymore.

So why do we do not want `0.4.6`?

zaidan:
```bash
Mutant on ruby corpus rubyspec does not fail on mutation generation
     Failure/Error: fail RegistryError, "No entry for: #{type.inspect}"
     
     Mutant::Registry::RegistryError:
       No entry for: :regexp_xgrapheme_type
```
see:
* https://circleci.com/gh/zaidan/mutant/19
* https://circleci.com/gh/zaidan/mutant/18

mbj: So instead of keeping mutant on an older regexp parser, why not simply add these new (unmutated) nodes to the noop handler?

zaidan: Do you mean adding :regexp_xgrapheme_type to `handle(...)` in `lib/mutant/mutator/node/noop.rb`? This produces the following error:
```bash
register': Invalid type registration: :regexp_xgrapheme_type (Mutant::Registry::RegistryError)
```

mbj: there is a regexp mutator level noop.

zaidan: do you mean some where in `lib/mutant/mutator/node/regexp/**`? There is no file called `noop`. I did a grep for `noop` and could not find something useful.

mbj: There is a list of nodes we do not explicitly mutate inside that namespace. I do, right now, not have the time to point you to it, sorry. Busy with commercial clients.

zaidan: ok, fixed this and another issue with `>= 0.4.7`. Got it working with `~> 0.4.9`.

mbj: good, this is the much better change.

zaidan:
bump

PR #739, opened 2018-06-17
Stop mutation of `#to_i` to implicit `self`

### LINKED ISSUE #738: Overeager mutation of `#to_i` leads to `unparser` nil bomb
I have managed to reproduce this in a minimal example, but I am having a hard time finding where to add the test within Mutant and Unparser to contribute. I am happy to work on a fix if I can get some pointers on where to start!

## Reproduction

I made a reproduction script [here](https://github.com/michaelherold/mutant_to_i_bug) that you can pull down to see the bug in action.

Here's a quick way to pull down the reproduction:

    git clone git@github.com:michaelherold/mutant_to_i_bug.git
    cd mutant_to_i_bug
    ./run

For ease of reference, here is the code that causes the bug:

```ruby
class Foo
  EPOCH_TIME = 1_000

  def to_i
    1_528_771_800
  end

  def to_time
    Time.at(to_i + EPOCH_TIME)
  end
end
```

My guess is that [this mutator](https://github.com/mbj/mutant/blob/24f6d42fa2b3d30fbf206169d08383972c803242/lib/mutant/mutator/node/send.rb#L181-L188) is the one causing the bug. 

That lone call to `to_i` instead of `self.to_i` causes the error. If I change it to `self.to_i`, the mutation test runs to completion (and reports that I should remove the `self.` 😆 ).

## Stack Trace

Here is the top of the stack trace:

```
bundler: failed to load command: mutant (/Users/michael/.gem/ruby/2.4.2/bin/mutant)
NoMethodError: undefined method `type' for nil:NilClass
  /Users/michael/.gem/ruby/2.4.2/gems/unparser-0.2.7/lib/unparser/emitter/send/arguments.rb:32:in `block in effective_arguments'
  /Users/michael/.gem/ruby/2.4.2/gems/unparser-0.2.7/lib/unparser/emitter/send/arguments.rb:31:in `each'
  /Users/michael/.gem/ruby/2.4.2/gems/unparser-0.2.7/lib/unparser/emitter/send/arguments.rb:31:in `each_with_index'
  /Users/michael/.gem/ruby/2.4.2/gems/unparser-0.2.7/lib/unparser/emitter/send/arguments.rb:31:in `each'
  /Users/michael/.gem/ruby/2.4.2/gems/unparser-0.2.7/lib/unparser/emitter/send/arguments.rb:31:in `map'
  /Users/michael/.gem/ruby/2.4.2/gems/unparser-0.2.7/lib/unparser/emitter/send/arguments.rb:31:in `effective_arguments'
  /Users/michael/.gem/ruby/2.4.2/gems/unparser-0.2.7/lib/unparser/emitter/send/arguments.rb:19:in `block in dispatch'
  /Users/michael/.gem/ruby/2.4.2/gems/unparser-0.2.7/lib/unparser/emitter.rb:185:in `parentheses'
  /Users/michael/.gem/ruby/2.4.2/gems/unparser-0.2.7/lib/unparser/emitter/send/arguments.rb:18:in `dispatch'
  /Users/michael/.gem/ruby/2.4.2/gems/unparser-0.2.7/lib/unparser/emitter.rb:104:in `write_to_buffer'
  /Users/michael/.gem/ruby/2.4.2/gems/memoizable-0.4.2/lib/memoizable/method_builder.rb:117:in `call'
  /Users/michael/.gem/ruby/2.4.2/gems/memoizable-0.4.2/lib/memoizable/method_builder.rb:117:in `block (3 levels) in create_memoized_method'
  /Users/michael/.gem/ruby/2.4.2/gems/memoizable-0.4.2/lib/memoizable/memory.rb:63:in `block (3 levels) in fetch'
  /Users/michael/.gem/ruby/2.4.2/gems/thread_safe-0.3.6/lib/thread_safe/cache.rb:56:in `fetch'
  /Users/michael/.gem/ruby/2.4.2/gems/memoizable-0.4.2/lib/memoizable/memory.rb:62:in `block (2 levels) in fetch'
  /Users/michael/.rubies/ruby-2.4.2/lib/ruby/2.4.0/monitor.rb:214:in `mon_synchronize'
  /Users/michael/.gem/ruby/2.4.2/gems/memoizable-0.4.2/lib/memoizable/memory.rb:61:in `block in fetch'
  /Users/michael/.gem/ruby/2.4.2/gems/thread_safe-0.3.6/lib/thread_safe/cache.rb:56:in `fetch'
  /Users/michael/.gem/ruby/2.4.2/gems/memoizable-0.4.2/lib/memoizable/memory.rb:60:in `fetch'
  /Users/michael/.gem/ruby/2.4.2/gems/memoizable-0.4.2/lib/memoizable/method_builder.rb:116:in `block (2 levels) in create_memoized_method'
  /Users/michael/.gem/ruby/2.4.2/gems/unparser-0.2.7/lib/unparser/emitter.rb:499:in `run'
```

michaelherold:
The mutations of all calls to `#to_i` include one that mutates the call into a
call to `Integer`. This is a great mutation in all calses where there is an
explicit receiver for the `#to_i` call. However, the way that the implicit call
to `self.to_i` is parsed results in a receiver of `nil` instead of `self`.

As such, we want to limit this particular mutation to calls to `#to_i` that have
an explicit receiver. This feels like the correct behavior because the result of
the mutation would be (if it parsed): `Integer(nil)`, which doesn't retain the
intended behavior of the original source.

Another way to approach this would be to mutate the code to `Integer(self)`, but
that doesn't feel as correct to me because if you're using the implicit `self`
with a call to `#to_i`, you likely are not implementing the strict `#to_int`
method as well, which `Integer` relies on.

This fix feels like the right mix of correctness and minimal invasiveness.

Closes #738

### DIFF
--- a/config/flay.yml
+++ b/config/flay.yml
@@ -1,3 +1,3 @@
 ---
 threshold: 16
-total_score: 1321
+total_score: 1316
--- a/lib/mutant/mutator/node/send.rb
+++ b/lib/mutant/mutator/node/send.rb
@@ -182,7 +182,7 @@ module Mutant
         #
         # @return [undefined]
         def emit_integer_mutation
-          return unless selector.equal?(:to_i)
+          return unless receiver && selector.equal?(:to_i)
 
           emit(s(:send, nil, :Integer, receiver))
         end
--- a/meta/send.rb
+++ b/meta/send.rb
@@ -136,6 +136,13 @@ Mutant::Meta::Example.add :send do
 end
 
 Mutant::Meta::Example.add :send do
+  source 'to_i'
+
+  singleton_mutations
+  mutation 'to_int'
+end
+
+Mutant::Meta::Example.add :send do
   source 'foo.to_s'
 
   singleton_mutations


michaelherold:
I adjusted the Flay metric down because it was failing CI. Since it's a downward mass, that feels right, but if it's not, feel free to let me know!

mbj:
@michaelherold I agree to this change. Can you squash the flay change into the previous commit that caused it? I do like a so called "hygenic" commit history, which would be satisfied with the squash.

michaelherold:
Done! I go back-and-forth on whether I like to see those types of commits separately or not.

mbj:
@michaelherold I had the same discussions with my teams, but over the time instead of looking at a "single" case I discovered some higher level axioms should be used to determine the boundaries of commits. There are some interesting ones we typically use:

* No commit is allowed to ever fail the build, or more generally "knowingly reduce the value of the system".
* A commit is either a:
  * Equivalent observable semantics refactoring (example a style fix)
  * Fixing an observale behavior that should never have entered the system in the first place.
  * Adding new behavior.
* If any of the above rules are violated the commit has to be split/merged/changed to comply.

As this repository at some point in the past has decided to make the flay score a hard metric, and changes to the expected metrics fail the build: According to the rules above would demand to do the metric expectation change the same moment you do the code change.

Context on why we track this metric: We do want to acknowledge small drifts explicitly, and react to big, unexpected drifts. If your change for example had bumped the score up by 10% I'd likely have vetoed it for that reason. Because the semantic diff is super small, so the equivalent code diff should also be proportinally small.

michaelherold:
Wow, thanks for the thorough explanation! That's a really interesting way to look at it.

mbj:
> Wow, thanks for the thorough explanation! That's a really interesting way to look at it.

The overall idea is to have a set of axioms you can derive almost all decision trees from as long you agree on the axioms.

This increases the chance that 2 different developers end up in the same solution for a given system, and overall reduces the churn from reviews/discussions.

Basically the idea allows to eliminate large parts of subjective opinions.

PR #740, opened 2018-06-27
Add method body -> super mutation

### LINKED ISSUE #154: Mutation to replace method with proxy to super
This will spot cases where the super implementation is equivalent to the superclasses implementation. This happens in a push down refactoring as a leftover. (at least to me).

Original:

```
def foo(bar)
  computation
end
```

Mutation:

```
def foo(*)
  super
end
```

### LINKED ISSUE #673: Mutate method body to `super`
This mutation would serve the purpose of what https://github.com/mbj/mutant/issues/108 was intended to do

dgollahon:
- Mutates method bodies to `super` to catch cases where a method has
  been needlessly re-implemented and the parent class provides the
  equivalent behavior.
- Closes #154
- Closes #673

### DIFF
--- a/lib/mutant/mutator/node/define.rb
+++ b/lib/mutant/mutator/node/define.rb
@@ -14,6 +14,7 @@ module Mutant
           emit_optarg_body_assignments
           emit_restarg_body_mutation
           emit_body(N_RAISE)
+          emit_body(N_ZSUPER)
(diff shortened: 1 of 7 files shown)


mbj:
@dgollahon Good one. Thx.

mbj:
(wrong button)

PR #742, opened 2018-07-17
Fix boot time issue

mbj:
Context:

* RSpec generates a "semi dynamic constant" per shared/regular example
* These constants have dynamically implemented `#name` which are super
  slow to evaluate in mass. As they leak lots of intermediary objects
  behind and cause lots of GC pressure on BIG code bases.
* Mutant cannot currently mutate 99.95% of typical generated class
  bodies for various reaons (literal code only, no support for mutating
  metaprogrammed code).

Effect:

* On big code bases mutant initialization is much faster now, observed
  speedup is a reduction of boot time between 60 - 95%.
* If your subjects where previously autoloaded implicitly from mutant
  loading all rspec examples (via hitting your constant names that
  trigger autoloads) you'll now be forced to explicitly trigger that
  autloads in a different way. Use `--require
  some-file-that-triggers-all.rb` this will be loaded before mutant
  exands subjects and should be fine. I do not recommend to load rspec
  from that file, or autoloads in general.

TLDR: Loading rspec before subject matching was a hack to make people
that rely on autoloads a bit more content, that was a big mistake, just
as autoloads in general. Lets call autoloads "indeterministic constant
scope" from now on.


spec/support/warnings.yml
```diff
@@ -1,7 +1,9 @@
 ---
+- "lib/mutant/subject/method/instance.rb:9: warning: undefining `initialize' may cause serious problems"
 - 'lib/parallel.rb:222: warning: shadowing outer local variable - args'
 - 'lib/parallel.rb:227: warning: shadowing outer local variable - args'
 - 'lib/parser/lexer.rb:10922: warning: assigned but unused variable - testEof'
 - 'lib/regexp_parser/scanner.rb:1675: warning: assigned but unused variable - testEof'
-- 'lib/regexp_parser/scanner.rb:1692: warning: assigned but unused variable - testEof'
 - 'lib/regexp_parser/scanner.rb:1689: warning: assigned but unused variable - testEof'
+- 'lib/regexp_parser/scanner.rb:1692: warning: assigned but unused variable - testEof'
+
```
snusnu: @mbj Is this newline intentional?

mbj: good catch.

PR #749, opened 2018-09-24
Add regexp unicode print node support

dkubb:
This branch adds support for two unicode regexp nodes:

```ruby
\p{Print}  # only match printable unicode characters
\P{Print}  # only match nonprintable unicode characters
```


lib/mutant/ast/types.rb
```diff
@@ -57,9 +57,8 @@ module Types
         regexp_capture_group
         regexp_carriage_escape
         regexp_character_set
-        regexp_character_set
```
dkubb: @mbj I will break this change out into a separate commit.

PR #751, opened 2018-09-28
Use regexp_parser 1.2.0

janosch-x:
### tl;dr

- updates regexp_parser dependency
- fixes incorrect or missing transformations of some expressions

### long version

hi!

i've done a bit of work on regexp_parser recently. as i'm using both regexp_parser and mutant in one of my gems, i've run into conflicts with mutant still depending on an old version of regexp_parser, so here is an update:

- character sets are now a regular `Subexpression`, so no longer need a custom transformer
- alternations now use the standard constructor, so no longer need a custom transformer
- adds support for conditionals, so `AST::Regexp::supported?` is no longer needed
- adds support for set intersections
- adds support for linebreak type ("\R")
- adds (exemplary) support for posix classes ("[[:print:]]" and "[[:^print:]]")
- fixes expression class mapping for some escape sequences
  - this includes previously incorrect ones and some that changed in regexp_parser 1.0.0
- fixes option group handling
  - uses `#option_changes` since `#options` refers to overall active options
  - no longer ignores negative flags
  - handles switching/childless option groups (e.g. "(?i)") correctly
- fixes recently added `:regexp_print_nonproperty`
  - it has type/token `:nonproperty, :print`, not `:property, :nonprint`
- updates names of a few unicode property tokens, particularly scripts
  - if there is any interest in this, property tokens are reverse compatible with `Regexp` since regexp_parser 1.0.0, so all 777 of them (1554 if you count negative versions as well, or 3108 if you count posix styles ...) could probably be supported in a simpler fashion

let me know if you have any questions or need any more help with your regexp integration.


janosch-x:
The spec should probably be OK now, though I don't have any in-depth knowledge of your architecture.

> happy to see you as author of regexp-parser

[Ammar Ali](https://github.com/ammar/) is the original author, I'm more like a maintainer.

> fixing all the mistakes we did on our side!

I didn't mean to bang on about it. mutant is making nice, exhaustive usage of regexp_parser, so it helped me find as least as many issues in regexp_parser itself. 

Will get back to you (or interested contributors) with a few questions or suggestions soon-ish.

mbj:
@janosch-x Thanks for the spec fix, now only the rubocop violations are left.

https://circleci.com/gh/mbj/mutant/3402?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

I'm fine to fix these by myself, your call.

mbj:
@janosch-x Can you squash so I can merge?

janosch-x:
@mbj we can merge this but i have to say i'm not fully happy with the integration atm. things i'd still like to do:

1. support some more fairly common regex nodes
  - missing ones cause mutant to fail with a `RegistryError`, e.g. `/.??/`
2. spec and fix handling of negated character sets (`/[^a]/`)
3. support properties/posixclasses, maybe with a catchall
  - people actually use them; supporting only the ones tested by rubyspec seems very arbitrary
  - i doubt it makes sense to add 3k items to `Mutant::AST::Types::ALL` and/or `Transformer::Direct`
4. least important: improve existing mutations so they don't return illegal regexes, e.g. for `/()\1/`
  - not sure why they don't fail; does mutant catch `SyntaxError` and treat it as a kill?

mbj:
> @janosch-x Can you squash so I can merge?

What about removing all additions from this PR and doing them in a followup? Just keep the dependency bump and API fixes.

The reminder can be done in a dedicated iteration, but I could release the dependency bump already. People that hit regexps we do not fully support can whitelist those subjects.

janosch-x:
@mbj ok, squashed.

by the way when try to run `bundle exec rake ci` locally, i always get this. am i doing it wrong?

```
evil:TestApp::Literal#uncovered_string:.../mutant/test_app/lib/test_app/literal.rb:16:a2250
@@ -1,4 +1,4 @@
 def uncovered_string
-  "string"
+  super
 end
-----------------------
Mutant configuration:
Matcher:         #<Mutant::Matcher::Config match_expressions: [TestApp::Literal#uncovered_string]>
Integration:     Mutant::Integration::Rspec
Jobs:            8
Includes:        ["lib"]
Requires:        ["test_app"]
Subjects:        1
Mutations:       6
Results:         6
Kills:           1
Alive:           5
Runtime:         0.39s
Killtime:        0.09s
Overhead:        330.21%
Mutations/s:     15.32
Coverage:        16.67%
.

Failures:

  1) Mutant on ruby corpus auom does have expected mutation coverage
     Failure/Error: fail "System command failed!: #{arguments.join(' ')}"

     RuntimeError:
       System command failed!: bundle exec mutant --use rspec --include lib --require auom AUOM*
     # ./spec/support/corpus.rb:247:in `system'
```

mbj:
> by the way when try to run `bundle exec rake ci` locally, i always get this. am i doing it wrong?

If it exits 0 (which I think it does as CI runs this exact same command) its right, as this print out is from an integration test case that exists mutant can exit nonzero on uncoverage. I should probably silence it, its a source of confusion.

janosch-x:
@mbj JFYI it *does* exit with status 1 at the point shown above, making local testing pretty difficult. Seems to be due to me running Ruby 2.5.1. The build passes if I use Ruby 2.3.4 like circle ci.

PR #756, opened 2018-11-06
Fix Memoizable test

dkubb:


mbj:
@dkubb, @snusnu I see no value in the 2.3 and 2.4 runs, lets kill them. If people want to use older ruby, they can use older mutant. so we can keep AUOM corpus.

lib/mutant/matcher/method/instance.rb
```diff
@@ -12,10 +12,12 @@ class Instance < self
         # @param [UnboundMethod] method
         #
         # @return [Matcher::Method::Instance]
+        #
+        # :reek:ManualDispatch
         def self.new(scope, target_method)
           name = target_method.name
           evaluator =
-            if scope.include?(Memoizable) && scope.memoized?(name)
+            if scope.respond_to?(:memoized?) && scope.memoized?(name)
```
mbj: Given mutant has a wide user base, and `memoized?` is a rellative common method name, should we test for both `include?(Memoizable) && respond_to?(:memoized?)``

dkubb: @mbj I wasn't able to kill that without creating a completely synthetic test case.

mbj: @dkubb I'm fine with a synthetic test case. This class has verbose unit test coverage, adding this case should be possible. I'm fine to do such a fixup to this commit.

Changelog.md
```diff
@@ -1,3 +1,7 @@
+# v0.8.19 2018-11-06
+
+* Fix bug when mutating a module that includes Memoizable
```
mbj: Lets also mention we dropped Ruby < 2.5 support.

dkubb: @mbj Done.

PR #781, opened 2018-11-28
Fix wording for integration test run through example

mbj:


jschirp:
I pushed some commit to fix some spelling problems. Please squash if you agree.

mbj:
@jschirp this is ready for your next pass.

PR #789, opened 2018-11-30
Improve generation invariants

mbj:


spec/support/corpus.rb
```diff
@@ -129,33 +129,61 @@ def checkout
...
+      # Check generation invariants
+      #
+      # @param [Parser::AST::Node] original
+      # @param [Parser::AST::Node] mutation
+      #
+      # @return [undefined]
+      #
+      # @raise [Exception]
+      def check_generation_invariants(original, mutation)
+        return unless ENV['MUTANT_CORPUS_EXPENSIVE']
```
mbj: Doing this per mutation is to slow. And having an ENV based flip sucks. This all hints to create a first class verification CLI.

Also cannot wait for my 2990WX to ship so I can parallelize this much more.

PR #973, opened 2019-10-28
Fix memoized method subjects to preserve freezer option

### LINKED ISSUE #639: Fix memoizable insertion to preserve freezer option
Mutant is right now configured to detect and re-produce memoized methods via the `memoziable` gem. In case the memoizable is used with `adamantium` the additional `freezer` option is currently silently dropped.

Resulting in an unintend behavior change observable in some noop failures. While certainly an interesting mutation, this mutation should be inserted explicitly, but not inside the noops.

snusnu:
Close #639

### DIFF
--- a/spec/unit/mutant/subject/method/instance_spec.rb
+++ b/spec/unit/mutant/subject/method/instance_spec.rb
@@ -116,7 +116,6 @@ RSpec.describe Mutant::Subject::Method::Instance::Memoized do
   end
 
   describe '#prepare' do
-
     let(:context) do
       Mutant::Context.new(scope, double('Source Path'))
     end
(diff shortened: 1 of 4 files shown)


snusnu:
@mbj Ready for review. I see CI is failing integration tests but I'm not sure if you need me to recurse into these.

mbj:
@snusnu Code wise I'm fine, but the CI rusted. Hence it has to be fixed first.

spec/unit/mutant/subject/method/instance_spec.rb
```diff
@@ -137,7 +153,7 @@ def foo; end
     it_should_behave_like 'a command method'
   end
 
-  describe '#mutations', mutant_expression: 'Mutant::Subject#mutations' do
+  describe '#mutations' do
```
mbj, line 156: @snusnu The fix for the last alive mutation was removing this overly specific overwrite!

PR #975, opened 2019-11-06
Fix a warning due to an unset ivar

michaelherold:
It looks like the instance variable name here was changed at some point
during the design and that the check/warn is now looking at the
incorrect variable, resulting in a warning for a use of an unset
instance variable.


mbj:
@michaelherold I'm closing this as I solved it differently in #981. Thank you for letting me know.

PR #982, opened 2020-01-01
Fix missing node types in generic mutator

mbj:


lib/mutant/mutator/node/generic.rb
```diff
@@ -24,13 +25,19 @@ class Generic < self
           kwrestarg
           kwsplat
           match_with_lvasgn
+          meth_ref
```
mbj: So sad parser implemented that one and Ruby meanwhile reverted :(

PR #1013, opened 2020-08-02
Add dedicated error message on mutant-license-0.0.0 from rubygems

mbj:


mbj:
> Or it might be better to not use a regex at all since this is a simple match.

I forgot that ruby has a substring match. Will followup.

PR #1018, opened 2020-08-23
Remove boolean -> `nil` mutations

dgollahon:
Removes the following mutations:
- `true` -> `nil`
- `false` -> `nil`

Rationale:
- Since `true` is already mutated to `false` and vice versa, we already have a truthiness change and `nil` does not provide anything else in that direction. It is very unlikely to get a mutation from mutating to `nil` that we do not get from inverting the boolean.
- `nil` is actually the less primitive construct and easier to misuse.
  - nil.methods - false.methods # => [:to_a, :to_i, :to_f, :to_h, :to_r, :rationalize, :to_c]
    This allows for more cases where a `nil` is somehow accidentally turned into an empty hash, array, 0 value, etc.
  - false.methods - nil.methods # => []
  - false.methods == true.methods # => true
- It forces `nil` to be present in cases that are extremely hard to assert (inside conditionals that do not return the boolean expression) or, even if they are possible to assert, it is often awkward to do so. As a motivating example I like to use hashes as an asserted case statement (ex: `if { 'confirmed' => true, 'denied' => false }.fetch(enum_value) ...` but `mutant` would force me to make the value of the `'denied'` key to `nil` which is not an improvement in semantics or readability and very difficult to assert if the result of that expression is never part of a public API.
- It also results in fewer mutations being generated which is beneficial for CI runs and faster local feedback loops.
- This change also has the slight side benefit of removing more code from `mutant` than it adds.

@mbj Let me know what you think. I figured it was probably easier to just go ahead an open a PR for discussion rather than create an issue since it was a fairly easy change to make.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+- Remove `true` -> `nil` and `false` -> `nil` mutations. [#1018](https://github.com/mbj/mutant/pull/1018)
+
 # v0.9.8 2020-08-02
 
 * Change to generic catch all node mutator. This allows better cross parser version compatibility.
--- a/lib/mutant/mutator/node/literal/boolean.rb
+++ b/lib/mutant/mutator/node/literal/boolean.rb
@@ -20,7 +20,6 @@ module Mutant
           #
           # @return [undefined]
           def dispatch
-            emit_nil
             emit(s(MAP.fetch(node.type)))
           end
 
--- a/meta/and.rb
+++ b/meta/and.rb
@@ -7,8 +7,6 @@ Mutant::Meta::Example.add :and do
   mutation 'true'
   mutation 'false'
   mutation 'true or false'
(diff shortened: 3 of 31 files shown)


dgollahon:
Updated the changelog.

mbj:
To put a bit of a more axiomatic reasoning behind this change:

* Lets assume a function that currently returns `true` or `false` in haskell terms is of type `Bool`.
* A mutation to `nil` would turn it into `Maybe Bool`.

A mutation of the type signature from `Bool` to `Maybe Bool` is adding *more* possible states to the return value, which is NOT the point of mutation testing (in mutants flavor). Mutnat is about:

* Semantic reductions (reducing the observable value / semantic space strictly)
* Orthogonal replacement

The exusting mutaiton from `true -> nil` and `false -> nil` violates this property.

mbj:
@dgollahon merged in repush as #1012

PR #1020, opened 2020-08-24
Remove non canonical relational operator mutations

mbj:


mbj:
This change was released in 0.9.9.

PR #1027, opened 2020-08-25
Fix endless range mutations

### LINKED ISSUE #1026: Endless ranges crash mutant
Simple reproduction case I tried:

```ruby
class Test
  def self.slice_after(array, index)
    array[index..]
  end
end
```

Reproduction spec:

```ruby
RSpec.describe Test do
  it 'works' do
    expect(described_class.slice_after([1,2], 1)).to eql([2])
  end
end
```

Executing `mutant --use rspec Test` produces a long trace ending in
> 1: from /Users/gollahon/.rvm/gems/ruby-2.6.6/gems/mutant-0.9.9/lib/mutant/mutator/node.rb:42:in `mutate_child'
/Users/gollahon/.rvm/gems/ruby-2.6.6/gems/mutant-0.9.9/lib/mutant/mutator.rb:18:in `mutate': undefined method `type' for nil:NilClass (NoMethodError)

Versions
`mutant-rspec` version: 0.9.9
`unparser` version: 0.4.7

mbj:
[Fix #1026]

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ['2.5', '2.6', '2.7']
+        ruby: ['2.6', '2.7']
         os: [ubuntu-latest]
     steps:
     - uses: actions/checkout@v2
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,6 @@
 # v0.9.10 unreleased
 
+* Fix crash on endless ranges [#1026](https://github.com/mbj/mutant/pull/1026).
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -35,7 +35,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ['2.5', '2.6', '2.7']
+        ruby: ['2.6', '2.7']
```
mbj, line 38: For now mutant internally always use the ruby parser that is based on the current ruby version.

Attempting to mutation cover branches that are only required for `> 2.5` would fail. Instead mutant needs a mechanism of a "target ruby". Than even 2.5 could pass the mutation test on > 2.5 syntax.

For now this is the easier way to get the bugfix in.

Issue #172, opened 2014-04-08
Add a mutation from Array(value) to value and [value].


mbj:

mbj:
We do have a generic `method(foo)` to -> `foo` so we can mark this as closed.

PR #1039, opened 2020-08-26
Mutate Array(foo) -> [foo]

dgollahon:
- See #172. While #172 was closed outright, I think the rationale behind closing it is incorrect since the generic mutation there does not cover the proposed case. The benefit of using `[foo]` over `Array(foo)` if you have only scalar values is that you are using the simpler/narrower tool.


dgollahon:
Closing in favor of #1043 (resolves CI issue).

PR #1040, opened 2020-08-26
Mutate dynamic send to static send

dgollahon:
Mutates:
- `foo.__send__(:bar)` -> `foo.bar`
- `foo.send(:bar)` -> `foo.bar`
- `foo.public_send(:bar)` -> `foo.bar`

...and so on, including arguments.

Closes #732

PR #1041, opened 2020-08-26
Add pull request CI trigger

mbj:


dgollahon:
@mbj The strange thing about the CI stuff is if you take #1040 from my perspective it looks like it hasn't ever built... but then if I go to the associated issue, #732, I can see that the build built and is green. I'm not sure if that information might affects the solution here.

PR view:
image

Associated issue view:
image

mbj:
I'll close this for the moment as @dgollahon is a collaborator now.

Still have to figure out a better setup, I'd love to see 3rd party pull requests being build *while* pushes (to new branches) get build before I make a PR *without* duplicating the build on having both `push` and `pull_request` as hooked events.

PR #1043, opened 2020-08-26
Mutate Array(foo) -> [foo]

dgollahon:
- See #172. While #172 was closed outright, I think the rationale behind closing it is incorrect since the generic mutation there does not cover the proposed case. The benefit of using `[foo]` over `Array(foo)` if you have only scalar values is that you are using the simpler/narrower tool.

### DIFF
--- a/meta/send.rb
+++ b/meta/send.rb
@@ -660,3 +660,39 @@ Mutant::Meta::Example.add :send do
   mutation 'foo(self)'
   mutation 'foo(nil)'
 end
+
+Mutant::Meta::Example.add :send do
+  source 'Array(a)'
+
+  singleton_mutations
+  mutation 'a'
+  mutation 'Array()'
+  mutation 'Array(nil)'
+  mutation 'Array(self)'
+  mutation '[a]'
+end
+
+Mutant::Meta::Example.add :send do
+  source 'Kernel.Array(a)'
+
+  singleton_mutations
+  mutation 'a'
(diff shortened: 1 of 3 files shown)


meta/send.rb
```diff
@@ -598,3 +598,39 @@
   mutation 'foo(self)'
   mutation 'foo(nil)'
 end
+
+Mutant::Meta::Example.add :send do
+  source 'Array(a)'
+
+  singleton_mutations
+  mutation 'a'
+  mutation s(:send, nil, :Array)
```
mbj, line 607: "normally" using AST input for specifying a migration is a sign of a bug (or an unparser issue).

I've to research the reason here.

dgollahon: I am not sure why I put it that way but I have fixed it.

mbj: I suspect at the time there was an unparser issue, or it was a debugging leftover.

mbj:
@dgollahon I agree with mutation and code. I've to go AFK, and will merge on green after rebase. Maybe you beat me to the rebase.

mbj:
> * See #172. While #172 was closed outright, I think the rationale behind closing it is incorrect since the generic mutation there does not cover the proposed case. The benefit of using `[foo]` over `Array(foo)` if you have only scalar values is that you are using the simpler/narrower tool.

I agree that #172 was closed unjustified.

mbj (approved):

PR #1044, opened 2020-08-26
Mutate dynamic send to static send

### LINKED ISSUE #732: Add mutation from `send(:foo) -> foo`. (with arguments)

dgollahon:
Mutates:
- `foo.__send__(:bar)` -> `foo.bar`
- `foo.send(:bar)` -> `foo.bar`
- `foo.public_send(:bar)` -> `foo.bar`

...and so on, including arguments.

Closes #732

### DIFF
--- a/meta/send.rb
+++ b/meta/send.rb
@@ -308,6 +308,68 @@ Mutant::Meta::Example.add :send do
 end
 
 Mutant::Meta::Example.add :send do
+  source '__send__(:foo)'
+
+  singleton_mutations
+  mutation 'foo'
+  mutation '__send__'
+  mutation 'public_send(:foo)'
+  mutation ':foo'
+  mutation '__send__(nil)'
+  mutation '__send__(self)'
+  mutation '__send__(:foo__mutant__)'
+end
+
+Mutant::Meta::Example.add :send do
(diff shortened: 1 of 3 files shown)


meta/send.rb
```diff
@@ -307,6 +307,68 @@
...
+  mutation 'foo'
+  mutation ':bar'
+  mutation 'self.send(:bar)'
+  mutation 'foo.send(nil)'
+  mutation 'foo.send(self)'
+  mutation 'foo.send(:bar__mutant__)'
+end
+
+Mutant::Meta::Example.add :send do # rubocop:disable Metrics/BlockLength
+  source 'foo.public_send(:bar, 1, two: true, **kwargs, &block)'
```
mbj, line 340: We need to prove that we forward "all kinds of arguments" hence yes this test case is "warranted". Even if it exposes us to secondary mutations in this tests.

I was thinking about a strategy to filter the secondaries, but it depends on making the AST matching language. for now we have to eat it.

mbj (approved):

meta/send.rb
dgollahon: #415 Might also be a good partial solution if you would like me to investigate adding something like that.

PR #1045, opened 2020-08-27
Guard against mutations to loop { }

### LINKED ISSUE #444: Add guard against mutation providing `loop do; end`
With [reference to this discussion](https://github.com/mbj/mutant/issues/443#issuecomment-140130642)

dgollahon:
- Closes #444

### DIFF
--- a/lib/mutant/mutator/node/block.rb
+++ b/lib/mutant/mutator/node/block.rb
@@ -21,7 +21,7 @@ module Mutant
         end
 
         def mutate_body
-          emit_body(nil)
+          emit_body(nil) unless unconditional_loop?
           emit_body(N_RAISE)
 
           return unless body
@@ -29,6 +29,10 @@ module Mutant
           emit_body_mutations
 
           mutate_body_receiver
+        end
+
+        def unconditional_loop?
+          send.eql?(s(:send, nil, :loop))
         end
 
         def body_has_control?
--- a/meta/block.rb
+++ b/meta/block.rb
@@ -204,3 +204,13 @@ Mutant::Meta::Example.add :block do
   mutation 'foo(nil, &nil).baz {}'
   mutation 'foo(nil, &:bar__mutant__).baz {}'
 end
+
+Mutant::Meta::Example.add :block do
+  source 'loop { true }'
+
+  singleton_mutations
+  mutation 'true'
+  mutation 'loop { false }'
+  mutation 'loop { raise }'
+  mutation 'loop'
+end


dgollahon:
Hm, I'll have to come back to this and look at what's not covered / breaking the mutation tests. I was hopeful based on the unit tests.

lib/mutant/mutator/node/block.rb
```diff
@@ -31,6 +31,10 @@ def mutate_body
           mutate_body_receiver
         end
 
+        def unconditional_loop?
+          send.eql?(s(:send, nil, :loop))
+        end
```
dgollahon, line 36: @mbj Something weird is going on here. Somehow this formulation breaks something in `NodePredicates` for reasons I do not yet understand. I have found a couple of different ways of expressing the same idea that do work though:

```ruby
def unconditional_loop?
  _, method_name = *send
  method_name.equal?(:loop)
end
```

or

```ruby
def unconditional_loop?
  children.first.eql?(s(:send, nil, :loop))
end
```

Seems like something about the `send` child definition under mutation testing is causing a problem. I am happy to use a workaround such as above but it seems like there is a subtle problem somewhere and could cause other issues if it's not clearly understood so I'm leaving my original (failing) implementation here. LMK if you have any ideas. No rush though, I may not have time to come back to this for a couple of days.

mbj:
@dgollahon Just to put this into perspective. Mutant *will* create infinite loops, always.

There are many forms of infinite loops. And we can try blacklisting as much was we want we'll *always* have them.

I'm fine in removing mutations to *obvious* infinite loops. Still mutant needs native timeouts to correctly handle them generically.

I support this PR, but wanted to raise my thoughts on the problem class here.

dgollahon:
@mbj Yeah, fully understood and agreed. I do still think it is still beneficial for the end user to remove them in obvious/simple cases as it reduces runtime, especially if you have slower tests and have to have a wider timeout.

lib/mutant/mutator/node/block.rb
dgollahon: @mbj Did you have any insights on this failure / seen anything like it before? I can try to look into it more or just use a workaround but I was pretty surprised to the error only appearing in the `mutant` run on CI.

dgollahon: @mbj Curious if you had any insights on this, when you get a chance.

dgollahon: @mbj Strangely I can no longer reproduce this. Maybe something subtle got fixed since when I first opened this? Unfortunately I can't look back at the CI logs since they are gone now and I don't know how to reproduce the error. I think this is ready for review again.

meta/block.rb
```diff
@@ -204,3 +204,13 @@
...
   mutation 'foo(nil, &:bar__mutant__).baz {}'
 end
+
+Mutant::Meta::Example.add :block do
+  source 'loop { true }'
+
+  singleton_mutations
+  mutation 'true'
+  mutation 'loop { false }'
+  mutation 'loop { raise }'
```
mbj, line 214: In theory this one is also quite redundant, as it always raises. its equivalent to `raise` in observable semantics.

The list of these cases is endless, its not a blocker. Just a thougt.

Ultimatively: If we had a symbolic execution engine, we could find "static execution path of known result" and filter these cases out statically without casing them.

This commit removes some (non harmful noise), which is a small win I like to see.

But ultimatively we'd add (never ending) implementation noise in filtering this noise out. I'm thinking about biting the bullet and making a symbolic execution engine for Ruby...

mbj (approved):

meta/block.rb
dgollahon: Yeah. I would like to do some "lightweight" filtering of some of these obvious cases but we don't want to go too far in the weeds because there is no bottom here. I think this loop case is one of the better ones to filter because at least the `raise` case you've highlighted fails instantly and doesn't trigger a timeout. Before you have to enable coverage via timeouts and it takes as long as your timeout setting.

But yes, there is definitely a balance to strike here in terms of implementation complexity and how many mutations we are able to cull.

PR #1047, opened 2020-09-06
Add block-pass mutations

dgollahon:
- Adds `foo(&method(:bar))` -> `foo(&public_method(:bar))` and similar mutations.


meta/block.rb
```diff
@@ -184,6 +184,9 @@
   mutation 'foo.baz { }'
   mutation 'foo(&:bar).baz'
   mutation 'self.baz {}'
+  mutation 'foo(&nil).baz {}'
+  mutation 'foo(&self).baz {}'
```
dgollahon, line 188: We could optionally skip `&nil` and `&self` since I think they are unlikely to yield alives but they are technically valid.

mbj: I think this skip is valid and should be done in a followup. Does not block this PR.

mbj: To elaborate on why: They are "static" expressions. The outcome of `&nil` and `&self` is known ahead of time to not be useful.

mbj (approved):

PR #1050, opened 2020-09-08
Remove unused `let` statements

dgollahon:
- Various cases of `let` statements that are never referenced, as discovered with my little [rspectre](https://github.com/dgollahon/rspectre) tool.

### DIFF
--- a/spec/unit/mutant/bootstrap_spec.rb
+++ b/spec/unit/mutant/bootstrap_spec.rb
@@ -2,7 +2,6 @@
 
 RSpec.describe Mutant::Bootstrap do
   let(:integration)          { instance_double(Mutant::Integration)   }
-  let(:integration_name)     { instance_double(String)                }
   let(:integration_result)   { Mutant::Either::Right.new(integration) }
   let(:kernel)               { instance_double(Object, 'kernel')      }
   let(:load_path)            { %w[original]                           }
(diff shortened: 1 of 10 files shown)


mbj:
I'm suprised that a code base as old as mutant only has "that few" of these.

I'd be happy to hook up your tool into CI.

mbj (approved):

dgollahon:
> I'm suprised that a code base as old as mutant only has "that few" of these.

That is probably mostly due to the fact that there was already a cleanup pass a few years ago: https://github.com/mbj/mutant/pull/640

> I'd be happy to hook up your tool into CI.

We could look into that. In my experience so far it seems to work pretty well as a periodic cleanup tool since only a small amount seems to accumulate for well-run codebases. It can slow down CI a little bit because it needs to run over unit + integration tests to have reliable results since `let`s and `shared_context`s can be shared across various kinds of tests. In the future it might be possible to adjust that or parallelize testing but for now I suspect it would be a bottleneck for most codebases.

I recently made `rspectre` a little more thorough/robust so I used `mutant` as one of my test codebases. :)

mbj:
> . In my experience so far it seems to work pretty well as a periodic cleanup tool since only a small amount seems to accumulate for well-run codebases.

If there is extra cost to the CI, lets keep it manual. But feel free to commit a script we can run locally.

PR #1062, opened 2020-10-16
2.7 Syntax - Upgrade to unparser ~> 0.5.2

mbj:
* Remove all hacks for unparser normalization
* Adjust verification to enforce the better unparser constraints
* Add 2.7 syntax support


lib/mutant/subject/method/instance.rb
```diff
@@ -42,7 +42,7 @@ def prepare
         private
 
           def wrap_node(mutant)
-            s(:begin, mutant, s(:send, nil, :memoize, s(:args, s(:sym, name), *options)))
```
mbj, line 45: @snusnu All this years we where generating *invalid AST* that would silently be emitted as correct intended source.

meta/begin.rb
```diff
@@ -13,8 +13,7 @@
 end
 
 Mutant::Meta::Example.add :begin do
+  source '(true)'
 
-  source s(:begin, s(:true))
-  # Mutation of each statement in block
-  mutation s(:begin, s(:false))
```
mbj, line 19: @dgollahon ^^ was part of the unparser preprocessor hack artifacts.

meta/float.rb
```diff
@@ -8,9 +8,9 @@
   # edge cases
   mutation '0.0'
   mutation '1.0'
-  mutation '(0.0 / 0.0)'
-  mutation '(1.0 / 0.0)'
-  mutation '(-1.0 / 0.0)'
```
mbj, line 13: @dgollahon ^^ Another one. This redundant begin was needed as unparser pre 0.5 was not able to infer `or` vs `||` based on AST structure and had to preserve semantic equivalency via extra parens. This extra parsens where an artifact of that.

meta/lvar.rb
```diff
@@ -1,16 +1,10 @@
 # frozen_string_literal: true
 
 Mutant::Meta::Example.add :lvar do
-  source 'a = nil; a'
+  declare_lvar :a
```
mbj, line 4: Not really an unparser artifact. But correctly using the parser API to avoid redundancy.

meta/lvar.rb
```diff
@@ -1,16 +1,10 @@
...
 
-  mutation 'a = nil; nil'
-  mutation 'a = nil; self'
-  mutation 'a = nil'
-  # TODO: fix invalid AST
-  #   These ASTs are not valid and should NOT be emitted
-  #   Mutations of lvarasgn need to be special cased to avoid this.
-  mutation s(:begin, s(:lvasgn, :a__mutant__, s(:nil)), s(:lvar, :a))
```
mbj, line 12: No need anymore to specify AST with forwad lvar declaration.

meta/op_assgn.rb
```diff
@@ -1,17 +1,42 @@
...
+
+  mutation 'a.b += -1'
+  mutation 'a.b += 0'
+  mutation 'a.b += 2'
+  mutation 'a.b += nil'
+  mutation 'a.b += self'
   mutation 'self.b += 1'
-  # TODO: fix invalid AST
-  #   This should not get emitted as invalid AST with valid unparsed source
-  mutation s(:op_asgn, s(:ivar, :@a), :+, s(:int, 1))
```
mbj, line 16: Side effect of better unparser we can remove that one now.

meta/float.rb
dgollahon: This one in particular seems like a nice quality of life improvement.

mbj: Yeah, the unparser preprocessor hack was a bad shortcut of mine I took years ago when I had "nothing" and had to 80/20 a lot of subsystems to get something.

spec/unit/mutant/meta/example/dsl_spec.rb
```diff
@@ -55,19 +49,29 @@ def self.expect_error(message, &block)
...
+      let(:source) { 'false' }
+
+      let(:expected) do
+        [
+          Mutant::Meta::Example::Expected.new(
+            node:            s(:nil),
+            original_source: 'nil'
+          )
+        ]
+      end
```
dgollahon, line 74: This is kind of interesting. I'm not sure I fully understand the context of this spec, but does this mutate `false` -> `nil` somehow? I removed that mutation recently (I thought). Or is this just an artifact of test setup but wouldn't actually happen in practice?

mbj: This spec does not demonstrate the mutations. It never calls the mutation engine, AKA `Mutant::Mutator::Node` is never called. Instead it specifies on how mutations are specified. For this I run through the scenarios:

* There was a mutation being generated that was not expected.
* A mutation that was expected was not being generated.
* A mutation that was generated unparses to invalid syntax
* The original code that was being specified did not round trip.

And for each of these cases above I set-up an `Mutant::Meta::Example` object in that specific state. But I intentionally do NOT run the engine here. But still use AST nodes that can round trip to unparser at low complexity.

mbj: or TLDR: This code is the specification of the mutant mutation specifications and uses fake mutated nodes to not hard code the specification of properties of the specification against the mutation engine.

dgollahon: > or TLDR: This code is the specification of the mutant mutation specifications and uses fake mutated nodes to not hard code the specification of properties of the specification against the mutation engine.

Ok, cool. I thought that might be the case but wasn't 100% sure at a glance. Might make sense to use a valid mutation for documentation consistency but probably doesn't matter much.

mbj: >  Might make sense to use a valid mutation for documentation consistency but probably doesn't matter much.

it is a valid mutation example to demonstrate the specification of the specification on the properties I listed above. And I do not wish to call the real mutation engine here. I only wish to demonstrate I can find specific expected, missing and invalid mutations.

dgollahon: Right, I just meant I usually like to have examples in my tests match "real" behavior as much as possible. It is clearly valid in this context to mutate to anything but I thought it might be slightly less confusing to match the "real" behavior. Doesn't actually matter though, feel free to ignore.

mbj: If I where to start having to manually have generated mutations in sync here, I'd enforce something that adds no value to the test. And even would break as mutations are about to change drastically.

I'm experimenting a lot with "tiers" of mutations to default to only the most useful etc.

PR #1068, opened 2020-10-18
Add windows builds

mbj:


mbj:
Heh, WSL is not active by default and I have no desire to hack on that yet.

PR #1096, opened 2020-11-11
Change process model

mbj:
This will include external timeouts. A work in progress.


mbj:
@dgollahon, @pawelpacana This branch passes integration specs but not yet unit/mutation coverage.

Still it should optimize large runs significantly. I even see positive impact on smaller normal runs.

The key difference is that the main processes memory that grows over time when accumulating the result does NOT get inherited anymore into the individual killforks.

If you have time feel free to give this branch an early trial on your projects.

PR #1107, opened 2020-11-16
Document config files

gorgonshead:
Adds documentation for mutant's config file format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,7 @@ I'm wrong I'd be happy to improve mutant to the point I'm right again.
 * [Concurrency](/docs/concurrency.md)
 * [Rspec Integration](/docs/mutant-rspec.md)
 * [Minitest Integration](/docs/mutant-minitest.md)
+* [Configuration](/docs/configuration.md)
 
 ## Communication
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1,0 +1,61 @@
+# Configuration
+
+Mutant can be configured with a config file that can be named one of the following: `.mutant.yml`, `config/mutant.yml`, or `mutant.yml`
+
+The following options can be configured through the config file:
+
+#### `includes`
+
+The includes key takes a list of strings to add to ruby's `$LOAD_PATH`. This is similar to ruby's `-I` option.
+
+```yml
+---
+includes:
+  - lib
+```
+
+Additional includes can be added by providing the `-I` or `--include` option to the CLI.
+
+#### `requires`
+
+The requires key takes a list of ruby files to be required. This is how mutant loads your application code.
+
+```yml
+---
+requires:
+  - my_app
+```
+
+Additional requires can be added by providing the `-r` or `--require` option to the CLI.
+
+#### `integration`
+
+Specifies which mutant integration to use. If your tests are writen in [RSpec](https://rspec.info/), this should be set to `rspec`. If your tests are written in [minitest](https://github.com/seattlerb/minitest), this should be set to `minitest`.
+
+```yml
+---
+integration: rspec
+```
+
+The integration can be overridden by providing the `--use` option to the CLI.
+
+#### `fail_fast`
+
+When `fail_fast` is enabled, mutant will stop as soon as it encounters an alive mutation. This can be helpful for incremental workflows where you want to see the output of a failed mutation immediately. Valid values are `true` and `false`.
+
+```yml
+---
+fail_fast: true
+```
+
+#### `jobs`
+
+Specify how many processes mutant uses to kill mutations. Defaults to the number of processors on your system.
+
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -158,6 +158,7 @@ I'm wrong I'd be happy to improve mutant to the point I'm right again.
 * [Concurrency](/docs/concurrency.md)
 * [Rspec Integration](/docs/mutant-rspec.md)
 * [Minitest Integration](/docs/mutant-minitest.md)
+* [Configuration](/docs/confiuration.md)
```
telyn, line 161: The path has a typo - missing `g`

docs/configuration.md
```diff
@@ -0,0 +1,65 @@
...
+```yml
+---
+fail_fast: true
+```
+
+`fail_fast` cannot currently be overridden by the CLI.
+
+#### `jobs`
+
+Specifiy how many processes mutant uses to kill mutations. Defaults to the number of processors on your system.
```
telyn, line 55: another small typo at the beginning of the line `s/Specifiy/Specify/`

telyn (commented):
All looks good to me, except a couple of small typos.

docs/configuration.md
```diff
@@ -0,0 +1,65 @@
...
+#### `fail_fast`
+
+When `fail_fast` is enabled, mutant will stop as soon as it encounters an alive mutation. This can be helpful for incremental workflows where you want to see the output of a failed mutation immediately. Valid values are `true` and `false`.
+
+```yml
+---
+fail_fast: true
+```
+
+`fail_fast` cannot currently be overridden by the CLI.
```
mbj, line 51: mhh, this is not my intention, likely a bug I've to fix.

docs/configuration.md
```diff
@@ -0,0 +1,65 @@
...
+
+```yml
+---
+jobs: 8
+```
+
+The number of jobs can by overridden by the `-j` or `--jobs` flag in the CLI.
+
+See mutant's configuration file, [mutant.yml](/mutant.yml), for a complete example.
+
```
mbj, line 65: Extra whitespace here.

docs/configuration.md
mbj: it can, but it can only be turned from `false` in the config file (which is semantically redundant, as its the default anyway) to `true` in the CLI.

docs/configuration.md
```diff
@@ -0,0 +1,65 @@
...
+#### `jobs`
+
+Specifiy how many processes mutant uses to kill mutations. Defaults to the number of processors on your system.
+
+```yml
+---
+jobs: 8
+```
+
+The number of jobs can by overridden by the `-j` or `--jobs` flag in the CLI.
```
snusnu, line 62: `s/can by/can be/`

docs/configuration.md
dgollahon: > it can, but it can only be turned from false in the config file (which is semantically redundant, as its the default anyway) to true in the CLI.

Is this how you intend it to work? Or should there be a `--fail-fast false` or `--no-fail-fast` support? I realized this when I was showing @gorgonshead how it worked. You're right that it can go from `false` -> `true` though.

mbj: @dgollahon I think that we'd want to support `--no-fail-fast` to flip it back. For the moment this PR should state: 

> `fail_fast: true` in the config file cannot be undone in the CLI

docs/configuration.md
mbj: Technically, if a CLI argument requires another argument its not a flag. `-j` is not self standing, its an option. Hence this should state:

```
> The number of jobs can be ovcerwritten by the `-j/--jobs` option in the CLI.

The word `flag` AFAIK describes `--version` and friends that consume no other argument.

dgollahon: I have used / sometimes seen them used interchangeably but that's an interesting distinction. 👍

mbj:
@gorgonshead Thanks for your contribution! And sorry your first PR got "swarmed" by people. No single mind is sharp enough to find all these things by its own and we do development here in the open so we can reach the aggregate of many peoples local optimum.

I can tell you out of knowing each person who flagged here personally: They are all well meaning.

Would you be willing to revise your PR and fix the flags? Than ping me again, as I've processed your CLA I'm ready to merge this once its flag free.´

gorgonshead:
Thanks @mbj, appreciate the support. Working on it now.

gorgonshead:
I think I've addressed all the feedback and I'm ready for more.

mbj (approved):

mbj:
@gorgonshead Definitively an improvement. Merging. There is a lot where I want to expand on it, but we have to start somewhere!

Issue #1105, opened 2020-11-15
Handle exceptions during mutation generation gracefully


mbj:
Its the sad truth that its very unlikely that the mutation generation engine will at some point run into runtime exceptions from visiting AST structures it does not expect.

In this case the subject needs to be reported separately but not terminate the entire run.

PR #1108, opened 2020-11-16
Fix crash on static send mutation

mbj:


dgollahon:
Nice catch. I think I must have introduced that one.

mbj:
Yeah, but I reviewed it. So no blame assigned.

BTW: I think we need https://github.com/mbj/mutant/issues/1105 to handle these more gracefully, mutants AST is "to wide" to assume a dynlang can ever be correctly generate mutations in all cases. With types this would not happen.

dgollahon:
We could probably do some automated testing of removing arguments from sends, etc. in the meta and making sure they don't crash. I suspect there are more cases like this generally.

Yeah, we very well might. My only concern about #1105 is that it will probably make it hard to receive bug reports since most users won't notice failures. That's still probably worth it, but a little unfortunate.

PR #1112, opened 2020-11-21
Add timeout result configuration

mbj:


dgollahon:
> Add bundller cache setup on CI

FYI `bundller` is a typo.

mbj:
> FYI `bundller` is a typo.

thx. fixed.

lib/mutant/isolation/none.rb
```diff
@@ -12,12 +12,25 @@ class None < self
       #
       # @return [Result]
       #
-      # ignore :reek:UtilityFunction
+      # rubocop:disable Lint/SuppressedException
```
mbj, line 15: Rubocop has something to learn here. `=> expression` is an assignment. Unparser also has to learn this to reproduce lvar scope specific syntax.

Issue #1054, opened 2020-09-22
Stack level too deep when mutating included hook


pawelpacana:
We have recently switched from running mutant on Ruby 2.6 to 2.7. On the code that was not changed, and the same mutant version there are different results:

```
bundle list |grep mutant                                                                                                                  
  * mutant (0.9.12)
  * mutant-license (0.1.0)
  * mutant-rspec (0.9.10)
```

### 2.6.6p146

```
Mutant environment:
Matcher:         #<Mutant::Matcher::Config match_expressions: [AggregateRoot*]>
Integration:     rspec
Jobs:            8
Includes:        ["lib"]
Requires:        ["aggregate_root"]
Subjects:        27
Total-Tests:     35
Selected-Tests:  31
Tests/Subject:   1.15 avg
Mutations:       985
Results:         985
Kills:           985
Alive:           0
Runtime:         18.61s
Killtime:        107.46s
Overhead:        -82.69%
Mutations/s:     52.94
Coverage:        100.00%
```

### 2.7.0p0

```
Mutant environment:
Matcher:         #<Mutant::Matcher::Config match_expressions: [AggregateRoot*]>
Integration:     rspec
Jobs:            8
Includes:        ["lib"]
Requires:        ["aggregate_root"]
Subjects:        27
Total-Tests:     35
Selected-Tests:  31
Tests/Subject:   1.15 avg
Mutations:       985
Results:         985
Kills:           981
Alive:           4
Runtime:         17.27s
Killtime:        84.41s
Overhead:        -79.55%
Mutations/s:     57.05
Coverage:        99.59%

```

One of the failing mutations:

```
evil:AggregateRoot.included:/Users/pawelpacana/Code/rails_event_store/aggregate_root/lib/aggregate_root.rb:86:d3c17
@@ -1,4 +1,4 @@
 def self.included(host_class)
-  host_class.include(with_default_apply_strategy)
+  host_class.include(self)
 end
-----------------------
Killfork exited nonzero. Its result (if any) was ignored.
Process status:
#<Process::Status: pid 78659 SIGILL (signal 4)>
Killing the mutation resulted in an integration error.
This is the case when the tests selected for the current mutation
did not produce a test result, but instead an exception was raised.

This may point to the following problems:
* Bug in mutant
* Bug in the ruby interpreter
* Bug in your test suite
* Bug in your test suite under concurrency

The following exception was raised:


#<ArgumentError: marshal data too short>
[...]

PR #1118, opened 2020-11-22
Add timeout as configurable coverage criteria

mbj:


lib/mutant/config.rb
```diff
@@ -60,6 +35,32 @@ class Config
       mutant.yml
     ].freeze
 
+    private_constant(*constants(false))
+
+    class CoverageCriteria
+      include Anima.new(:timeout, :test_result)
```
mbj, line 41: @dgollahon The next step is to allow to configure the coverage criteria *per subject* in the config file an in-line.

Same for the mutation timeout.

The ideal setting than is a global "high" timeout with `coverage_criteria.timeout = false` and a per subject one after manual review (From an alive, example an infinite loop) setting it to true.

mbj: Also there are more possible criteria like `warnings` etc.

lib/mutant/reporter/cli/printer/coverage_result.rb
```diff
@@ -0,0 +1,19 @@
...
+  class Reporter
+    class CLI
+      class Printer
+        # Reporter for mutation coverage results
+        class CoverageResult < self
+          # Run report printer
+          #
+          # @return [undefined]
+          def run
+            visit(MutationResult, object.mutation_result)
```
mbj, line 13: This is a pass through ATM, I could evision a verbose mode where we show covered mutations, and tell why they where killed.

lib/mutant/reporter/cli/printer/subject_result.rb
```diff
@@ -7,7 +7,7 @@ class Printer
         # Subject result printer
         class SubjectResult < self
 
-          delegate :subject, :alive_mutation_results, :tests
+          delegate :subject, :uncovered_results, :tests
```
mbj, line 10: I'm slowly moving the wording away from "alive" and "killed" to "uncovered" and "covered".

lib/mutant/result.rb
```diff
@@ -206,25 +245,39 @@ class Mutation
...
       # Time the tests had been running
       #
       # @return [Float]
       def killtime
-        if isolation_result.success?
-          isolation_result.value.runtime
-        else
-          0.0
-        end
+        isolation_result.value&.runtime || 0.0
```
mbj, line 262: Its now perfectly valid there is no value returned from the isolation. And the status of the islation result cannot be determined anymore without the coverage criteria coming from the outside.

lib/mutant/config.rb
mbj: And the most important criteria for @pawelpacana counting ruby aborts as killed.

https://github.com/mbj/mutant/issues/1054.

Next major feature for 0.10.8. The infrastructure exists now.

Per subject config will be added in 0.10.9 possibly also in-line configuration.

docs/configuration.md
```diff
@@ -63,5 +63,38 @@ See mutant's configuration file, [mutant.yml](/mutant.yml), for a complete examp
...
+Use `timeout` setting under `coverage_criteria` in the config file to control
+if timeouts are allowed to cover mutations.
+
+#### `coverage_criteria`
+
+A configuration file only setting to control which criteria apply to determine mutation coverage.
+
+```yml
+---
+coverage_criteria:
```
dgollahon, line 88: Is it possible to set this via CLI flag? Once we have per-subject / per file config then customizing it that way would be good but I am thinking of doing something like having the fork abort coverage off in my local config (so I can find and report bugs) but considering it as a kill in CI so that PRs are never gated by segfaults.

mbj: Allowing to configure each subject separately is the next step. 

But: Its unlikely it makes sense in the CLI, simply because the CLI would become very long. My next initiative would be to support inheriting (prefix match) based per subject configuration in the code, and use the config file as input.

Than the overnext step to support it from in-line code comments.

Only after that I'll consider making CLI (per subject) support for these.

dgollahon: Ok, yeah, I think that makes sense. Looking forward to it 👍

PR #1125, opened 2020-11-26
Support multiple subject expressions [RSpec]

### LINKED ISSUE #1079: Support multiple coverage expressions in mutant-rspec
Support was recently added for `mutant-minitest` but per [this comment](https://github.com/mbj/mutant/pull/1076#issue-512312493)
> Rspec integrations could also be adapted. Feel free to open an issue for this in the future.

 `mutant-rspec` does not yet have support. This is a feature my team would use and I think it will be helpful to have this issue open to track it.

dgollahon:
- Allows mutant users to specify that a test covers multiple subjects using `mutant_expression`, not just one.
- Closes #1079

@mbj Is this what you had in mind in terms of implementation? I thought about concatenating the subject but then I'm not sure how you could override the default subject. I was thinking just an array of subjects would be a good starting point.

Additionally, I tried to mutate this locally but it didn't seem to pick up the subject with `--since` or a direct expression. I must be doing something wrong there--not sure if CI picked up on the changes or not.

### DIFF
--- a/mutant.yml
+++ b/mutant.yml
@@ -4,4 +4,5 @@ includes:
 integration: rspec
 requires:
 - mutant
+- mutant/integration/rspec
 - mutant/meta
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+* Add support for specifying multiple subject expressions with the RSpec integration. [#1125](https://github.com/mbj/mutant/pull/1125)
+
 # v0.10.8 2020-11-22
 
 * Add support for process abort as coverage condition.
--- a/lib/mutant/integration/rspec.rb
+++ b/lib/mutant/integration/rspec.rb
@@ -92,17 +92,22 @@ module Mutant
         }
 
         Test.new(
-          expressions: [parse_metadata(metadata)],
+          expressions: parse_metadata(metadata),
           id:          id
         )
       end
 
       def parse_metadata(metadata)
         if metadata.key?(:mutant_expression)
-          parse_expression(metadata.fetch(:mutant_expression))
+          expression = metadata.fetch(:mutant_expression)
+
+          expressions =
+            expression.instance_of?(Array) ? expression : [expression]
+
+          expressions.map(&method(:parse_expression))
         else
           match = EXPRESSION_CANDIDATE.match(metadata.fetch(:full_description))
-          parse_expression(match.captures.first) { ALL_EXPRESSION }
+          [parse_expression(match.captures.first) { ALL_EXPRESSION }]
         end
       end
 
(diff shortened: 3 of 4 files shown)


dgollahon:
> Additionally, I tried to mutate this locally but it didn't seem to pick up the subject with --since or a direct expression. I must be doing something wrong there--not sure if CI picked up on the changes or not.

CI [doesn't seem to be generating other mutations either](https://github.com/mbj/mutant/pull/1125/checks?check_run_id=1460853636). Am I missing something here?

mbj:
@dgollahon I'm fine with your choices of overwriting. Code wise I'm also okay, I want to do some exploratory testing + verifying the mutation coverage self test you highlighted before merge.

mbj:
@dgollahon So very good observation that the rspec integration was not included. This is basically the case as mutant, at the time it runs subject matching had not loaded the rspec integration.

Applying this diff to your commit:

```diff
diff --git a/lib/mutant/integration/rspec.rb b/lib/mutant/integration/rspec.rb
index fba19d4a..efc7b77e 100644
--- a/lib/mutant/integration/rspec.rb
+++ b/lib/mutant/integration/rspec.rb
@@ -102,7 +102,7 @@ module Mutant
           expression = metadata.fetch(:mutant_expression)

           expressions =
-            expression.kind_of?(Array) ? expression : [expression]
+            expression.instance_of?(Array) ? expression : [expression]

           expressions.map(&method(:parse_expression))
         else
diff --git a/mutant.yml b/mutant.yml
index cdc17b0f..540e82f4 100644
--- a/mutant.yml
+++ b/mutant.yml
@@ -4,4 +4,5 @@ includes:
 integration: rspec
 requires:
 - mutant
+- mutant/integration/rspec
 - mutant/meta
```

This makes the integration visible to mutant + covers the one mutation that was reported.

mbj (approved):

mbj:
@dgollahon I've released this as v0.10.9. Cheers, thanks for the contribution.

mutant.yml
```diff
@@ -4,4 +4,5 @@ includes:
 integration: rspec
 requires:
 - mutant
+- mutant/integration/rspec
```
dgollahon, line 7: @mbj Do we need to do the same thing for the minitest integration?

dgollahon:
Thanks for the release!

mutant.yml
mbj: @dgollahon I'm not considering this the final solution. I do not want to load rspec and minitest together at this point.

For now I chose this route as mutant is specified via rspec, for minitest I want to use a different solution, likely a separate mutant config altogether.

dgollahon: Cool, makes sense--just wanted to make sure that possible issue wasn't overlooked.

PR #1129, opened 2020-12-01
Add dedicated build in rails environment loading

mbj:


lib/mutant/bootstrap/rails.rb
```diff
@@ -0,0 +1,49 @@
...
+        # TODO: allow custom preload hooks.
+
+        add_rails_matchers
+      end
+
+    private
+
+      def add_rails_matchers
+        # This logic sucks, instead rails should be come a match expression
+        # possibly allow match expressions like: `ActionController.subclasses`
```
mbj, line 30: So expressions of the form `$Constant.$method` would expect to call a method on an object that returns the actual subclasses.

And: `$Constant{,.$method}` would match both the constant and the classes returned by the subclass.

We'd than still need an expression for the "entirety" of rails owned constants which could be: `<Rails>` or similar.

lib/mutant/bootstrap/rails.rb
```diff
@@ -0,0 +1,49 @@
...
+          matcher: config.matcher.with(match_expressions: rails_expressions)
+        )
+      end
+
+      def rails_expressions
+        expressions = [
+          ApplicationController,
+          *ApplicationController.subclasses,
+          ApplicationRecord,
+          *ApplicationRecord.subclasses
```
mbj, line 43: There is likely way more that should be included in here.

PR #1136, opened 2020-12-03
[WIP] replace thread safe cache backend

mbj:
* Mutant was transitively using thread safe as a cache backend through
  memoizable through equalizer through concord/anima.
* Thread safe on MRI has some outdated assumptions on the GIL, that are
  exposed via very high IO/thread/fork usage.
* This is an in-line change to experiment with a cache replacement.


mbj:
This was done in another initiative.

PR #1139, opened 2020-12-06
Construct file paths relative to repository root

pawelpacana:
* git diff-index output is relative to git root
* running mutant deeper from the git root breaks assumption of git root
  being equal to work dir

### DIFF
--- a/lib/mutant/repository/diff.rb
+++ b/lib/mutant/repository/diff.rb
@@ -31,15 +31,21 @@ module Mutant
 
       # rubocop:disable Metrics/MethodLength
       def touched_paths
-        pathname = world.pathname
-        work_dir = pathname.pwd
+        root_stdout =
+          world
+            .capture_stdout(%w[git rev-parse --show-toplevel])
+            .from_right
+            .lines
+            .first
+            .strip
+        root_dir = Pathname.new(root_stdout)
 
         world
           .capture_stdout(%W[git diff-index #{to}])
           .from_right
           .lines
           .map do |line|
-            path = parse_line(work_dir, line)
+            path = parse_line(root_dir, line)
             [path.path, path]
           end
           .to_h
--- a/spec/unit/mutant/repository/diff_spec.rb
+++ b/spec/unit/mutant/repository/diff_spec.rb
@@ -14,6 +14,12 @@ describe Mutant::Repository::Diff do
     let(:path)       { Pathname.new('/foo/bar.rb')      }
     let(:line_range) { 4..5                             }
 
+    let(:root_stdout) do
+      <<~STR
+        /foo
+      STR
+    end
+
     let(:world) do
       instance_double(
         Mutant::World,
@@ -32,6 +38,12 @@ describe Mutant::Repository::Diff do
 
     let(:raw_expectations) do
       [
+        {
+          receiver:  world,
+          selector:  :capture_stdout,
+          arguments: [%w[git rev-parse --show-toplevel]],
+          reaction:  { return: Mutant::Either::Right.new(root_stdout) }
+        },
         {
           receiver:  world,
           selector:  :capture_stdout,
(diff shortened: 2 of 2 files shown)


mbj:
@pawelpacana Lol, I thought you opened an issue, so I made my own version in #1142 + some other cleanups. When I prepared the release I was trying to find that issue, and now found this PR.

I'm closing yours as I've got no CLA agreement with you (at this point, happy to get you the opportunity to sign it), and I think my version is slightly superior.

pawelpacana:
@mbj whatever works to have a released solution, happy to sign CLA for the future

lib/mutant/repository/diff.rb
```diff
@@ -31,15 +31,21 @@ def touches?(path, line_range)
...
       def touched_paths
-        pathname = world.pathname
-        work_dir = pathname.pwd
+        root_stdout =
+          world
+            .capture_stdout(%w[git rev-parse --show-toplevel])
+            .from_right
+            .lines
+            .first
+            .strip
```
mbj, line 40: Note that `String#strip` is dangerous in this context:


```
irb(main):003:0> " /some/path\n".strip
=> "/some/path"
irb(main):004:0> " /some/path\n".chomp
=> " /some/path"
```

It could be argued that the extra whitespace in the beggining of the string should "never" exist. Hence its good to filter it out. But to me "silently" filtering it out is even more an issue than letting it bubble up, where it can produce an explicit error on followup operations.

My version in #1142 uses `String#chomp` for that reason. Which arguably is still "bad" as `String#chomp` silently does accept this `"/some/path"` (without the expected newline).

The real solution is to parse the path, and enforce the presence of the expected newline and fail in all other cases.

For now, given ruby is such a "high overhead" language I'll go with `String#chomp`.

lib/mutant/repository/diff.rb
```diff
@@ -31,15 +31,21 @@ def touches?(path, line_range)
 
       # rubocop:disable Metrics/MethodLength
       def touched_paths
-        pathname = world.pathname
-        work_dir = pathname.pwd
+        root_stdout =
+          world
+            .capture_stdout(%w[git rev-parse --show-toplevel])
+            .from_right
+            .lines
```
mbj, line 38: The `lines.first` combo is also IMO not a good idea. We know t hat `git rev-parse --show-toplevel` is supposed to print a single line. If it prints more than one line we actually "want to blow up" and not silently truncate all but the first lines.

Also `.first` in case no line is printed: would result in `nil` which is hard to debug.

Short of parsing the entire output, I think we should do the minimal transformation (truncating the `"\n"`) and bubble up the result. This will make the "wrong" values noticable at least.

OT: my haskell mutant main process does this correctly, its still not release ready unfortunately.

spec/unit/mutant/repository/diff_spec.rb
```diff
@@ -117,5 +129,45 @@ def apply
         end
       end
     end
+
+    context 'work dir is subdirectory of git root' do
```
mbj, line 133: this example group proves "nothing" as `/foo` could only have entered the system via `--show-toplevel` in the first place. Hence the semantics of this tests are nothing extra to the existing one. I did not make an equivalent in #1142

mbj:
@pawelpacana I've left some review comments, despite the fact this PR is closed that feedback may be of value to you. Also happy to discuss each individual comment of mine should you disagree.

mbj:
@pawelpacana #1142 is released as `0.10.15`. Cheers.

lib/mutant/repository/diff.rb
pawelpacana: 👍

lib/mutant/repository/diff.rb
pawelpacana: good point on `nil`

spec/unit/mutant/repository/diff_spec.rb
pawelpacana: the existing ones remained untouched with regard to paths and did not justify adding the ceremony of relative

mbj: I've got to disagree.

The specs proof that all paths are constructed relative to `--show-toplevel` when we changed out the expectation from `Pathname.pwd` to `git rev-parse --show-toplevel`.

Basically this spec used to describe the situation "relative to `Pathname.pwd`" and once changed (in #1142) to use `git rev-parse --show-toplevel` (and not anymore `Pathname.pwd), the specs now describe the situation relative to the new base. We also "proof" (by example only) that `/foo` could only have entered the system via the one method now used.

We changed the "method the base" enters the code base. The spec you added "did not prove anything extra" on top of what was already proven. As additional evidence: I had not to add your example over in #1142 to cover all mutations.

IMO tests should not "re-specify" that the boundaries (before `Pathname.pwd`, now `git rev-parse --show-toplevel`) do, insted they are supposed to specify what our code does on top. We changed the code to use a new base, and hence I only expect this change to be reflected in the specs.

PR #1145, opened 2020-12-08
Change back to lazy mutation

mbj:
* It turns out that the original reason for the change to AOT
  mutation details is not valid anymore after the MRI bug that causes
  the VM to get stuck was reproduced outside memoizable sharing.

### DIFF
--- a/lib/mutant/mutation.rb
+++ b/lib/mutant/mutation.rb
@@ -9,34 +9,37 @@ module Mutant
     CODE_DELIMITER = "\0"
     CODE_RANGE     = (0..4).freeze
 
-    def initialize(subject, node)
-      super(subject, node)
-
-      @source         = Unparser.unparse(node)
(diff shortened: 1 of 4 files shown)


lib/mutant/mutation.rb
```diff
@@ -9,34 +9,37 @@ class Mutation
     CODE_DELIMITER = "\0"
     CODE_RANGE     = (0..4).freeze
 
-    def initialize(subject, node)
-      super(subject, node)
-
-      @source         = Unparser.unparse(node)
-      @code           = sha1[CODE_RANGE]
-      @identification = "#{self.class::SYMBOL}:#{subject.identification}:#{code}"
-      @monkeypatch    = Unparser.unparse(subject.context.root(node))
```
mbj, line 18: This work used to block the main thread. But its specific to an individual mutation. Moving unique to a mutation work into the killforks is key to improve concurrency.

PR #1146, opened 2020-12-08
[WIP] Documentation revamp

telyn:
Hi Markus,

Please ignore the long lines, double-spaces and fancy quotes - they are not my preference. I've been trying out a new editor and it **seemed** like it was correctly automatically putting new-lines in, but it was all a front! I'll be going back over everything quickly at the end with vim, my one true love.

I'll be continuing to alter this PR over the next week or so - this push is just so you can see the current style and give any early feedback.


docs/limitations.md
```diff
@@ -41,6 +41,21 @@ Mutant cannot emit mutations for some subjects.
     end
     ```
 
+* multiline-style singleton methods not defined on `self`
```
dgollahon, line 44: grouping by `inline-style` and `multiline-style` seems like a confusing distinction to me.

docs/nomenclature.md
```diff
@@ -18,25 +27,81 @@ Mutant currently supports the following subjects:
 * Instance methods
 * Singleton (class) methods
 
-Other subjects (constants, class bodies for DSLs, ...) are possible but aren't
-implemented in the OSS version.
+Other subjects are possible (even project-specific subjects) but aren't implemented in the OSS version. Please get in touch with the authors. Some examples are:
+* Constants
+* Class bodies for DSLs
+
+The more subjects that mutant can alter in your project, the more mutations it can create, and so the higher your confidence you can have that your tests cover the semantics of your application. Please get in touch if you require subjects beyond those implemented in Mutant already - support may be available in the commercial version.
```
dgollahon, line 34: This advises you to get in touch twice--feels slightly redundant.

docs/nomenclature.md
```diff
@@ -18,25 +27,81 @@ Mutant currently supports the following subjects:
...
+
+## Mutation Operator
```
dgollahon, line 36: I don't know if this was discussed in the discord or elsewhere, but this terminology seems slightly confusing. What if this heading was just something like "Kinds of Mutations".

docs/nomenclature.md
```diff
@@ -18,25 +27,81 @@ Mutant currently supports the following subjects:
...
+
+Please be aware that there is no reason to learn these terms (there’s no quiz at the end of this semester!) - we’re just providing them to give you some ideas of what high-level classes of changes mutant makes to your code. For a full list of all the Mutation Operators in mutant, see the code in the [meta directory](https://github.com/mbj/mutant/tree/master/meta).
+
+* **Semantic Reduction**
+
+    This type of transformation replaces a piece of code which has somewhat complex semantics with one that has simpler semantics. To aid understanding, here are a couple of different sub-categories you could put them into. 
+    
+    * **Method call replacement** - for example, `#==` -> `#eql?` -> `#equal?`
+
+        `#==` commonly performs conversion between types in addition to checking equality, while `#eql?`  tends to check only that the class and instance variables are equal. Therefore we would say that #== is semantically simpler.
```
dgollahon, line 48: You might consider adding a small example. I like explaining this by `1 == 1.0` is true but `1.eql?(1.0)` is false.

docs/nomenclature.md
```diff
@@ -18,25 +27,81 @@ Mutant currently supports the following subjects:
...
+        
+        You could also think of a semantic reduction in these cases as an increase in “strictness” of the code. `#equal?` is a stricter equality test than `#eql?`, and so on
+        
+    * **Code removal** - For example, `def my_method; do_something; end` -> `def my_method; end`
```
dgollahon, line 52: Might be clearer rendered as a multiline ```diff?

docs/nomenclature.md
```diff
@@ -18,25 +27,81 @@ Mutant currently supports the following subjects:
...
+     
+        We could argue[^1] that this category includes all of the following:
+        
+        *  Removing a particular expression, from the AST entirely
+        *  Replacing an expression with `nil`, or `true`, or `false`, or other simple literals.
```
dgollahon, line 57: I don't tend to think of those as quite the same as "code removal" but maybe it belongs here

docs/nomenclature.md
```diff
@@ -18,25 +27,81 @@ Mutant currently supports the following subjects:
...
+[^1]: After all; these are not official terms handed down from an authority on mutation testing. We’ve invented them for the purposes of introducing the concepts here.
 
-A transformation applied to the AST of a subject. Mutant knows the following high level operator
-classes:
+* **Orthogonal Replacement** - for example `>` -> `<`
 
-* Semantic Reduction
-* Orthogonal Replacement
-* [Noop](#neutral-noop-tests)
+   Unlike semantic reduction, where the result is a “simpler” version of the input, an orthogonal replacement transforms some code with a given function into code with a similar semantic complexity, but which does something different (usually opposite). This category is probably better understood by examples:
```
dgollahon, line 68: WDYT about "usually opposite" -> "usually the opposite"?

docs/nomenclature.md
```diff
@@ -1,13 +1,22 @@
...
 ============
 
-The following explains several nouns you may experience in mutant's documentation.
+This document explains several nouns you may experience in mutant's documentation.
 It's a good idea to familiarize yourself before moving on.
 
+## Mutation Testing
+The practice of systematically applying small changes one at a time to a codebase then re-running the (relevant) tests for each change. If the tests fail, then they cover the semantics of those changes. If the tests continue to pass with the changes in place, the tests do not cover the complete semantics of the changed parts of the codebase.
+
+Each type of change (for example, “change `a && b` to `a`”) is known as a [Mutation Operator](#mutation-operator), and is applied to the pristine codebase - that is, the changes are not stacked up, but applied in isolation. The places in which the Mutation Operators can make their changes are called [Subjects](#subject). The changed Subject is then referred to as a [Mutation](#mutation).
```
dgollahon, line 10: "Each type of change (for example, “change `a && b` to `a`”) is known as a ..."

This sounds like the change itself is a "Mutation Operator" but I think that would be a [Mutation](https://github.com/mbj/mutant/pull/1146/files#diff-30f1590c9d25c9533cc089fcdfed706a2b237024223e18ca16db6d857b2de656R80-R81). You mean to be talking about kinds of changes but I think the example confuses it a bit because it is an instance of a mutation.

docs/reading-reports.md
```diff
@@ -21,7 +21,7 @@ Mutation output is grouped by selection groups. Each group contains three sectio
 
    **Format**:
 
-   ```text
+,   ```text
```
dgollahon, line 24: Typo?

meta/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+* `str` and `sym` are string and symbol literals that do not contain dynamic elements (`"I'm a dstring #{i_am_the_dynamic_element}"`)
+    * `dst` and `dsym` are the equivalents for strings and symbols which do
+* `regexp` is a regular expression without any options (`/like this/`)
+    * `regopt` is a regular expression with options (`/like this/i`)
+* a few come in regular and `-asgn` variants - the latter are the assignment versions. E.g., `||=` is the assignment version (`or_asgn.rb`) of `||`  (`or.rb`)
+* `gvar`, `cvar`, `ivar` and `lvar` refer to global (`$VERBOSE`), class (`@@debug`), instance (`@amount`) and local (`response`) variables, respectively.
+* `csend` is the conditional send operator (`&.`)
+* `cbase` is a top-level constant (`::Kernel` or `::Errno::ENOENT`), whereas `const` is a "normal" constant (`Kernel` or `Errno::ENOENT`)
+* `casgn` is an assignment to a constant (`VERSION = "0.1.0"`)
```
dgollahon, line 16: This strikes me as very helpful to new contributors :)

dgollahon:
Lots of improvements here 🎉 Glad to see the docs getting additional attention.

I left a few thoughts I had as I read through.

telyn:
> I left a few thoughts I had as I read through.

Thanks for the feedback - I haven't read any of it (yet) as it's midnight but I'll incorporate it into my TODOs up top tomorrow ❤️

docs/nomenclature.md
telyn: Yeah, I had meant to supplant the first one with the latter 😅

docs/nomenclature.md
telyn: Yeah. Probably would reduce the visual density of this section a bit too.. It is a bit hefty.

docs/nomenclature.md
telyn: I agree that it's confusing, and I think it's an unnecessary type of thing to have as its own heading.

I used it because it was used in the original document, but I think it may be better to have it as a sort-of "here are some types of mutations that can occur" sub-section in the "Mutations" section. What do you & @mbj think to that?

docs/nomenclature.md
telyn: Yeah. I am a bit worried that I've gone too far into the weeds with describing this one - in an un-committed draft I had exactly that example, hah.

docs/nomenclature.md
telyn: Mm. That's true, it's more of an extreme semantic reduction, I suppose 😛

I think one thing that's missing from this whole section is a motivation for choosing to categorise things, as well as a motivation for these specific categories. I'm mostly using it as a way to structure the documentation to break it up into smaller more digestible pieces.

In my experience, when I run mutant on a subject and it reduces 2-3 lines of code to `nil` or `self`, I'm thinking "ah I just haven't specified those features in my tests", or "perhaps I could restructure that into its own method and test it individually". When I run it and it removes or reduces a smaller amount of code, I've found it's more likely that it's actually removing / reducing something that isn't necessary. This may just be a symptom of the fact my mutation testing has so far been limited to things I have built since setting up mutant.

I realise I am coming to this from a bit of a beginner's standpoint, which is perhaps not helping me to write this document in a universally applicable way. I'm happy to continue a discussion on this and to defer to experts!

meta/README.md
telyn: , or Things I Had To Learn When I Was Adding Metaclass Support 😁

docs/nomenclature.md
telyn: I've reworked this a bit. As well as combining the mutation and mutation operator sections, I've reworked this section a bit to add in some of that motivation. Let me know what you think of it now!

PR #1153, opened 2020-12-12
Add test failure rendering for minitest

mbj:


lib/mutant/integration/minitest.rb
```diff
@@ -95,6 +95,8 @@ def call(tests)
           break unless test.call(reporter)
         end
 
+        reporter.report
```
mbj, line 98: The minitest integration itself does not have a full mutation covering test right now. It only has integration tests.

This is for the reason mutant (at this point) is not prepared to load multiple integrations the same time, and as mutant itself requires rspec on itself this is a bit of a tie.

This situation should improve in the future.

PR #1155, opened 2020-12-13
Add defined? -> instance_variable_defined? mutation

### LINKED ISSUE #1138: Add mutation from `defined(@ivar)` to `instance_variable_defined?(:ivar)`.

dgollahon:
- Mutates expressions where `defined?` is being used to check specifically for an instance variable such as `defined?(@a)` to the more narrow `instance_variable_defined?(:@a)` instead.
- Closes #1138.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,7 +2,13 @@
 
 * [#1155](https://github.com/mbj/mutant/pull/1155)
 
-  Add `defined?(@a)` -> `instance_variable_defined?(:@a)` mutation. 
+  * Add `defined?(@a)` -> `instance_variable_defined?(:@a)` mutation.
+  * Remove invalid mutations from `defined?` -> `true`.
+  * Remove mutations of `defined?()` arguments, the `defined?` method
+    is not fully evaluating the argument but instead partially evaluates the
+    AST, or inspects the AST.
+    Delegating to the value semantics based "generic" mutation engine does create
+    too many hard to cover mutations.
 
 # v0.10.18 2020-12-13
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -32,8 +32,8 @@
     This is useful when reacting to noop errors.
 
 * [#1154](https://github.com/mbj/mutant/pull/1154)
- 
-  * Add subcommand `environment subject list`. It allows to list 
+
+  * Add subcommand `environment subject list`. It allows to list
     all matched subjects.
 
 # v0.10.17 2020-12-09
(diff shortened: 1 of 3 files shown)


meta/defined.rb
```diff
@@ -7,3 +7,13 @@
...
   mutation 'true'
 end
+
+Mutant::Meta::Example.add :defined? do
+  source 'defined?(@foo)'
+
+  singleton_mutations
+  mutation 'defined?(nil)'
+  mutation 'defined?(foo)'
+  mutation 'true'
```
dgollahon, line 17: I would do this as a separate PR, but should the `defined?` mutator also emit `false`? It's interesting that it is only true. Is there a reason for that?

Changelog.md
```diff
@@ -20,8 +24,8 @@
     This is useful when reacting to noop errors.
 
 * [#1154](https://github.com/mbj/mutant/pulls/1154)
- 
-  * Add subcommand `environment subject list`. It allows to list 
+
+  * Add subcommand `environment subject list`. It allows to list
```
dgollahon, line 28: My editor trimmed some trailing whitespace. I can remove this from the commit if you want but also seems like it is worth fixing.

meta/defined.rb
mbj: I think this was a deliberate decision as `singleton_mutations` never emits `false`:

```
irb(main):001:0> defined?(@a)
=> nil
irb(main):002:0> @a = "foo"
=> "foo"
irb(main):003:0> defined?(@a)
=> "instance-variable"
```

emitting `true` is also wrong in that light ;) I'll add a commit to improve this.

Changelog.md
mbj: Added an extra commit, and my editors whitespace trimming broke ;) Good catch.

meta/defined.rb
dgollahon: Ah, I guess since we are emitting `nil` we already have something falsey.

I actually think `true` was a good mutation--don't we normally replace predicates with `true` and `false`? (or maybe there is an issue for this somewhere.

Or are you thinking it's just too hard to cover normal `defined?` mutations because it's hard to have both states?

dgollahon: oh! nevermind. I see that `defined?` never returns `true` normally. Ok, cool, good call, ignore me. I misunderstood what you were saying for a moment. Nice improvement.

mbj: no worries, `defined?` is weird in other ways.

It actually gives information about the AST, but when the AST references a constant, than it also checks the presence of that constant. Ruby....

PR #1157, opened 2020-12-13
Fix changelog pull request links

dgollahon:
- The correct destination is `.../pull/[number]`, not `.../pulls/[number]`.


mbj:
@dgollahon lol, I introduced the mistake than copy and pasted it to the next entries. Thanks!

PR #1176, opened 2020-12-23
Add matcher config in configuration file

mbj:


docs/configuration.md
```diff
@@ -48,6 +48,42 @@ When `fail_fast` is enabled, mutant will stop as soon as it encounters an alive
...
+matcher:
+  # Subject expressions to find subjects for mutation testing.
+  # Multiple entries are allowed and matches from each expression
+  # are unioned.
+  #
+  # Subject expressions can also be specified on the command line. Example:
+  # `bundle exec mutant run YourSubject`
+  #
+  # Note that expressions from the command line replace the subjects
+  # configured # in the config file!
```
dgollahon, line 65: Looks like an errant `#` here.

mbj: fix & fixed.

.github/workflows/ci.yml
```diff
@@ -43,7 +43,7 @@ jobs:
       with:
         ruby-version: ${{ matrix.ruby }}
         bundler-cache: true
-    - run: ./mutant.sh --since HEAD~1 -- 'Mutant*'
+    - run: bundle exec mutant run --zombie --since HEAD~1
```
dgollahon, line 46: Should `mutant.sh` be deleted?

lib/mutant/config.rb
```diff
@@ -159,13 +112,14 @@ def self.env
         Transform::Exception.new(YAML::SyntaxError, YAML.method(:safe_load)),
         Transform::Hash.new(
           optional: [
-            Transform::Hash::Key.new('coverage_criteria', CoverageCriteria::TRANSFORM),
+            Transform::Hash::Key.new('coverage_criteria', ->(value) { CoverageCriteria::TRANSFORM.call(value) }),
```
mbj, line 115: I "really" like this require sequence:

```
require 'foo'
require 'foo/bar'
```

But we have to access the constant from `foo/bar` here, hence as we are in class scope and the require for `foo/bar' did not happen yet we have to use a lambda indirection.

The alternative would be:

```
require 'foo/bar'
require 'foo'
```
which sucks. For the moment I eat it. Would love of Ruby had a module system ;).

docs/mutant-minitest.md
```diff
@@ -64,7 +64,7 @@ This prints a report like:
 
 ```sh
 Mutant environment:
-Matcher:         #<Mutant::Matcher::Config match_expressions: [AUOM*]>
+Matcher:         #<Mutant::Matcher::Config subjects: [AUOM*]>
```
dgollahon, line 67: This is a good change I think--it reads more clearly.

mbj, line 67: I think this is much nicer to read. Its clear the domain of this method is matching, else the class would not be named `Mutant::Matcher`.

So the `match_` prefix was redundant always.

Just naming the key `expressions` would IMO also suck as it than would have a sibling `ignore_expressions` which does not clearly indicate what the difference between both would be.

Hence replacing this with `subjects` makes it very clear. And `subject_expressions` is kind of redundant as we can only match expressions.

lib/mutant.rb
```diff
@@ -41,6 +41,7 @@ module Mutant
   SCOPE_OPERATOR = '::'
 end # Mutant
 
+require 'mutant/transform'
```
mbj, line 44: The fact this has to move up the require chain to toplevel, indicates it likely belongs to a separate library.

Once I make nicer "input error location rendering" this is worthy of a separate lib.

lib/mutant/config.rb
dgollahon: > Would love of Ruby had a module system ;)

yeahh... I feel that one!

lib/mutant/config.rb
```diff
@@ -159,13 +112,14 @@ def self.env
...
             Transform::Hash::Key.new('fail_fast',         Transform::BOOLEAN),
             Transform::Hash::Key.new('includes',          Transform::STRING_ARRAY),
             Transform::Hash::Key.new('integration',       Transform::STRING),
             Transform::Hash::Key.new('jobs',              Transform::INTEGER),
             Transform::Hash::Key.new('mutation_timeout',  Transform::FLOAT),
-            Transform::Hash::Key.new('requires',          Transform::STRING_ARRAY)
+            Transform::Hash::Key.new('requires',          Transform::STRING_ARRAY),
+            Transform::Hash::Key.new('matcher',           Matcher::Config::LOADER)
```
mbj, line 122: This one, paired with adding the constant `Mutant::Config::LOADER` is the only real change to make config file support for matchers.

.github/workflows/ci.yml
mbj: yes, good catch. Happy to se it go.

I fought myself to add `zombie: true` to the config file loader, and decided against it.

Nobody who is not willing to read mutants source should ever use `--zombie`.

dgollahon: > I fought myself to add zombie: true to the config file loader, and decided against it.

That seems like the right call.

dgollahon (dismissed):
Woo! I'm excited about this one 🎉 LGTM assuming CI passes.

mbj:
> Woo! I'm excited about this one  LGTM assuming CI passes.

it should, I repushed a lot so have to manage the queue. Merging once done.

PR #1178, opened 2020-12-25
Add Ruby 3.0 support

mbj:


mutant.sh
```diff
@@ -1,11 +1,12 @@
...
-  --ignore-subject Mutant::CLI#add_debug_options                 \
-  --ignore-subject Mutant::Isolation::Fork::Parent#call          \
-  --ignore-subject Mutant::Mutator::Node::Argument#skip?         \
-  --ignore-subject Mutant::Mutator::Node::ProcargZero#dispatch   \
-  --ignore-subject Mutant::Mutator::Node::When#mutate_conditions \
-  --ignore-subject Mutant::Zombifier#call                        \
+bundle exec mutant run                                                 \
+  --zombie                                                             \
+  --ignore-subject Mutant::CLI#add_debug_options                       \
+  --ignore-subject Mutant::Expression::Namespace::Recursive#initialize \
```
mbj, line 6: A mutation for 3.0 specific code here. We need version specific configuration to exclude the subject (or mutation) on specific versions.

.github/workflows/ci.yml
```diff
@@ -119,7 +119,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [ruby-2.7]
+        ruby: [ruby-2.7, ruby-3.0]
```
dgollahon, line 122: Are you intentionally running rubocop on two versions of ruby?

lib/mutant/expression/namespace.rb
```diff
@@ -17,7 +17,7 @@ class Recursive < self
         def initialize(*)
           super
 
-          @syntax = "#{scope_name}*"
+          @syntax = "#{scope_name}*".freeze # rubocop:disable Style/RedundantFreeze
```
dgollahon, line 20: Also curious if you know something I don't about this--I thought this was automatically frozen even with interpolation?

```ruby
# frozen_string_literal: true

class Foo
  def initialize
    @test = "#{rand}-test"
    pp @test.frozen?
  end
end

Foo.new # outputs `true`
```

.github/workflows/ci.yml
mbj: for the moment yes, will soon remove 2.7.

lib/mutant/expression/namespace.rb
mbj: https://twitter.com/_m_b_j_/status/1342635659547340800

TLDR: MRI decided that, as each dstr gets allocated always freezing makes no sense.

I disagree as freezing is also a correctness mechanism, not just a memory optimization.

Instead: They should have patched MRI such as `+"some-#{dstr}"` would create a mutable DSTR, without frozen intermediary.

dgollahon: ...oh. I don't like that at all. :(

PR #1181, opened 2020-12-30
Remove bad kwarg mutations

mbj:


Changelog.md
```diff
@@ -1,5 +1,13 @@
 # v0.10.22 2020-12-23
 
+* [#1179](https://github.com/mbj/mutant/pull/1179)
+
+  * Remove unuseful kwarg mutations.
+  * Remove argument promotion on kwarg mutations. These yield AST that
+    has when unpared invalid syntax.
```
dgollahon, line 7: `unpared` typo

dgollahon: That sentence is also a bit confusing

meta/send.rb
```diff
@@ -379,8 +379,6 @@
   mutation 'foo.public_send(:bar, -1, two: true, **kwargs, &block)'
   mutation 'foo.public_send(:bar, 2, two: true, **kwargs, &block)'
   mutation 'foo.public_send(:bar, two: true, **kwargs, &block)'
-  mutation 'foo.public_send(:bar, 1, nil => true, **kwargs, &block)'
-  mutation 'foo.public_send(:bar, 1, self => true, **kwargs, &block)'
```
dgollahon, line 383: nice :)

mbj: @dgollahon The reason this was not found before is that argument promotion only happens on one argument.

So we had to find a single kwarg to see it.

Mutants corpus test is quite heavy and so far it "only" generates the mutation, but does NOT unparse + parse (full validation).

With mutant now being able to natively use multiple processes (The parallel engine is generic) this can be fixed.

I'll not use `ractor` as its experimental, and it'll take a long time till the last Ruby that does not have `ractor` will be EOL.

dgollahon: Ahh, full parse/unparse validation would be great to add.

PR #1183, opened 2020-12-30
Change to avoid negation

mbj:


lib/mutant/mutator/node/kwargs.rb
```diff
@@ -28,15 +28,15 @@ def emit_argument_presence
         def emit_argument_mutations
           children.each_with_index do |child, index|
             Mutator.mutate(child).each do |mutant|
-              if forbid_argument?(mutant)
+              unless forbid_argument?(mutant)
```
mbj, line 31: I actually originally intend to negate in the name and call side, but only carried out the name.

PR #1188, opened 2021-01-02
Add `a === b` -> `a.is_a?(b)` mutation

### LINKED ISSUE #689: Add mutation from `foo === bar` → `bar.kind_of?(foo)`
Threequals might be more powerful than the use case requires

```ruby
String === 'foo'                       # => true
/foo/  === 'foo'                       # => true
->(value) { value == 'foo' } === 'foo' # => true
```

dgollahon:
- Closes #689
	* NOTE: I have gone with `is_a?` over `kind_of?` because [this is more commonly used](https://github.com/rubocop-hq/ruby-style-guide#is-a-vs-kind-of)

### DIFF
--- a/lib/mutant/mutator/node/send.rb
+++ b/lib/mutant/mutator/node/send.rb
@@ -17,6 +17,7 @@ module Mutant
           :< =>          %i[== eql? equal?],
           :<= =>         %i[< == eql? equal?],
           :== =>         %i[eql? equal?],
+          :=== =>        %i[is_a?],
           :=~ =>         %i[match?],
           :> =>          %i[== eql? equal?],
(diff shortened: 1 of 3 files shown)


lib/mutant/mutator/node/send.rb
```diff
@@ -17,6 +17,7 @@ class Send < self
           :< =>          %i[== eql? equal?],
           :<= =>         %i[< == eql? equal?],
           :== =>         %i[eql? equal?],
+          :=== =>        %i[is_a?],
```
dgollahon, line 20: @mbj I wonder if this should also have a mutation to `#eql?`?

mbj (approved):

mbj (approved):

PR #1191, opened 2021-01-02
Remove duplicate meta examples

dgollahon:
I found a couple of meta examples that were duplicated so I am removing them. I have additionally added a test to prevent this from recurring.

### DIFF
--- a/spec/unit/mutant/meta/example_spec.rb
+++ b/spec/unit/mutant/meta/example_spec.rb
@@ -30,6 +30,14 @@ RSpec.describe Mutant::Meta::Example do
     end
   end
 
+  it 'does not define any repeated examples' do
+    source_counts = described_class::ALL.each_with_object(Hash.new(0)) do |example, counts|
+      counts["#{example.location.path}:#{example.original_source}"] += 1
+    end
+
+    expect(source_counts.select { |_source, count| count > 1 }.keys).to eql([])
+  end
+
   describe '#original_source_generated' do
     subject { object.original_source_generated }
 
--- a/meta/block.rb
+++ b/meta/block.rb
@@ -80,16 +80,6 @@ Mutant::Meta::Example.add :block do
 end
 
 Mutant::Meta::Example.add :block do
-  source 'foo { |a| }'
-
-  singleton_mutations
-  mutation 'foo { || }'
-  mutation 'foo { |a| raise }'
-  mutation 'foo { |_a| }'
-  mutation 'foo'
-end
-
-Mutant::Meta::Example.add :block do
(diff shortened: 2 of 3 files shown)


spec/unit/mutant/meta/example_spec.rb
```diff
@@ -30,6 +30,18 @@
     end
   end
 
+  it 'does not define any repeated examples' do
+    source_by_path = described_class::ALL.group_by do |example|
+      "#{example.location.path}:#{example.original_source}"
+    end
+
+    expect(
+      source_by_path.transform_values(&:count)
```
mbj, line 39: You can avoid this reprojection via:

```
counts = described_class::ALL.each_with_object(Hash.new(0)) do |example, counts|
  counts["#{example.location.path}":#{example.original_source}"] += 1
 end
 ```

dgollahon: I can do that if you prefer. I tend to like to transform in steps instead of manually building up counts. Normally I'd use `#tally` but we support rubies that don't have it.

mbj: this one is a hard one to decide, unless the following triggers you over: I'm happy to stick with the authors original choice for the moment.

There is an intermediary representation with lots of redundancy, a hash with a path as key and example arrays that never get accessed but their length.

Instead to produce an array to than only ever use its length: why not produce the length directly?

dgollahon: I think it's mostly a matter of taste for this one. What I have is less performant but I personally find it slightly clearer. You're correct that it does more "work" than is necessary.

I have to rebase anyway so I'll just go with your style as it is your project :)

mbj: As we are golfing some non blocking thoughts:

* Instead of `select { |_, count| count > 1}` why not `reject { |_, count| count.equal?(1)}?`


this is IMO a less complex composition than the alternative. Relational operators are inherently more complex than equality.

but it carries the additional knowledge that `count = 0` does never exist. Which is IMO not too big a problem, compared to getting a rid of the relational operator.

dgollahon: > this is IMO a less complex composition than the alternative. Relational operators are inherently more complex than equality.

While that is technically true, I think that it is clearer as written because it is looking for repeated examples so `> 1` is very obvious in context whereas rejecting exactly one feels like writing the condition backward. I tend to like to think in terms of positive selecting when it's simple to do so.

mbj (approved):

PR #1192, opened 2021-01-02
Add predicate method mutations 

### LINKED ISSUE #263: Predicate call site specific mutations
Rubies convention of postfixing predicate methods with an `?` should allow to add predicate call site specific mutations.
- Replace call site with `true` (already included in other mutation operators)
- Replace call site with `false` (currently not included, we emit `nil` already, an explicit `false` will cover more semantics`).

dgollahon:
- Adds mutations from predicate-like methods (methods ending in ?) to `true`/`false`
  * `a.b?` -> `false`
  * `a.b?` -> `true`
- Closes #263

### DIFF
--- a/lib/mutant/mutator/node/send.rb
+++ b/lib/mutant/mutator/node/send.rb
@@ -84,6 +84,7 @@ module Mutant
         end
 
         def emit_selector_specific_mutations
+          emit_predicate_mutations
           emit_array_mutation
           emit_static_send
           emit_const_get_mutation
@@ -91,6 +92,13 @@ module Mutant
           emit_dig_mutation
           emit_double_negation_mutation
           emit_lambda_mutation
+        end
+
+        def emit_predicate_mutations
+          return unless selector.match?(/\?\z/) && !selector.equal?(:defined?)
+
+          emit(s(:true))
+          emit(s(:false))
         end
 
         def emit_array_mutation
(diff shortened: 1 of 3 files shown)


lib/mutant/mutator/node/send.rb
```diff
@@ -92,6 +93,13 @@ def emit_selector_specific_mutations
           emit_lambda_mutation
         end
 
+        def emit_predicate_mutations
+          return unless selector.match?(/\?\z/) && !selector.equal?(:defined?)
```
mbj, line 97: I think we should go for `selector.to_s.end_with?('?')` instead of the regexp.

Rationale: Asking if a string is a suffix is provided natively by `String`. And regexp are a "too powerful tool" for this operation. We should use the simplest API that gets the job done and not use "untapped power" (of regexp in this case).

dgollahon: I agree but this was to git around adding an `--ignore-subject`. Later rubies support `end_with?('?')` directly on `Symbol`.

dgollahon: I am actually working on adding a mutation to `#start_with?` and `#end_with?` (there is an issue for the former) so this may become an alive again later anyway. How would you like me to resolve this?

mbj: lets stay with regexp. For the moment, till mutant forces us to think about it again when we add the static suffix and prefix regexp to native method call mutations.

mbj (approved):

PR #1195, opened 2021-01-02
Fix flakey specs

dgollahon:
- This seems to be caused by stubbing `Object.const_get`. `IceNine` relies on `.const_get` to build an internal cache that it uses to perform deep freezing. This was introduced in b7c1fc788 but removing it still allows the tests to pass and gives full mutation coverage.

### DIFF
--- a/spec/unit/mutant/subject/method/instance_spec.rb
+++ b/spec/unit/mutant/subject/method/instance_spec.rb
@@ -81,10 +81,6 @@ RSpec.describe Mutant::Subject::Method::Instance::Memoized do
   let(:context)  { Mutant::Context.new(scope, double('Source Path')) }
   let(:node)     { Unparser.parse('def foo; end')                    }
 
-  before do
-    allow(Object).to receive_messages(const_get: scope)
-  end
-
   shared_context 'memoizable scope setup' do
     let(:scope) do
       Class.new do


spec/unit/mutant/subject/method/instance_spec.rb
```diff
@@ -81,10 +81,6 @@ def self.name
   let(:context)  { Mutant::Context.new(scope, double('Source Path')) }
   let(:node)     { Unparser.parse('def foo; end')                    }
 
-  before do
-    allow(Object).to receive_messages(const_get: scope)
-  end
-
```
dgollahon, line 87: @mbj Can you help me understand why this was added originally?

mbj: I suspect this is very old. Before mutant started to use the `Mutant::World` to incrementally phase out this entire concept of "mocking IO globals".

mbj: nope, its me adding it. I must have been at sleep at the wheel.

mbj: I suspect that this was a temporary change to test something that leaded into whats committed unintentionally.

mbj (approved):

mbj (approved):

PR #1198, opened 2021-01-03
Fix expression transform to conform to transform interface

mbj:
* This is a typical case where a type system would never let me go.
* And a case where mutation testing in Ruby could not even force the
  `Left` here as there are no types (yet) for mutant to discover this
  case.

### DIFF
--- a/lib/mutant/matcher/config.rb
+++ b/lib/mutant/matcher/config.rb
@@ -26,7 +26,9 @@ module Mutant
 
       DEFAULT = new(Hash[anima.attribute_names.map { |name| [name, []] }])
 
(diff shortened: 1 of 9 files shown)


lib/mutant/matcher/config.rb
```diff
@@ -26,7 +26,9 @@ class Config
 
       DEFAULT = new(Hash[anima.attribute_names.map { |name| [name, []] }])
 
-      expression = ->(input) { Mutant::Config::DEFAULT.expression_parser.call(input) }
+      expression = Transform::Block.capture(:expression) do |input|
+        Mutant::Config::DEFAULT.expression_parser.call(input)
```
mbj, line 30: I want function composition as a syntax feature so badly. Miss you `Data.Function.`.``.

PR #1200, opened 2021-01-03
Add named capture to non-capturing group mutation

dgollahon:
- Mutate named capture groups to passive groups: `/(?<foo>bar)/` -> `/(?:bar)/`.


lib/mutant/mutator/node/regexp/named_group.rb
```diff
@@ -15,8 +15,22 @@ class NamedGroup < Node
           def dispatch
             return unless group
 
-            emit(s(:regexp_passive_group, group))
             emit_group_mutations
+
+            # Allows unused captures to be kept and named if they are explicitly prefixed with an
+            # underscore, like we allow with unused local variables.
+            unless name_underscored?
+              emit(s(:regexp_passive_group, group))
```
dgollahon, line 23: @mbj What do you think of this behavior? I added the passive group mutation to the guard so that if you deliberately underscore a named group (sort of like a comment) you are able to keep it and not be forced to the non-capturing passive group. I could also move it outside of the guard but this was my intuition for what I would likely prefer if I were a user.

mbj (approved):

PR #1201, opened 2021-01-03
Add `#start_with?` and `#end_with?` mutations

### LINKED ISSUE #169: Mutate regexps in form /\Astatic/ to String#start_with?

dgollahon:
- Adds the following mutations:
  * `a.match(/\Atext/)` -> `b.start_with?('text')`
  * `a.match?(/\Atext/)` -> `b.start_with?('text')`
  * `a =~ /\Atext/` -> `b.start_with?('text')`
- NOTE: Adds a `Mutant::AST::Regexp.expand_regexp_ast` to avoid repeating AST expansion logic in the `send` mutator. At that level, the node appears as a simple `parser` `s(:regexp ...)` node so I chose to fully parse the regexp rather than try to do a string pattern.
- Also adds following mutations:
  * `a.match(/text\z/)` -> `b.end_with?('text')`
  * `a.match?(/text\z/)` -> `b.end_with?('text')`
  * `a =~ /text\z/` -> `b.end_with?('text')`
- Closes #169.

### DIFF
--- a/lib/mutant/ast/regexp.rb
+++ b/lib/mutant/ast/regexp.rb
@@ -32,6 +32,23 @@ module Mutant
       def self.to_expression(node)
         Transformer.lookup(node.type).to_expression(node)
       end
+
+      # Convert's a `parser` `regexp` node into more fine-grained AST nodes.
+      #
+      # @param node [Parser::AST::Node]
+      #
+      # @return [Parser::AST::Node]
+      def self.expand_regexp_ast(node)
+        *body, _opts = node.children
+
+        # NOTE: We only mutate parts of regexp body if the body is composed of
+        # only strings. Regular expressions with interpolation are skipped
+        return unless body.all? { |child| child.type.equal?(:str) }
+
+        body_expression = parse(body.map(&:children).join)
+
+        to_ast(body_expression)
+      end
     end # Regexp
   end # AST
 end # Mutant
--- a/spec/unit/mutant/ast/regexp/expand_regexp_ast_spec.rb
+++ b/spec/unit/mutant/ast/regexp/expand_regexp_ast_spec.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+RSpec.describe Mutant::AST::Regexp, '.expand_regexp_ast' do
+  it 'returns the expanded AST' do
+    parser_ast = Unparser.parse('/foo/')
+
+    expect(described_class.expand_regexp_ast(parser_ast)).to eql(
+      s(:regexp_root_expression,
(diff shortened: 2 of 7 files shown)


lib/mutant/ast/regexp.rb
```diff
@@ -32,6 +32,23 @@ def self.to_ast(expression)
...
         Transformer.lookup(node.type).to_expression(node)
       end
+
+      # Convert's a `parser` `regexp` node into more fine-grained AST nodes.
+      #
+      # @param node [Parser::AST::Node]
+      #
+      # @return [Parser::AST::Node]
+      def self.expand_regexp_ast(node)
+        *body, _opts = node.children
```
dgollahon, line 42: Note that destructuring instead of using `#slice` avoids the mutation [we chose to ignore before](https://github.com/mbj/mutant/pull/1201/files#diff-18b52ac8d1d29dd0bd12125e2a4e496ad0743ab551377dbf443ec10d1f37c991L15).

spec/unit/mutant/ast/regexp/expand_regexp_ast_spec.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+RSpec.describe Mutant::AST::Regexp, '.expand_regexp_ast' do
+  it 'returns the expanded AST' do
+    # /foo/
```
mbj, line 5: Non blocking: Instead of this comment: what about running the regexp parser / AST transform to get the `parser_ast` ? This way this test is more meaningful and would have canary properties for future changes?

mbj (approved):

spec/unit/mutant/ast/regexp/expand_regexp_ast_spec.rb
dgollahon: I think that makes sense--I have updated this to match. When I was originally writing the spec it felt like it might be clearer to have the raw AST so I started with that but now that I have just swapped it out for `Unparser.parse` I think it's a lot simpler to read and, as you say, it has a slight "canary" property.

lib/mutant/ast/regexp.rb
mbj: As a followup we could also promote interpolations.

```
a.match?(/\Afoo#{dynamic}/)
```

to:

```
a.end_with?(dynamic)
```

dgollahon: Yeah, something like that might be interesting but I am not sure I will go that deep right away. I have a couple of other small improvements for similar cases I am thinking of as well that are pending.

PR #1207, opened 2021-01-15
Remove `#eql?` -> `#equal?` mutation

dgollahon:
- This mutation has become very frustrating to deal with in certain situations. Notably, due to frozen string literals everywhere matching on `#equal?` one must ensure tests use unfrozen strings using `+` or other similar hacks. I also think that `#eql?` -> `#equal?` has very limited value in most situations.
- NOTE: I have left in mutations to `equal?` as long as there are alternatives to `equal?` in the possible mutations so it is not forced over `#eql?`.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+* [#1207](https://github.com/mbj/mutant/pull/1207)
+
+  * Remove `#eql?` -> `#equal?` mutation
+
 * [#1210](https://github.com/mbj/mutant/pull/1210)
   * Remove generic mutation to `self`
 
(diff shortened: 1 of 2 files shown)


dgollahon:
Per [this discussion on discord](https://discord.com/channels/767914934016802818/767915301874040844/799009759301861407).

dgollahon:
@mbj I am assuming you would prefer to keep the other mutations that include `#equal?` since they do not have the same issue that forcing `#equal?` over `#eql?` does but I am not sure they are very high value so they might make sense to remove as well. Let me know if you would like them to be removed as well.

mbj (approved):

PR #1210, opened 2021-01-16
Remove generic mutation to `self`

dgollahon:
- This rarely yields an alive mutation and, when it does, it can sometimes be frustrating to avoid or work around. It is also probably a bit confusing to newcomers. Removing this should also make mutant runs shorter since it is a frequently generated mutation.

### DIFF
--- a/lib/mutant/mutator/node/send.rb
+++ b/lib/mutant/mutator/node/send.rb
@@ -232,9 +232,16 @@ module Mutant
         def mutate_receiver
           return unless receiver
           emit_implicit_self
+          emit_explicit_self
           emit_receiver_mutations do |node|
             !n_nil?(node)
           end
+        end
+
+        def emit_explicit_self
+          return if UNARY_METHOD_OPERATORS.include?(selector)
+
+          emit_receiver(N_SELF) unless n_nil?(receiver)
         end
 
         def emit_implicit_self
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+* [#1210](https://github.com/mbj/mutant/pull/1210)
+  * Remove generic mutation to `self`
+
 # v0.10.26 2021-01-16
 
 * [#1201](https://github.com/mbj/mutant/pull/1201)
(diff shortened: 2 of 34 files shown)


lib/mutant/mutator/node/send.rb
```diff
@@ -232,11 +232,18 @@ def emit_argument_propagation
...
           emit_receiver_mutations do |node|
             !n_nil?(node)
           end
         end
 
+        def emit_explicit_self
+          return if UNARY_METHOD_OPERATORS.include?(selector)
+
+          emit_receiver(N_SELF) unless n_nil?(receiver)
+        end
```
dgollahon, line 245: @mbj I added this and [this](https://github.com/mbj/mutant/pull/1210/files#diff-46eac0c6f51ce46d1c40d448b59a3219d725f9a571d38c0566b27c3c5905a3c7R20) to keep the change in behavior a bit more narrow since it may be useful to change `foo.bar.baz` to `self.baz`.

That said, we could also just emit implicit self in more cases and not generate this most of the time if we wanted. There are only a few cases where the explicit self is necessary or preferred `self.foo = 1`, etc.

If we emitted the implicit self directly instead of explicit, that would probably be nice for cases where mutant emits `self.baz` and you have to run it again to see that actually you could have just done `baz`. I have a WIP commit that does something like this that I can show you (I may just push it to this branch).

mbj (approved):

mbj:
@dgollahon feel free to merge past rebase.

dgollahon:
Fun fact: on `auom`, this reduces the total number of mutations from 1000 to 805, so i guess this should make most `mutant` runs (after boot time of course) approximately 20% faster.

mbj:
@dgollahon In astronomy there is the concept of a [Standard Candle](https://universe-review.ca/R02-07-candle.htm). An event of known magnitude that can serve as a reference in an otherwise very relative observation space with too many variables.

As the standard candle event does not "change" lots of astrophysics measurements are first made possible. Unlike corpus tests where both the implementation and the content of the corpus can change, currently `auom` is our standard candle. Its code is frozen in time (small library and "done"), hence its "magnitude (of semantics)" is frozen.

Also the test suite is fully covering mutant both for rspec and minitest. Its a very good "non moving" optimization target.

dgollahon:
Yeah, I think it is a very cool reference point. The Standard Candle comparison is a good one.

PR #1214, opened 2021-01-16
Change unit spec timeout from 2s to 3s

dgollahon:
- We are getting fairly regular CI failures where a timeout is hit but will not be hit upon rebuild. Eventually other mechanisms may be superior, but this is a helpful short-term bandaid.


mbj (approved):
Before merging: Can you change your commit message to `Change unit spec timeout to slightly higher value`. I care mostly about having the first verb one of:

```
Add 
Remove
Fix
Change
```

PR #1215, opened 2021-01-17
Fix invalid module mutations

dgollahon:
- Although `class` is handled correctly, `module` previously used a generic mutator that would create invalid mutation candidates such as `module Foo` -> `module nil`. By special-casing the node, we can prevent the class constant from getting mutated to `nil`.


dgollahon:
@mbj I found this by trying out `Unparser::Validation` locally as part of the integration tests. It trickier than I was hoping to just add since there are a lot of false positives with `send` v.s. `lvar` and possibly other problems but I did find this issue early on and decided to do a quick fix. I'm now looking at other validation errors and seeing if I can find more interesting cases.

dgollahon:
Actually, does this matter? Is `module` or `class` ever a direct target of mutation in a real program since everything is scoped to methods? Or am I just confused by trying to resolve something that can only happen in our integration tests? Is there a reason that `module` has not had a mutator node but `class` has?

mbj:
> Actually, does this matter?

Do not assume just the constraints of the mutant version as published in this repo.

mbj (approved):

PR #1218, opened 2021-02-06
Remove integer literal negation mutation

dgollahon:
- Removes mutations of the form of `1` -> `-1`. These are unlikely to yield alive mutations and sometimes lead to frustrating, equivalent mutations.


dgollahon:
See [this discussion](https://discord.com/channels/767914934016802818/767915301874040844/807378465879883786) on Discord.

dgollahon:
@mbj Should we also consider removing negation for floats?

mbj:
> @mbj Should we also consider removing negation for floats?

yeah, lets followup.

mbj (approved):

PR #1249, opened 2021-08-25
Add util subcommand

mbj:


dgollahon:
this is an awesome one! I've been looking forward to having an easy way to do this for years! 😄

mbj:
more to come. I just wish I could abandon writing Ruby ;) and do this all in HS.

PR #1255, opened 2021-08-29
Remove invalid mutations to invalid syntax

mbj:
* Rubies syntax is not easily composable in all cases.
* Unparser should NOT make up for cases where mutant emits invalid AST
  (that on the first glance should be valid, as its just promoting
  expressions).

### DIFF
--- a/lib/mutant/mutator/node/binary.rb
+++ b/lib/mutant/mutator/node/binary.rb
@@ -21,7 +21,11 @@ module Mutant
           emit_singletons
           emit_promotions
           emit_operator_mutations
-          emit_left_mutations
+
+          emit_left_mutations do |mutation|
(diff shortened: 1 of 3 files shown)


lib/mutant/mutator/node/binary.rb
```diff
@@ -21,7 +21,11 @@ def dispatch
           emit_singletons
           emit_promotions
           emit_operator_mutations
-          emit_left_mutations
+
+          emit_left_mutations do |mutation|
+            !(n_irange?(mutation) || n_erange?(mutation)) || !mutation.children.fetch(1).nil?
```
mbj, line 26: Endless ranges to the left of an `||` are invalid syntax. Welcome to ruby syntax design.

PR #1270, opened 2021-10-18
Add sorbet method matching

mbj:
* Signatures with sorbet-runtime enabled wrap the method into a Proc.
* This proc hides the original source location, which would, by default
  not enable mutant to find the method source.
* Using sorbet private API to uncover the original method prevents this
  issue.


vaporyhumo:
This is awesome, thanks @mbj !

PR #1278, opened 2021-11-07
[WIP] Fix unicode property support

dgollahon:
- This is up for early review because I'm not sure about the dynamic creation of the table of unicode properties. I tried just creating a list of them but it was so slow for my editor to process that I couldn't even format the giant lookup table. I suspect that if we want to "bake" these to avoid however long it takes to compute the table and maybe avoid any unexpected drift, it might make sense to dump to YAML or something like that. I'm not sure the best approach.
- I'm also guessing there's a better option than just dumping all the regexp node types in the other list of supported regexp nodes.
- We probably should do this for other regex types--we might be missing some of the posix classes, for instance (I have not checked yet).
- Prevents crashes when having an unsupported property type in source.
- Related to #1234 (which was a very partial fix)
- Note that this turns our `\p{Latin}` formatting into `\p{latin}`. We could fix this with some very simple inflection but I wanted to do the simplest approach first to demonstrate the problem since this seems to be semantically equivalent. The ruby docs use the uppercase form. I have a text file from the upstream regex toolkit that we could use to confirm inflection rules if we want to.


lib/mutant/ast/regexp/transformer.rb
```diff
@@ -12,7 +12,7 @@ class Transformer
         include AbstractType
 
         REGISTRY = Registry.new(
-          ->(type) { fail "No regexp transformer registered for: #{type}" }
+          ->(type) { } # fail "No regexp transformer registered for: #{type}" }
```
dgollahon, line 15: actually my "failing test" [here](https://github.com/mbj/mutant/pull/1278/files#diff-16721c19d3917e959236dc238b8b4b202dee25963638b70deb7bb76070ac8b88R18-R21) probably won't fail with this disabled. i was trying to demonstrate that in another commit but I couldn't get CI to build separately. just try adding any not-explicitly-listed `\p{...}` group and it will fail mutant.

dgollahon:
Mutant also fails on these regexp expressions (taken from `regexp_parser`'s specs): which are `regexp_number_call_backref` nodes.

```ruby
'\g<1>'
"\\g'1'"
'\g<0>'
"\\g'0'"
```

I can file a separate issue for it but we have several problems with the regex handling in mutant. We have disabled these subjects in several places but it's a headache. I am also trying to diff my environment with some configuration changes using `mutant environment subject list` but it crashes so I can't do that. I have just realized I can pass `--ignore-subject` to that command but because I have a custom `mutant:disable` comment interpreter it's a little bit of a pain to do that. I also want to diff the environment without having to do `--ignore-subject` so I can make sure everything gets picked up regardless of the disables, but I can't easily do that right now.

dgollahon:
@mbj Could you take a look and suggest how we should handle the regex types registry? I can play around with it a little more but I don't know if I'll have time to bring this all the way to completion. It causes some headaches for the primary codebase we run `mutant` on and I wanted to demonstrate the problem we have.

mbj:
@dgollahon Overall I like the initiative. But I think we need some sync time on discord to discuss details. Hit me up.

mbj:
Closing in favor of #1319

PR #1285, opened 2021-11-15
Change to ignore multipath matches

mbj:


lib/mutant/matcher/chain.rb
```diff
@@ -14,7 +14,7 @@ class Chain < self
       def call(env)
         matchers.flat_map do |matcher|
           matcher.call(env)
-        end
+        end.uniq
```
mbj, line 17: On the upcoming matcher refactoring this can be made more efficient via using ordered set based intermediaries that are more efficient than `Array#uniq`. But there is no reason to wait for it to land before giving user the benefit of the reduced chance of duplicate subjects.

PR #1288, opened 2021-12-06
Prevent mutations under `T`

vaporyhumo:
This changes will:
* Prevent any mutations to methods sent to the constant `T`

### DIFF
--- a/lib/mutant/ast/meta/send.rb
+++ b/lib/mutant/ast/meta/send.rb
@@ -8,12 +8,14 @@ module Mutant
       # Metadata for send nodes
       class Send
         include NamedChildren, Concord.new(:node), NodePredicates
+        extend Sexp
 
         children :receiver, :selector
 
         public :receiver, :selector
 
         ATTRIBUTE_ASSIGNMENT_SELECTOR_SUFFIX = '='
+        SORBET_RECEIVERS = [s(:const, nil, :T), s(:const, s(:cbase), :T)].freeze
 
         # Arguments of mutated node
         #
@@ -27,6 +29,13 @@ module Mutant
         # @return [Boolean]
         def proc?
           naked_proc? || proc_new?
+        end
+
+        # Test if message receiver is :T
+        #
+        # @return [Boolean]
+        def sorbet_receiver?
+          SORBET_RECEIVERS.include?(receiver)
         end
 
         # Test if AST node is a valid attribute assignment
--- a/lib/mutant/mutator/node/send.rb
+++ b/lib/mutant/mutator/node/send.rb
@@ -57,6 +57,8 @@ module Mutant
       private
 
         def dispatch
+          return if meta.sorbet_receiver?
+
           emit_singletons
 
           if meta.binary_method_operator?
--- a/meta/sorbet.rb
+++ b/meta/sorbet.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+Mutant::Meta::Example.add :send do
+  source 'T.must(a > b)'
+end
+
+Mutant::Meta::Example.add :send do
+  source '::T.must(a > b)'
+end


lib/mutant/ast/meta/send.rb
```diff
@@ -29,6 +29,13 @@ def proc?
           naked_proc? || proc_new?
         end
 
+        # Test if message receiver is :T
+        #
+        # @return [Boolean]
+        def sorbet_receiver?
+          receiver.to_s == '(const nil :T)'
```
vaporyhumo, line 36: Is there a better way for doing this? Also, should I add `::T` too?

lib/mutant/mutator/node/send.rb
```diff
@@ -76,7 +76,10 @@ def meta
         alias_method :arguments, :remaining_children
         private :arguments
 
+
```
vaporyhumo, line 79: DELETE THIS

docs/sorbet.md
```diff
@@ -6,6 +6,19 @@ Mutant aims to implement full sorbet support. Full support includes:
 * Being able to cover mutations via the type system. This capability is being
   developed right now.
 
+For now, all of the mutations to `T` methods are skipped. To better leverage `mutant`'s capabilities
```
vaporyhumo, line 9: idk if this is the best place, placement, or phrasing, but it's an honest attempt.

lib/mutant/mutator/node/send.rb
```diff
@@ -77,6 +77,8 @@ def meta
         private :arguments
 
         def normal_dispatch
+          return if meta.sorbet_receiver?
```
vaporyhumo, line 80: With this we only leave the singleton mutation as far as I could tell.

lib/mutant/ast/meta/send.rb
dgollahon: You might be able to do `receiver.eql?(s(:const, nil, :T))`. If you can't, you may need to mix in `Mutant::AST::Sexp` (or, i suppose, you could also manually initialize `Parser::AST::Node` but it's probably better to mix in).

I think you need to check the `::T` case as well if you want that to be recognized and not mutated, but mutant will also strip off unnecessary `::T` prefixes so that would only matter in fairly rare cases where you shadow `T`.

docs/sorbet.md
```diff
@@ -6,6 +6,19 @@ Mutant aims to implement full sorbet support. Full support includes:
...
+use local variables or parameterless methods when possible, for example:
```
mbj, line 10: I think we should mention that we are making explicit visitors to evaluate arguments to well known sorbet APIs.

lib/mutant/ast/meta/send.rb
mbj: I'd assign it to a constant:

```
SORBET_RECEIVER = s(:const, nil, :t)
```

no need to re-construct the ast node each time.

lib/mutant/mutator/node/send.rb
mbj: I think the singletons should also be avoided in this case.

vaporyhumo: Oh right, they would die always (for `T.must` at least) anyway.

docs/sorbet.md
vaporyhumo: I'm sorry, didn't get this.

mbj: Lets remove this recommendation, I think its diluting the value of mutant, as it adds redundancy. We'll add an explicit visitor to the arguments of `T.must` soon. But asking people to refactor code to something I'd flag (Redundant lvar) is a bad idea.

lib/mutant/ast/meta/send.rb
```diff
@@ -29,6 +30,13 @@ def proc?
           naked_proc? || proc_new?
         end
 
+        # Test if message receiver is :T
+        #
+        # @return [Boolean]
+        def sorbet_receiver?
+          SORBET_RECEIVERS.include? receiver
```
mbj, line 37: Explicit parens please in method bodies, in-line with this entire lib.

docs/sorbet.md
vaporyhumo: Ok I get it now, I don't know what a visitor is but I think i get the idea. Removing this!

lib/mutant/ast/meta/send.rb
vaporyhumo: Sure!

meta/send.rb
```diff
@@ -1,5 +1,13 @@
 # frozen_string_literal: true
 
+Mutant::Meta::Example.add :send do
```
mbj, line 3: @vaporyhumo Lets move that one to `meta/sorbet.rb`.

lib/mutant/ast/meta/send.rb
```diff
@@ -7,13 +7,14 @@ module Meta
...
       class Send
-        include NamedChildren, Concord.new(:node), NodePredicates
+        include NamedChildren, Concord.new(:node), NodePredicates, Sexp
 
         children :receiver, :selector
 
         public :receiver, :selector
 
         ATTRIBUTE_ASSIGNMENT_SELECTOR_SUFFIX = '='
+        SORBET_RECEIVERS = [s(:const, nil, :T), s(:const, s(:cbase), :T)]
```
mbj, line 17: Lets freeze this constant.

vaporyhumo:
Changes are up!

mbj (approved):

mbj:
@vaporyhumo Mutant just gained this ability "generically", see #1341 so mutant will register an indirect global mutation filter for these, without having to manually catch all the cases in the mutation generation engine. Closing this for the moment as its simlper to implement on top of the generic system.

lib/mutant/mutator/node/send.rb
```diff
@@ -57,6 +57,8 @@ class Send < self
       private
 
         def dispatch
+          return if meta.sorbet_receiver?
```
mbj, line 60: @vaporyhumo This change was partial as it would not prevent mutatn to remove the method call entirely from a "mutation higher up".

the newer system in #1341 fixes this more generic.

PR #1290, opened 2021-12-06
Disable case else branch for `T.absurd`

vaporyhumo:
Merging this PR will make `mutant` ignore else branches for case statments when their content is a call to `T.absurd`.

### DIFF
--- a/lib/mutant/mutator/node/case.rb
+++ b/lib/mutant/mutator/node/case.rb
@@ -33,6 +33,8 @@ module Mutant
           else_branch = children.last
           else_index = children.length - 1
           return unless else_branch
+          return if else_branch.eql?(s(:send, s(:const, nil, :T), :absurd))
+
           mutate_child(else_index)
           emit_child_update(else_index, nil)
         end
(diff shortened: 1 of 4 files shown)


meta/case.rb
```diff
@@ -155,3 +155,29 @@
     end
   RUBY
 end
+
+Mutant::Meta::Example.add :case do
```
vaporyhumo, line 159: Would you like to tag sorbet related specs somehow?

lib/mutant/mutator/node/case.rb
```diff
@@ -33,6 +33,8 @@ def emit_else_mutations
           else_branch = children.last
           else_index = children.length - 1
           return unless else_branch
+          return if else_branch.to_s.start_with?("(send\n  (const nil :T) :absurd")
```
vaporyhumo, line 36: I'm looking for the proper way to do this.

mbj: match the exact node: `else_branch.eql?(s(:send, (:const, nil, :T), :absurd))`. Protip next time use:

```
bundle exec ruby-parse --emit-ruby -e 'T.absurd'
```

to discover the ruby syntax for a given AST. `s` is in scope here.

meta/case.rb
mbj: Yes, lets move this to `meta/sorbet.rb` for the moment.

vaporyhumo: Sure!

vaporyhumo:
Closing due to inactivity. I believe this can be done natively now.

PR #1292, opened 2021-12-12
Log compaction

omaik:


lib/mutant/reporter/cli_compact/printer/mutation_result.rb
```diff
@@ -0,0 +1,83 @@
...
+
+          delegate :mutation, :isolation_result
+
+          MAP = {
+            Mutant::Mutation::Evil    => :evil_details,
+            Mutant::Mutation::Neutral => :neutral_details,
+            Mutant::Mutation::Noop    => :noop_details
+          }.freeze
+
+          NEUTRAL_MESSAGE = <<~'MESSAGE'
```
mbj, line 18: I'm not a big fan of this messages being duplicated.

I think we have to share this with the other reporters.

lib/mutant/reporter/cli_compact/printer/env_progress.rb
```diff
@@ -0,0 +1,60 @@
...
+          def mutations_per_second
+            amount_mutation_results / runtime
+          end
+
+          def coverage_percent
+            coverage * 100
+          end
+
+          def overhead_percent
+            (overhead / killtime) * 100
```
mbj, line 54: I suspect this has to be lifted up into the result data structure to avoid duplication.

lib/mutant/reporter/cli_compact/printer.rb
```diff
@@ -0,0 +1,92 @@
+# frozen_string_literal: true
+
+module Mutant
+  class Reporter
+    class CliCompact
```
mbj, line 5: I suspect we have to make a base class for CLI reporters to reduce duplication between compact and "verbose". This way lots of  the private methods below get deduplicated.

lib/mutant.rb
```diff
@@ -281,7 +292,7 @@ class Config
       jobs:                  nil,
       matcher:               Matcher::Config::DEFAULT,
       mutation_timeout:      nil,
-      reporter:              Reporter::CLI.build(WORLD.stdout),
+      reporter:              Reporter::CliCompact.build(WORLD.stdout),
```
mbj, line 295: This is a global change, which works for your instance. I'll have to introduce the concept of selecting a set-of reporters at the CLI/config file.

PR #1304, opened 2022-02-13
Add ruby 3.1 support

mbj:


spec/support/corpus.rb
```diff
@@ -139,8 +139,7 @@ def checkout
       #
       # @return [Integer] mutations generated
       def check_generation(path)
-        node = Parser::CurrentRuby.parse(path.read)
-        fail "Cannot parse: #{path}" unless node
+        node = Parser::CurrentRuby.parse(path.read) or return 0
```
mbj, line 142: That exception was kinda stupid, as `parser` returns `nil` on empty `.rb` files, which is "unfortunate" but valid `.rb` files.

spec/unit/mutant/matcher/methods/instance_spec.rb
```diff
@@ -96,9 +96,11 @@ def apply
     it 'warns about degnerate interfacew' do
       apply
 
+      method_name = :foo
+
       exception =
         begin
-          scope.instance_method(:foo)
+          scope.instance_method(method_name)
```
mbj, line 103: This allows to replicate exactly the same exception we can match against, since ruby 3.1 does source location reflection in exception messages right now. `instance_method(:foo)` would fail this test.

PR #1320, opened 2022-04-25
Add inline disable configuration

mbj:


spec/unit/mutant/parser_spec.rb
```diff
@@ -6,16 +6,43 @@
...
+    let(:buffer) do
+      Unparser.buffer(<<~'RUBY')
+        :source # comment_a
+        :source # comment_b
+      RUBY
+    end
+
+    let(:expected_associations) do
+      associations = {}
+      associations.compare_by_identity
```
mbj, line 22: This is forced by parsers comment associator, I'm personally not super in favor to use object memory identity in any kind of API. 

I'd prefer parser would use node index to associate, but I cannot change its API.

spec/unit/mutant/parser_spec.rb
```diff
@@ -6,16 +6,43 @@
...
     before do
-      expect(path).to receive(:read)
-        .with(no_args)
-        .and_return(':source')
+      allow(path).to receive(:read).and_return(buffer.source)
     end
 
     it 'returns parsed source' do
-      expect(subject).to eql(s(:sym, :source))
+      expect(subject.inspect).to eql(
```
mbj, line 40: Comparing via inspect as `Parser::Source::*` only does identity on `#==` and `#eql?`. And writing peeking assertions is not as exhaustive. Basically an in-memory golden test :P

PR #1321, opened 2022-04-25
[WIP] Failing tests for backref calls

dgollahon:
These all independently crash the unit tests.

I left 4 tests commented out since they seem to be invalid regexps, but there still may be a need to map them. I'm not sure how to construct a valid example off the top of my head.

It might make sense to try to auto-generate regexp specs based off of the calls to [shared examples](https://github.com/ammar/regexp_parser/blob/master/spec/support/shared_examples.rb) in the regexp parser spec files.

### DIFF
--- a/meta/regexp.rb
+++ b/meta/regexp.rb
@@ -1,5 +1,143 @@
 # frozen_string_literal: true
 
+# Generated based on: https://github.com/ammar/regexp_parser/blob/311f70f8b4c3fbd99129846294bbd330ba538622/spec/lexer/refcalls_spec.rb
+Mutant::Meta::Example.add :regexp do
+  source '/(?<X>abc)\k<X>/'
+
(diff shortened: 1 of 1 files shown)


jaynetics:
a valid example for full pattern recursion (`\g<0>`) could be:

`/a(b|\g<0>)/` (which is equivalent to `/a+b/`)

the references `-n`, `+n` are a relative index of a preceding/succeeding group, so a valid example could be:

`/\g<+1>(abc)/` (which matches `abcabc`).

mbj:
@dgollahon I think that these are not an issue anymore with #1361, right?

mbj:
@dgollahon I just tested all of this examples they produce good mutations since 0.11.18. Since we do not have mutations specific to this AST structures I'm closing this without porting the tests into the test suite.

PR #1331, opened 2022-05-02
Add graceful exit on SIGINT

mbj:
* Mutant will now gracefully exit on SIGINT, but still with urgency.
* All active worker threads and their child processes will get killed.
* Only currently finished mutations will be reported.
* Exit status will depend on the current results, if they are all green:
  you get an exist 0, otherwise nonzero.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,8 +1,12 @@
-# Unreleased
+# v0.11.10 2022-05-02
 
 * [#1328](https://github.com/mbj/mutant/pull/1328)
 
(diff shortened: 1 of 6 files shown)


Changelog.md
```diff
@@ -1,3 +1,9 @@
+# v0.11.10 2022-05-02
+
+* [#1327](https://github.com/mbj/mutant/pull/1327)
+
+  Add explicit mutation handler for xstr nodes.
```
dgollahon, line 5: don't forget to update this

mbj: yeah I pushed early to get the PR #.

PR #1335, opened 2022-05-30
Add AST pattern matching

mbj:


lib/mutant/ast/pattern/lexer.rb
```diff
@@ -0,0 +1,173 @@
+# frozen_string_literal: true
+
+module Mutant
+  module AST
+    class Pattern
+      # rubocop:disable Metrics/ClassLength
+      class Lexer
+        WHITESPACE     = [' ', "\t", "\n"].to_set.freeze
+        STRING_PATTERN = /\A[a-zA-Z][_a-zA-Z0-9]*\z/.freeze
```
mbj, line 9: I'm well aware this does not eat all ruby method names, yet. I've to extend this to the wider set as supposed by targets. And eventually we've to support a full blown ruby string parser here to match `value` attributes of `str` tokens.

For "regular" use a quite simlpy regexp will do, unless people insist to use mutant on some crazy UTF-8 code base, while using pattern based mutation filtering on that methods.

lib/mutant/ast/structure.rb
```diff
@@ -0,0 +1,140 @@
...
+        end
+
+      private
+
+        def variable_descendants(node)
+          node.children[fixed.length..]
+        end
+      end
+
+      ALL = [
```
mbj, line 92: Needs more extensions ;) But can replace lots of meta data in `Mutant::Muator::Node` and even unparser if we wanted to.

PR #1373, opened 2023-04-30
Change to improved rspec example selection

mbj:
* Before we'd traverse the entire filtered examples explicitly removing the selection where required.
* This is inefficient especially on larger example numbers.
* The new strategy wipes all examples during setup once, and only re-enables selected examples efficiently and targeted.

### DIFF
--- a/lib/mutant/integration/rspec.rb
+++ b/lib/mutant/integration/rspec.rb
@@ -43,6 +43,8 @@ module Mutant
       # @return [self]
       def setup
         @runner.setup($stderr, $stdout)
+        example_group_map
+        reset_examples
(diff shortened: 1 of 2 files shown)


lib/mutant/integration/rspec.rb
```diff
@@ -119,12 +139,6 @@ def all_examples
         end
       end
 
-      def filter_examples(&predicate)
-        @rspec_world.filtered_examples.each_value do |examples|
```
mbj, line 123: Per mutation this was for 99% of the examples resulting in the same result: Clearing the array entirely.

PR #1374, opened 2023-05-04
Failing test for invalid mutation

dgollahon:


dgollahon:
Note that `bundle exec unparser {file with this code}` doesn't seem to crash

meta/kwarg.rb
```diff
@@ -8,3 +8,11 @@
   mutation 'def foo(bar:); super; end'
   mutation 'def foo(_bar:); end'
 end
+
+Mutant::Meta::Example.add :if do
+  source <<~RUBY
+    def foo(bar:)
+      return true ? [] : [bar]
```
mbj, line 15: BTW: I could reproduce this just on the ternary.

PR #1380, opened 2023-05-08
Add integration options

mbj:


Changelog.md
```diff
@@ -1,8 +1,15 @@
 # v0.11.20 unreleased
 
-* [#1378](https://github.com/mbj/mutant/pull/1379)
+* [#1330](https://github.com/mbj/mutant/pull/1330)
 
-  Add support for user defined mutation worker process hooks. 
+  Change to `--integration-name` from `--use` on CLI and `integration_name` from `integration`
```
dgollahon, line 5: I think `--integration` might be more succinct while equally clear. Either is fine though.

mbj: @dgollahon not if the next commit adds `--integration-argument` (you can repeat as many times as you wish to stack up arguments to pass to rspec/minitest`.

I do not like having one option be the prefix of the next. It makes the prefix version "too special"

dgollahon: Ok. Seems a bit verbose but your call 👍 

`type` might also be clearer than `name` but either works

mbj: well you should only use the verbose ones during experimentation and than persist in your config yaml ;)

`type` is nice, I'll consider it.

mbj: @dgollahon I convinced myself to go with `--integration`.

PR #1381, opened 2023-05-13
Remove unused let & fix malformed test

dgollahon:
- Removes an unused let that does not seem to serve an obvious purpose.
- Fixes a test where an unused let lead to a specified case not being tested.
- Mainly using `mutant` as a canary for an [`rspectre`](https://github.com/dgollahon/rspectre) release.

### DIFF
--- a/spec/unit/mutant/ast/pattern/source_spec.rb
+++ b/spec/unit/mutant/ast/pattern/source_spec.rb
@@ -29,12 +29,11 @@ RSpec.describe Mutant::AST::Pattern::Source do
     end
 
     context 'on non unix newline' do
-      let(:input) { "a\r\nb" }
-
+      let(:string) { "a\r\nb" }
       let(:line_index) { 0 }
 
       it 'considers the \r as part of the previous line' do
-        expect(apply).to eql('a')
+        expect(apply).to eql("a\r")
       end
     end
   end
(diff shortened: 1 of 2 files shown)


spec/unit/mutant/ast/pattern/source_spec.rb
```diff
@@ -29,12 +29,11 @@ def apply
...
 
     context 'on non unix newline' do
-      let(:input) { "a\r\nb" }
-
+      let(:string) { "a\r\nb" }
       let(:line_index) { 0 }
 
       it 'considers the \r as part of the previous line' do
-        expect(apply).to eql('a')
+        expect(apply).to eql("a\r")
```
dgollahon, line 36: @mbj I assume this is the behavior you meant by the docstring but maybe this deserves a closer look?

mbj (approved):

mbj:
@dgollahon Wanna add `rspectree` to the mutant/unparser CI? A Pr would be welcome, this way making both a corpus test subject would create less issues for you.

dgollahon:
@mbj Yes, I would be happy to! The only concern I had about adding it to mutant is that (right now) it needs to run across the whole rspec test suite and I didn't know if that would potentially slow down the build because right now it's split into different groups of test. I'll push a PR and you can evaluate it. I think Unparser should have less overhead so I will open a PR there as well.

PR #1382, opened 2023-05-14
Significantly optimize parsing overhead when `--since` is specified 

dgollahon:
- Massively speeds up boot time on larger codebases and should be noticeable on all sizes. On `mutant` itself, when making a relatively targeted change (such as this one), boot time is reduced by about 0.7 seconds (~20% faster). On a larger project I work on, the speed up was **24 seconds** or **2.5x** faster overall! I left the original, full, subject filtering in place to simplify this change. This just short-circuits when the files we are interested in are not modified and `--since` is specified.
  * `mutant` run before on this change:
    ```
    Benchmark 1: bundle exec mutant run --zombie --since HEAD~1 --profile
    Time (mean ± σ):      4.192 s ±  0.350 s    [User: 8.567 s, System: 6.081 s]
    Range (min … max):    3.850 s …  4.960 s    10 runs
    ```
  * `mutant` run after on this change:
    ```
    Benchmark 1: bundle exec mutant run --zombie --since HEAD~1 --profile
    Time (mean ± σ):      3.535 s ±  0.213 s    [User: 6.890 s, System: 5.784 s]
    Range (min … max):    3.188 s …  3.784 s    10 runs
    ```
- Also reduces memory consumption significantly since the ASTs for all source files are never constructed. In my larger project on very small single subject change it reduced memory allocations by 69MB (80% reduction) and reduced "retained" memory by 25MB (99.4% reduction).
- Also removes some redundant skipped candidate specs


lib/mutant/matcher/config.rb
```diff
@@ -84,6 +84,10 @@ def merge(other)
         )
       end
 
+      def relevant_path?(source_path)
+        subject_filters.all? { |filter| filter.modified_path?(source_path) }
```
mbj, line 88: The contract for subject filters so far is `#call` and I've got mutant clients that have custom subject filters that this code will fail if we where to merge it.

IHMO we should move the repository diff to a core concept in the matcher config allowing to directly reference it without having "all" subject filters having to support this specific concept.

lib/mutant/repository/diff.rb
```diff
@@ -22,12 +22,16 @@ class Error < RuntimeError; end
...
       #   when git command failed
       def touches?(path, line_range)
-        touched_paths
-          .from_right { |message| fail Error, message }
-          .fetch(path) { return false }
+        touched_path(path) { return false }
           .touches?(line_range)
       end
 
+      def touches_path?(path)
```
mbj, line 29: @dgollahon This is a very valid interface addition to `Repository::Diff` I'll extract this to a separate PR to base my universal solution for reducing to be parsed methods down to a smaller amount.

lib/mutant/matcher/config.rb
dgollahon: Ok, that makes sense. Would you like me to change something specifically here? Or how should we proceed?

lib/mutant/cli/command/environment.rb
```diff
@@ -103,10 +103,9 @@ def add_matcher_options(parser)
             add_matcher(:start_expressions, @config.expression_parser.call(pattern).from_right)
           end
           parser.on('--since REVISION', 'Only select subjects touched since REVISION') do |revision|
-            add_matcher(
-              :subject_filters,
-              Repository::SubjectFilter.new(diff: Repository::Diff.new(to: revision, world: world))
-            )
+            diff = Repository::Diff.new(to: revision, world: world)
+            add_matcher(:subject_filters, Repository::SubjectFilter.new(diff: diff))
+            add_matcher(:diffs, diff)
```
dgollahon, line 108: Seems slightly odd to pass the same diff to do different places but this was the simplest/smallest change I could think of.

lib/mutant/matcher/method.rb
```diff
@@ -120,13 +120,24 @@ def subject_config(node)
...
             .on_line(source_line)
             .select { |view| view.node.type.eql?(self.class::MATCH_NODE_TYPE) && match?(view.node) }
             .last
         end
         memoize :matched_view
 
+        def relevant_source_file?
+          diffs = env.config.matcher.diffs
+
+          diffs.empty? ? true : diffs.any? { |diff| diff.touches_path?(source_path) }
```
dgollahon, line 138: It occurs to me that this should maybe be `all?` based on how the subject filters work. It seems like the subject filters have an `all?` block so that you only get the intersection of changes when you supply multiple `--since` flags. I find that slightly surprising because I would expect to get the union of every `--since` flag if I supplied multiple but I also can't think of a case where I would supply multiple. The `any?` is _correct_ in the sense that the subject filter would remove everything else later that needs to be removed anyway but if you think this should be `all?` instead I'm happy to change it because in theory that would be fewer cases to parse.

lib/mutant/cli/command/environment.rb
mbj: Its re-using the same object for different purposes, re-using objects if fine by me.

lib/mutant/matcher/method.rb
mbj: I'll go for `#all?` semantics here to have symmetry with subject filters.

lib/mutant/cli/command/environment.rb
```diff
@@ -103,10 +102,7 @@ def add_matcher_options(parser)
             add_matcher(:start_expressions, @config.expression_parser.call(pattern).from_right)
           end
           parser.on('--since REVISION', 'Only select subjects touched since REVISION') do |revision|
-            add_matcher(
-              :subject_filters,
```
mbj, line 107: @dgollahon I could get a rid of the entire `subject_filters` after checking private integrations, they where ported to the hooks subsystem so no need to maintain this API. this overall makes this commit much nicer than previous versions.

lib/mutant/matcher/method.rb
```diff
@@ -120,6 +120,11 @@ def subject_config(node)
         def matched_view
```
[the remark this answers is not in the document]
mbj: Works perfectly for me on main. Can you share a reproducing CLI?

mbj (approved):

lib/mutant/cli/command/environment.rb
dgollahon: Awesome! Thank you for validating and simplifying that. I think it's much nicer code-wise, just didn't realize it was an option.

lib/mutant/matcher/config.rb
mbj: I've done the change for you.

PR #1383, opened 2023-05-21
Add extracted Repository::Diff#touches_path

mbj:
Extraction from #1382


dgollahon (approved):
Cool, thanks for extracting 👍

PR #1384, opened 2023-05-21
Add RSpectre to CI

dgollahon:
- This can help detect unused test setup on an ongoing basis and also helps me by giving me a bigger corpus of [`rspectre`](https://github.com/dgollahon/rspectre) users.


dgollahon:
So you may not want to add this as-is because it runs the whole test suite. This isn't currently avoidable in `rspectre` because that's part of how it avoids false positives.

What we could consider instead is making it a scheduled job that runs periodically which Github Actions seems to support.

In the future I may make it possible to run on subsets of tests correctly but it will mean false negatives because it can't trace the entire test suite.

mbj:
> What we could consider instead is making it a scheduled job that runs periodically which Github Actions seems to support.

Happy to set this up. But: Mutant will soon have a parallel "non mutation test" runner, this may be fast enough to run more regular, at least for just `spec/unit` subset?

dgollahon:
> But: Mutant will soon have a parallel "non mutation test" runner

I won't be able to make use of it since `rspectre` is itself an rspec runner

> Edit: maybe just scoping this integration to spec/unit for now is the best trade off?

Yes, I think that might make sense but I need to add support in `rspectre` to exclude shared examples generally or add `rspectre:disable` comments or something to that effect, otherwise we will get false positives. It's something I want to think about the right way to support anyway but outside of a cron-based run I'm not sure what a good option would be fore now.

Cron run may just be the right way to use it though since it's an infrequent issue.

PR #1400, opened 2023-10-04
Add support for forwarded rest arguments

dgollahon:
- Allows `def foo(*); bar(*); end` to work without crashing `mutant`.

### DIFF
--- a/lib/mutant/mutator/node/arguments.rb
+++ b/lib/mutant/mutator/node/arguments.rb
@@ -25,7 +25,7 @@ module Mutant
           children.each_with_index do |removed, index|
             new_arguments = children.dup
             new_arguments.delete_at(index)
-            unless n_forward_arg?(removed) || removed_block_arg?(new_arguments) || only_mlhs?(new_arguments)
+            unless forward_type?(removed) || removed_block_arg?(new_arguments) || only_mlhs?(new_arguments)
               emit_type(*new_arguments)
             end
           end
@@ -36,7 +36,12 @@ module Mutant
         end
 
         def forward_arg?
-          children.last && n_forward_arg?(children.last)
+          children.any?(&method(:forward_type?))
+        end
+
+        def forward_type?(node)
+          n_forward_arg?(node) ||
+              ((n_restarg?(node) || n_kwrestarg?(node)) && node.children.empty?)
         end
 
         def removed_block_arg?(new_arguments)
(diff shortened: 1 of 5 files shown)


lib/mutant/mutator/node/arguments.rb
```diff
@@ -36,7 +36,12 @@ def only_mlhs?(new_arguments)
         end
 
         def forward_arg?
-          children.last && n_forward_arg?(children.last)
+          children.any?(&method(:forward_type?))
+        end
+
+        def forward_type?(node)
+          n_forward_arg?(node) ||
+              ((n_restarg?(node) || n_kwrestarg?(node)) && node.children.empty?)
```
dgollahon, line 44: @mbj This isn't as strict as it could theoretically be. I believe this will remove mutations like:

`def foo(*); end` -> `def foo; end` (and the same for kwargs) even if there is no argument forwarding. I think this is fine. The alternative would be to recurse into the body and look for references to the arguments being forwarded.

lib/mutant/mutator/node/send.rb
```diff
@@ -223,7 +223,7 @@ def emit_argument_propagation
 
           argument = Mutant::Util.one(arguments)
 
-          return if n_kwargs?(argument) || n_forwarded_args?(argument)
+          return if n_kwargs?(argument) || n_forwarded_args?(argument) || n_forwarded_restarg?(argument)
```
dgollahon, line 226: I wonder if this should check for forwarded keyword arguments too? https://github.com/mbj/mutant/pull/1401

dgollahon:
Had to pause for now because I hit my timebox and it still needs a bit more mutation coverage. Feel free to fix it if you have bandwidth @mbj.

mbj:
@dgollahon I've tried to reproduce the crash you are mentioning in the PR, I cannot crash mutant via:

```
bundle exec mutant util mutation -e 'def a(&); a(&); end; def b(*); b(*); end; def c(**); c(**); end'
```

lib/mutant/mutator/node/send.rb
mbj: no as this is covered by `n_kwargs?` as `kwrestarg` forwarding is inside `s(:kwargs)` nodes.

mbj:
@dgollahon I've ported this up to #1401 with more fixes. Closing this one, you'll get co-author over there.

PR #1401, opened 2023-10-04
Fix forwarded kwrestarg

dgollahon:
@mbj  Seems like this can also fail too but I'm not sure I have time to look into it today but I thought I'd throw up a failing test at least. I assume it probably makes sense to define the AST structure but not sure exactly where that is used.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+* Add support for forwarded rest arguments `def foo(*); bar(*); end` which previously caused `mutant` to crash.
+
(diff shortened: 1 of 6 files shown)


mbj:
@dgollahon This is at minimum, first level an unparser issue patching unparser to round trip this first.

mbj:
Unparser fixed in 0.6.9: https://github.com/mbj/unparser/blob/main/Changelog.md#v068-2023-10-08

@dgollahon I'm centralizing these changes in #1402, closing this PR, will add you as co-author to my commit.

PR #1430, opened 2024-03-20
Fix rspec is quitting signal

mbj:


lib/mutant/integration/rspec.rb
```diff
@@ -102,6 +102,14 @@ def available_tests
 
     private
 
+      def rspec_setup_failure?
+        @rspec_world.wants_to_quit || rspec_is_quitting?
```
mbj, line 106: RSpec (3.9 and above) have "more than one signal" that setup failed. Mutant needs to respect both. One would assume that `rspec_is_quitting` would imply `wants_to_quit` but this is not the case.

PR #1444, opened 2024-05-12
Add ordering mutations

### LINKED ISSUE #1442: feat: Additional mutations around ordering
Suggestion:

Mutant should include mutations that replace methods like `find/detect/min/max/min_by/max_by/last` with `first` and `last`

Justification

The methods `min/max/min_by/max_by` all accept a block. However if there is only 1 item in the collection, the block will not executed, and potentially remain untested. It's an easy fact to overlook in your tests. In those situations a mutation to `first` will result in a live mutation, making the missing coverage obvious.

Additionally mutating to last will detect when your sort conditions aren't being properly exercised because all the records sort in the same order.

`find/detect` are some what different, since the blocks for those methods will always execute, but it will still help to ensure that your test setup creates collections that both include and exclude the values you're working with.

mbj:
[fix #1442]

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,17 @@
 # v0.12.1 2024-05-11
 
+* [#1444](https://github.com/mbj/mutant/pull/1444)
+
+  Add ordering mutations to full and light operator set,
+  `#find/#detect/#min/#max/#min_by/#max_by -> #first/#last`. These mutations force
+  to provide test cases with more than one element to observe the extra
+  semantics by these methods over just `#first` and `#last`.
+
+  Add `#first -> #last` and `#last -> #first` to full operator set. These where
+  highly controversial in the past and thus where removed. People who do not care
+  about ordering of their collections can now choose the light operator set, so people who care
+  (and should care) can now care (again).
+
 * [#1443](https://github.com/mbj/mutant/pull/1443)
 
   Change mutants test runner to always report in test definition order deterministically.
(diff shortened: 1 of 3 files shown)


Changelog.md
```diff
@@ -1,5 +1,17 @@
 # v0.12.1 2024-05-11
 
+* [#1444](https://github.com/mbj/mutant/pull/1444)
+
+  Add ordering mutations to full and light operator set,
+  `#find/#detect/#min/#max/#min_by/#max_by -> #first/#last`. These mutations force
+  to provide test cases with more than one element to observe the extra
+  semantics by these methods over just `#first` and `#last`.
+
+  Add `#first -> #last` and `#last -> #first` to full operator set. These where
```
dgollahon, line 10: typo: where -> were

Changelog.md
```diff
@@ -1,5 +1,17 @@
 # v0.12.1 2024-05-11
 
+* [#1444](https://github.com/mbj/mutant/pull/1444)
+
+  Add ordering mutations to full and light operator set,
```
dgollahon, line 5: I am slightly confused as to whether or not these were added to the full operator set because i think the deletion logic makes this just the full set and you also say:

>  People who do not care about ordering of their collections can now choose the light operator s

mbj: @dgollahon Sorry for the late reply, the light operators do not include these operators. I've clarified the readme somewhat.

PR #1445, opened 2024-05-12
Fix coverage

mbj:


lib/mutant/util.rb
```diff
@@ -12,9 +12,12 @@ module Util
     #
     # @return [Object] first entry
     def self.one(array)
-      return array.first if array.one?
```
mbj, line 15: `Array#one?` comes form `Enumerable#one?` which returns true if there is one item that returns truthy. So `[nil].one?` would return `false` which is not the intention of this code.

There is no call side in mutant that would have been affected by the behavior difference, still it has to be fixed.

PR #1455, opened 2024-06-30
Fix rspec metadata usage

mbj:


mbj:
This was released as `mutant-v0.12.4`.

PR #1467, opened 2025-05-29
Tweak `Util::SizeError` comment

viralpraxis:


lib/mutant/util.rb
```diff
@@ -3,7 +3,7 @@
 module Mutant
   # Utility methods
   module Util
-    # Error raised by `Util.one` if size is less than zero or greater than one
+    # Error raised by `Util.one` if size is not exactly one
```
mbj, line 6: @viralpraxis Since mutant is commercial software I need you to sign over the rights before I can merge even the slightest improvement. I'll ping you out of band ;)

Its a good catch.

mbj:
This is redundant now as the code was removed, using unparser version now.

PR #1484, opened 2025-11-03
Fix test runner docs

mbj:
* I documented what I meant the test runner to be like, not what they actually are.


samandmoore (approved):
thanks for updating!

PR #1488, opened 2025-12-26
Allow minitest 6

sferik:
I also updated the `irb` dependency to allow versions >= 1.15 and < 2. Plus some other cleanup.

Once this is merged, please push a `mutant-minitest` gem release so I can start using Minitest 6, which [shipped on December 17](https://rubygems.org/gems/minitest/versions/6.0.0) and [is bundled with Ruby 4.0.0](https://www.ruby-lang.org/en/news/2025/12/25/ruby-4-0-0-released/).


mbj:
Superseeded by https://github.com/mbj/mutant/pull/1490, & ty.

PR #1500, opened 2025-12-30
Add Ruby-4

mbj:


sferik:
I noticed [a random test failure in your branch](https://github.com/mbj/mutant/actions/runs/20604646738/job/59179653343?pr=1500), despite it [passing in my branch](https://github.com/mbj/mutant/actions/runs/20604323184).

I believe this is due to the fact that, in Ruby 4.0, `read_nonblock` with `exception: false` can return `:wait_readable` even after `select` indicates readability (spurious wakeup). I believe we could catch the `when :wait_readable` case here and return `false`: https://github.com/mbj/mutant/blob/a0413a28d6937b62ba91af9262c5d5a4fe059194/ruby/lib/mutant/parallel/connection.rb#L117-L133

Let me know what you think of this approach and I can push up another commit.

PR #1510, opened 2026-01-08
Add visual progress bar for TTY output

sferik:
Replace periodic text-based progress output with a (nextest-style) progress bar that updates in place when running in a TTY. It looks something like this:

```
RUNNING 45/100 (45.0%) ████████████░░░░░░░░░░░░ alive: 12 23.4s 1.89/s
```

It falls back to text format for non-TTY output.

### DIFF
--- a/ruby/lib/mutant/reporter/cli/progress_bar.rb
+++ b/ruby/lib/mutant/reporter/cli/progress_bar.rb
@@ -63,7 +63,7 @@ module Mutant
         end
 
         def empty_width
-          [width - filled_width, 0].max
+          width - filled_width
         end
 
         def filled
--- a/ruby/spec/unit/mutant/reporter/cli/progress_bar_spec.rb
+++ b/ruby/spec/unit/mutant/reporter/cli/progress_bar_spec.rb
@@ -10,6 +10,34 @@ RSpec.describe Mutant::Reporter::CLI::ProgressBar do
     it 'creates a progress bar with default settings' do
       expect(subject).to be_a(described_class)
     end
+
+    it 'sets current from parameter' do
+      expect(subject.current).to eql(current)
+    end
+
+    it 'sets total from parameter' do
+      expect(subject.total).to eql(total)
+    end
+
+    it 'uses DEFAULT_WIDTH for width' do
+      expect(subject.width).to eql(described_class::DEFAULT_WIDTH)
+    end
+
+    it 'uses FILLED_CHAR for filled_char' do
+      expect(subject.filled_char).to eql(described_class::FILLED_CHAR)
+    end
+
+    it 'uses EMPTY_CHAR for empty_char' do
+      expect(subject.empty_char).to eql(described_class::EMPTY_CHAR)
+    end
+
+    context 'with custom width' do
+      subject { described_class.build(current:, total:, width: 20) }
+
+      it 'uses provided width' do
+        expect(subject.width).to eql(20)
+      end
+    end
   end
 
   describe '#render' do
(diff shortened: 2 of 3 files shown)


mbj:
@sferik I used to have a *per subject* progress bar back in the days and had lots of stability issues as terminal emulators where shit. Now as this one is a single one I suspect it'll not be a problem. Conceptually I'm fine with a more rich one and with the upcoming rust wrapper I'll also create them. But no reason to not release a ruby one meanwhile.

But I need some people with various setup to test this. Lets create a small test command we can post here and ask some friends to post their arch/os/terminal results? I do not want to break reporting display for people "unknowingly".

ruby/lib/mutant/reporter/cli/printer/test.rb
```diff
@@ -117,7 +117,9 @@ def run
 
           # Reporter for progressive output format on scheduler Status objects
           class StatusProgressive < self
-            FORMAT = 'progress: %02d/%02d failed: %d runtime: %0.02fs testtime: %0.02fs tests/s: %0.02f'
+            FALLBACK_FORMAT = 'progress: %02d/%02d failed: %d runtime: %0.02fs testtime: %0.02fs tests/s: %0.02f'
+
+            BAR_WIDTH = 24
```
mbj: whatever prefix we choose for the TTY format this prefix should be mirrored here in the const name `TTY_BAR_WIDTH`?

ruby/lib/mutant/reporter/cli/printer/test.rb
```diff
@@ -142,14 +162,37 @@ def run
...
+                bar.percentage,
+                bar.render,
+                amount_tests_failed,
+                runtime,
+                tests_per_second
+              )
+            end
+
+            def progress_prefix
+              if amount_tests_failed.positive?
```
mbj: Same value on both branches?

ruby/lib/mutant/reporter/cli/printer/status_progressive.rb
```diff
@@ -21,8 +22,26 @@ class StatusProgressive < self
           #
           # @return [undefined]
           def run
+            if tty?
```
mbj: Lots of "static to the execution" runtime branches, you could lift this up and make 2 classes `SatusProgressivePiped/StatusProgressiveTTY` and instantiate the fitting class in evaluating `tty?` once.

ruby/lib/mutant/reporter/cli/printer/test.rb
sferik: I think I'm actually going to try to calculate the width dynamically.

ruby/lib/mutant/reporter/cli/printer/status_progressive.rb
sferik: Done in 619e80e0e5081e73f28c59c8213b308d96f89bba.

ruby/lib/mutant/reporter/cli/printer/test.rb
sferik: Done in https://github.com/mbj/mutant/pull/1510/commits/8ed1f10b4d4b444d769cb655a06488f8caaa17ff.

ruby/lib/mutant/reporter/cli.rb
```diff
@@ -6,19 +6,34 @@ class Reporter
...
         )
       end
 
+      def self.detect_terminal_width(output)
+        if output.respond_to?(:winsize)
+          output.winsize.last
+        else
+          DEFAULT_TERMINAL_WIDTH
+        end
+      rescue Errno::ENOTTY
```
mbj: Can this branch be triggered even? Can we only run this on `tty?` true in the first place?

sferik:
Okay, this should be ready for re-review and testing on different platforms and terminal emulators.

For what it's worth, my testing has been on macOS Tahoe (26.2). It works as expected on Ghostty (1.2.3), iTerm2 (3.6.6), and the default Terminal (2.15).

sferik:
@mbj Any *progress* on this?

Issue #201, opened 2014-06-05
Mutating break to next violates mutate to simpler primitive rule


dkubb:
Right now if you have a `break` it will be mutated to `next`. I think this violates the simpler primitive rule.

In general, if you can preserve the functionality of the method and `break` early, this is preferable to using `next` to skip over each iteration.

PR #1525, opened 2026-01-19
Add break → next mutation for symmetry with next → break

sferik:
The `next` → `break` mutation already exists, but the reverse was missing. This adds the `break` → `next` mutation to provide bidirectional coverage for loop control flow mutations.


sferik:
See https://github.com/mbj/mutant/issues/201.

PR #1526, opened 2026-01-19
Add empty collection return value mutations

sferik:
Add mutations that replace collection-returning expressions with empty values.

* `to_a`/`to_ary` calls mutate to empty array
* `to_h`/`to_hash` calls mutate to empty hash
* `to_s`/`to_str` calls mutate to empty string
* String literals mutate to empty string


mbj:
@sferik this one needs a rebase as we have merged the empty string #1547 here.

PR #1548, opened 2026-01-29
Add extend ↔ include ↔ prepend module inclusion mutations

sferik:


mbj:
Unfortunately boot mutation restriction: https://github.com/mbj/mutant/pull/1545#issuecomment-3818897882

Issue #1609, opened 2026-04-16
`JSON::GeneratorError` crash in `Result::JSONWriter` when test output contains binary data


sferik:
`Result::JSONWriter` (added in https://github.com/mbj/mutant/commit/c8a4a4a482b47cb3d5b66b49b95c6e428dbf910b) crashes with `JSON::GeneratorError` when any mutation test log contains ASCII-8BIT strings with bytes that are invalid UTF-8. The crash occurs in the main process after all mutations have been processed, so the run reports 100% coverage but exits with code 1.

Any project with a binary fixture file referenced in tests can trigger this. A minimal case:

1. A test uses `File.binread` to load a binary fixture (e.g., a GIF file)
2. That fixture is used as a WebMock request body
3. A mutation causes a request mismatch, so WebMock's failure message includes the raw binary data
4. The mutation is killed (correctly), but the test log now contains binary bytes
5. After the run completes, `JSONWriter#json` calls `JSON.generate` on the session data, which includes the binary log
6. `JSON.generate` raises `JSON::GeneratorError` because the log string is ASCII-8BIT and contains bytes like `\xF7` that cannot be transcoded to UTF-8

Test output logs are captured from pipes opened with `binmode: true` (`Parallel::Pipe.from_io`), so they always have ASCII-8BIT encoding. `JSONWriter` passes these strings directly to `JSON.generate` without any encoding conversion.

See https://github.com/sferik/twitter-ruby/actions/runs/24430151086 for details.

mbj:
I promised in DM to fix this tonight, but I'm out of energy (speaking at Wrocloverb) and travel zapped my energy will fix this tomorrow. Good catch and thanks for the report!

mbj:
> Test output logs are captured from pipes opened with binmode: true (Parallel::Pipe.from_io), so they always have ASCII-8BIT > encoding. JSONWriter passes these strings directly to JSON.generate without any encoding conversion.

These kinds of bugs are in part the result of not using a proper typesystem should be "conceptually" fixed with mutant-v1.0.0 that is upcomgin and ships a rust reporter that has proper end 2 end typing.

sferik:
I’m excited for v1, but in the mean time this is breaking my build. 😜

mbj:
To be clear this is going to be fixed before v1 ;)

PR #1612, opened 2026-04-20
Fix JSON::GeneratorError on non-UTF-8 log bytes

sferik:
My hacky attempt at fixing https://github.com/mbj/mutant/issues/1609.

@mbj You’ll probably come up with something cleaner but this fixes the problem.


sferik:
Closed in favor of https://github.com/mbj/mutant/pull/1613.

PR #1624, opened 2026-04-30
Integrate with tldr

mostlyobvious:
* adds support for running mutations on codebase tested with https://github.com/tendersearls/tldr
* mostly follows minitest integration
* similarity of test file patters required to reorganize how test_app tests are located in relation to the particular integraion, so that tldr does not load minitest classes and vice-versa


mbj:
@mostlyobvious chances you can add a corpus test (ideally a PR to `anima` we can referenc here as its fully covered with minitest/rspec already) so we have an 100% mutation covered reference?

mostlyobvious:
@mbj will do

mbj:
@mostlyobvious still interested?

Issue #1360, opened 2022-12-15
SQL mutations


swistak35:
I've thought recently about rewriting part of my code to use (almost, because escaping is still useful for security) raw SQL instead of `ActiveRecord` code. Of course that would result in the raw SQL strings not being covered by mutation tests. Have you ever thought about mutating SQLs, or maybe more reasonably, having ability to integrate with external tooling which could mutate SQL queries (or something very close to them, because again, escaping). SQL could be differentiated from other strings by using specific heredoc :wink: More like a dream/discussion than the feature suggestion, but I'm curious whether you were thinking about it at any point.

mbj:
@swistak35 I've got a private POC but not a full tool for SQL mutations. I need to find time during a sabattical to work on it. Note it would be commercial only with free for OSS model just like mutant.

PR #1656, opened 2026-08-25
Addind SQL mutations

bsene:
resolve issue #1360

### DIFF
--- a/ruby/lib/mutant/mutator/sql.rb
+++ b/ruby/lib/mutant/mutator/sql.rb
@@ -2,59 +2,193 @@
 
 module Mutant
   class Mutator
-    # Mutator for SQL embedded in Ruby heredoc strings
+    # SQL mutation backed by a real parser
     #
-    # Operates on the raw SQL text and emits mutated SQL strings.
-    # All mutations are orthogonal replacements following mutant's
-    # direction rules — see CONTRIBUTING.md §"Mutation Direction Rules".
+    # The SQL heredoc body is parsed with +pg_query+ into a PostgreSQL parse
+    # tree, mutated one typed node at a time, and deparsed back to SQL text.
+    #
+    # Using the AST — instead of flipping keywords with a regex — avoids the
+    # edge cases a regex cannot handle: the literal string "OR" inside a
+    # query, "AND" inside a comment, "IN" as a substring of an identifier.
+    # The parser distinguishes string literals and comments from operators,
+    # so only real operators are ever mutated.
+    #
+    # A heredoc body that is not valid PostgreSQL yields no SQL mutations
+    # rather than raising, so a non-PostgreSQL dialect is simply left alone.
     module Sql
-      # Ordered longest-first so multi-word keywords match before their
-      # sub-keywords (e.g. "IS NOT NULL" before "IS NULL", "NOT IN" before "IN").
-      REPLACEMENTS = [
-        ['IS NOT NULL', 'IS NULL'],
-        ['IS NULL',     'IS NOT NULL'],
-        ['NOT EXISTS',  'EXISTS'],
-        ['EXISTS',      'NOT EXISTS'],
-        ['NOT IN',      'IN'],
-        ['IN',          'NOT IN'],
-        ['AND',         'OR'],
-        ['OR',          'AND'],
-        ['ASC',         'DESC'],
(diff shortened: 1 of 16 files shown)


ruby/lib/mutant/mutator/sql.rb
```diff
@@ -0,0 +1,60 @@
...
+        ['AND',         'OR'],
+        ['OR',          'AND'],
+        ['ASC',         'DESC'],
+        ['DESC',        'ASC']
+      ].freeze
+
+      MAP = REPLACEMENTS.to_h.freeze
+
+      # Combined pattern matching SQL keywords to mutate.
+      PATTERN = ::Regexp.new(
```
mbj: @bsene Thank you for your contribution, but I think regexp based syntax modification will never work at scale. It has to many edge cases. What if an SQL literal contains the string `OR` and all of the other interesting edge cases.

For mutant to gain proper SQL mutations a proper AST should be used. `pg_query` (gem) comes to mind it would have to be paired with an SQL unparser. Just like I made the ruby unparser gem to power mutant.


Would you consider to change this PR to use a proper SQL AST parser/unparser?

bsene: Yes fair point, I rewrote this with a real SQL/AST tool (eg: `pg_query` as you suggested).

What do you think ?

ruby/mutant.yml
```diff
@@ -24,3 +24,9 @@ matcher:
   - Mutant::Mutator::Node::Arguments#emit_argument_presence
   - Mutant::Mutator::Node::Arguments#removed_block_arg?
   - Mutant::Mutator::Node::BlockPass#dispatch
+  # NOT-wrap mutator: mutations that strip the wrapped operand (args:[node]→
+  # nil/[], or dropping bool_expr/boolop) yield a NOT with no operand, which
+  # pg_query's native deparser cannot rebuild (it segfaults) or drops to a bare
```
mbj, line 29: mhh this makes me very vary about the quality of `pg_query` ans this segfault would take mutant with it.

Sure while you work around *this* segfault, there is a high likelyness ther are other segfaults still hiding.

I wounder if we should add a very SQL heavy corpus test before merging to avoid shipping a mutant version that segfaults on yet to be learned SQL?

ruby/lib/mutant/mutator/sql.rb
bsene: The segfault in pg_query's unparser on malformed operands is genuinely an upstream bug:  a deparser should never crash, it should raise or return empty. 

-> I'll submit an issue against [pg_query](https://github.com/pganalyze/pg_query). 

The ignore: entry is just a pragmatic workaround: those degenerate trees are only produced by mutant's self-mutation of `wrap_exists`, never through the public mutate(sql) API, so no real test can distinguish them IMO.

mbj: @bsene if there is one crash there are many. You have an idea about a source of diverse SQL Statements we can hook up to the corpus tests to proof we filtered out the "only" segfault?

bsene: I rely on those [SQL statements](https://sql.sh).

My strategy was first focus on small set of operator flips , fix them then complete the corpus as you suggested although  that should belongs to the pg_query internal tests (if there are present)  -> I'll check it soon 

## UPDATE

Here what I've in mind after investigating:
-  WHERE: AND/OR +  comparison ops
- IN / NOT IN
- IS NULL / IS NOT NULL
- LIKE / NOT LIKE
- BETWEEN / NOT BETWEEN / BETWEEN SYMMETRIC (postgres)
-  IS DISTINCT FROM (postgres) / IS NOT DISTINCT FROM (postgres)
- ORDER BY ASC/DESC + NULLS FIRST/LAST (postgres)
-  EXISTS / NOT EXISTS  -> potential segfault
- NOT ->  potential segfault
- UNION / INTERSECT / EXCEPT
- JOIN
- GROUP BY / HAVING
- Window functions -> OVER(ORDER/PARTITION BY....)
- CASE
- Common Table Expression -> WITH
- Subqueries


What do you think @mbj

ruby/mutant.yml
mbj: @bsene bump

PR #1660, opened 2026-09-04
Select tests from a per test coverage recording

sferik:
The latest version of SimpleCov (1.2.0) adds the ability to `track_tests`, which creates a bitmap in `coverage/coverage.json` from lines of code to the specific tests/specs that exercise them. If this bitmap exists, it can more efficiently select which tests to run (and in which order) to kill a mutation on a particular line of code. 

The selected tests come back ordered likeliest killer first: the ones the expressions would also have picked, then whichever spends the largest share of its own footprint inside this subject, then reach, and finally the framework’s test id to break a tie. Minitest and Test::Unit already run the order they are given and stop at the first failure. RSpec picks its own order, so Mutant installs an ordering strategy. 

The default is `auto`, meaning use the recording when there is a usable one, otherwise use expressions, so nothing changes for a project that doesn’t use SimpleCov. `--selection context_map` instead turns every reason the recording cannot be used into a hard failure, `--selection expression` never looks at it, and `--selection-path` points at a report outside `coverage`. Users can specify the same configuration options under a `selection` key in `mutant.yml`. Every run prints which one it settled on, because auto-detection you cannot see is a trap.

I’ve added full documentation in `docs/test-selection.md` as well as a `selection` section to `docs/configuration.md`.

To benchmark this change, I ran mutant for SimpleCov (206 subjects and 7000 mutations) against this branch. Median wall clock dropped from 230s to 185s (20% faster), owing largely to running fewer tests per subject (from 161 down to 79).


ruby/lib/mutant/context_map/loader.rb
```diff
@@ -0,0 +1,208 @@
...
+      # Load the recording
+      #
+      # @return [Either<String, ContextMap>]
+      def call
+        return error(MISSING_ARTIFACT) unless artifact.file?
+
+        read.bind(&method(:parse)).bind(&method(:from_document))
+      end
+
+    private
```
mbj, line 183: @sferik All of the decoders that exist down from here are basically a manual implementation of hte transform pattern we use in this code base. Chances you can change to a transformer?

sferik: Done. The loader now declares the report shape as transforms (`DOCUMENT`, `META`, `COVERAGE`, `FILE`, `TABLE`) and the manual type checks and regexp decoders are gone.

There were two addition things that `Transform` did not have that SimpleCov's report needs, so I added them under `Transform::Hash`:

* `Hash::Slice` narrows a hash to the keys it names. SimpleCov writes many more keys than Mutant reads (`$schema`, `total`, `groups`, `errors`, `history`, ...) and adds new ones between minor schema versions, so the strict `Hash` transform can't see the raw document. Slicing first keeps `Hash` strict about the keys it does know.
* `Hash::Map` maps a data keyed hash (the `coverage` section keyed by file path, the per file `contexts` table keyed by context index) over key and value transforms, wrapping errors in `Hash::Key` so the path in the message names the offending key. It rejects a key transform that collapses two input keys onto one, since a silently dropped pair would be a silent gap in coverage.

The remaining decoder is the digit parser behind `Transform::Block`. `Kernel#Integer` also accepts a sign, a `0x` prefix, whitespace and underscores, none of which the format allows, so it stays behind a regexp.

mbj (approved):

